PR #169, opened 2014-03-30
Use shell instead of nix-env for system profile setup

rickynils:
This makes it possible for nixops to activate systems
even if their Nix db is read-only. nix-env always
require write-access to the db, but since we already have
a built system path, we don't have to use nix-env to just
set up the system profile symlinks.


domenkozar:
Closing this as original PR was also closed.

PR #170, opened 2014-03-30
Add deployment.useHostNixStore option

rickynils:
If true, no store paths are copied to the deployed machine. Instead
the Nix store of host is mounted on the deployed machine and used in
read-only mode. This option is not supported by all nixops target
environments. The <literal>none</literal> target environment supports it
but don't provide an automatic way of setting up the required store
mount.


domenkozar:
@rickynils what's the status on this? It needs changes for `none` and `virtualbox` backends to actually mount the nix store.

rickynils:
I'll close this. It's too long since I had this running, and I don't need this use case right now.

Issue #156, opened 2013-12-09
Multiple exceptions: need more than 1 value to unpack, need more than 1 value to unpack


peti:
nixops can no longer deploy machines in EC2. I tried the 1.1.1 release and the latest version from git, but both versions just give this error:

    $ nixops deploy
    slave2> waiting for IP address... 
    slave1> waiting for IP address... 
    slave2> [running] 79.125.81.202 / 10.73.183.230
    slave1> [running] 54.217.176.190 / 10.73.133.146
    error: Multiple exceptions: need more than 1 value to unpack, need more than 1 value to unpack

The --debug option adds this output:

    error: Multiple exceptions: need more than 1 value to unpack, need more than 1 value to unpack
    ------------------------------
    Traceback (most recent call last):
      File "/nix/store/1c7r69fya6xyrly2zdpd5mpw7vk955rw-nixops-1.1.1/lib/python2.7/site-packages/nixops-1.1.1-py2.7.egg/nixops/parallel.py", line 41, in thread_fun
        result_queue.put((worker_fun(t), None))
      File "/nix/store/1c7r69fya6xyrly2zdpd5mpw7vk955rw-nixops-1.1.1/lib/python2.7/site-packages/nixops-1.1.1-py2.7.egg/nixops/deployment.py", line 869, in worker
        r.create(self.definitions[r.name], check=check, allow_reboot=allow_reboot, allow_recreate=allow_recreate)
      File "/nix/store/1c7r69fya6xyrly2zdpd5mpw7vk955rw-nixops-1.1.1/lib/python2.7/site-packages/nixops-1.1.1-py2.7.egg/nixops/backends/ec2.py", line 825, in create
        self._wait_for_ip(instance)
      File "/nix/store/1c7r69fya6xyrly2zdpd5mpw7vk955rw-nixops-1.1.1/lib/python2.7/site-packages/nixops-1.1.1-py2.7.egg/nixops/backends/ec2.py", line 295, in _wait_for_ip
        nixops.known_hosts.add(instance.ip_address, self.public_host_key)
      File "/nix/store/1c7r69fya6xyrly2zdpd5mpw7vk955rw-nixops-1.1.1/lib/python2.7/site-packages/nixops-1.1.1-py2.7.egg/nixops/known_hosts.py", line 44, in add
        _rewrite(ip_address, public_host_key)
[...]

PR #257, opened 2015-01-19
Need more than one value

blast-hardcheese:
Some work towards resolving #156

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -317,6 +317,10 @@ class MachineState(nixops.resources.ResourceState):
     def has_really_fast_connection(self):
         return False
 
+    def write_host_key(self):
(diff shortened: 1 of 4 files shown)


blast-hardcheese:
I'm not convinced this stuff is ready for prime-time, my current thoughts are in #156. The current state of the world is this:

- Refactored `nixops.known_hosts` to `add` keys with the comment `# nixops` at the end of the line, made `remove` ignore all lines that don't end with `# nixops`.
- On build, use `ssh-keyscan` and `nixops.known_hosts` instead of relying on `-oStrictHostKeyChecking=no`
- Remove keys from `known_hosts` on resource delete

domenkozar:
This can be closed now right, since it's fixed. Reopen if I'm mistaken.

PR #264, opened 2015-02-12
Fixed calling the wrong known_hosts.remove function

basvandijk:
This bug prevents deploying to Hetzner.


basvandijk:
Thanks for the quick merge!

PR #269, opened 2015-02-16
Azure Backend

Phreedom:
WIP.

Currently you have to build and register a nixos bootstrap image by hand.

A sample deployment:
```
let
  # change this as necessary or wipe and use ENV vars
  credentials = {
    subscriptionId = "f1ce4500-ab06-495a-8d59-a7cfe9e46dae";
    certificatePath = "path/mycert.pem";
  };
in
{

  resources.azureAffinityGroups.testAG = credentials // {
    location = "West US";
    label = "who needs this?";
    description = "random";
  };

  resources.azureReservedIPAddresses.testIP = credentials // {
    location = "West US";
  };

  resources.azureReservedIPAddresses.debugIP = credentials // {
    location = "West US";
  };

  resources.azureHostedServices.testHS = {resources,...}: credentials // {
    #location = "West US";
    affinityGroup = resources.azureAffinityGroups.testAG;
    label = "who needs this?";
  };

  resources.azureStorages.testS = {resources,...}: credentials // {
    name = "mystorage";
    location = "West US";
    label = "who cares";
    description = "asdfafa";
  };

  resources.azureBlobContainers.testBC =  {resources,...}: credentials // {
    name = "mycontainer";
    storage = resources.azureStorages.testS;
    acl = "blob";
  };

  resources.azureBlobs.testblob =  {resources,...}: credentials // {
    container = resources.azureBlobContainers.testBC;
    contentType = "text/plain";
    filePath = "path/blob.html";
  };

  machine = { resources, ...}: {
    networking.firewall.allowPing = true;
    deployment.targetEnv = "azure";
    deployment.azure = credentials // {
      ipAddress = resources.azureReservedIPAddresses.testIP;
      storage = resources.azureStorages.testS;
      hostedService = resources.azureHostedServices.testHS;
      rootDiskImage = "nixos";
      rootDiskUrl = "http://mystorage.blob.core.windows.net/mycontainer/machine-root";
    };

  };

}
```

### DIFF
--- a/doc/manual/manual.xml
+++ b/doc/manual/manual.xml
@@ -205,6 +205,203 @@
 
     </section>
 
+    <section>
+      <title>Azure Resources</title>
+
+      <para>This section lists resource types associated with the
+      Microsoft Azure (Azure) cloud computing environment.</para>
+
+      <section>
+        <title>Azure Availability Sets</title>
+        <para>An Azure Availability Set is defined by setting
+        <literal>resources.azureAvailabilitySets.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-availability-set-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure BLOB Containers</title>
+        <para>An Azure BLOB Container is defined by setting
+        <literal>resources.azureBlobContainers.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-blob-container-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure BLOBs</title>
+        <para>An Azure BLOB is defined by setting
+        <literal>resources.azureBlobs.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-blob-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Directories</title>
+        <para>An Azure Directory is defined by setting
+        <literal>resources.azureDirectories.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-directory-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure DNS Record Sets</title>
+        <para>An Azure DNS Record Set is defined by setting
+        <literal>resources.azureDNSRecordSets.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-dns-record-set-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure DNS Zones</title>
+        <para>An Azure DNS Zone is defined by setting
+        <literal>resources.azureDNSZones.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-dns-zone-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure ExpressRoute Circuits</title>
+        <para>An Azure ExpressRoute Circuit is defined by setting
+        <literal>resources.azureExpressRouteCircuits.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-express-route-circuit-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Files</title>
+        <para>An Azure File is defined by setting
+        <literal>resources.azureFiles.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-file-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Gateway Connections</title>
+        <para>An Azure Gateway Connection is defined by setting
+        <literal>resources.azureGatewayConnections.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-gateway-connection-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Load Balancers</title>
+        <para>An Azure Load Balancer is defined by setting
+        <literal>resources.azureLoadBalancers.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-load-balancer-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Local Network Gateways</title>
+        <para>An Azure Local Network Gateway is defined by setting
+        <literal>resources.azureLocalNetworkGateways.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-local-network-gateway-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Network Security Groups</title>
+        <para>An Azure Network Security Group is defined by setting
+        <literal>resources.azureSecurityGroups.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-network-security-group-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Queues</title>
+        <para>An Azure Queue is defined by setting
+        <literal>resources.azureQueues.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-queue-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Reserved IP Addresses</title>
+        <para>An Azure Reserved IP Address is defined by setting
+        <literal>resources.azureReservedIPAddresses.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-reserved-ip-address-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Resource Groups</title>
+        <para>An Azure Resource Group is defined by setting
+        <literal>resources.azureResourceGroups.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-resource-group-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Shares</title>
+        <para>An Azure Share is defined by setting
+        <literal>resources.azureShares.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-share-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Storages</title>
+        <para>An Azure Storage is defined by setting
+        <literal>resources.azureStorages.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-storage-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Tables</title>
+        <para>An Azure Table is defined by setting
+        <literal>resources.azureTables.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-table-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Traffic Manager Profiles</title>
+        <para>An Azure Traffic Manager Profile is defined by setting
+        <literal>resources.azureTrafficManagerProfiles.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-traffic-manager-profile-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Virtual Networks</title>
+        <para>An Azure Virtual Network is defined by setting
+        <literal>resources.azureVirtualNetworks.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-virtual-network-options.xml" />
+      </section>
+
+      <section>
+        <title>Azure Virtual Network Gateways</title>
+        <para>An Azure Virtual Network Gateway is defined by setting
+        <literal>resources.azureVirtualNetworkGateways.<replaceable>name</replaceable></literal>
+        to an attribute set containing values for the following
+        options.</para>
+        <xi:include href="azure-virtual-network-gateway-options.xml" />
+      </section>
+
+    </section>
+
   </appendix>
 
   <xi:include href="hacking.xml" />
--- a/doc/manual/nixops.xml
+++ b/doc/manual/nixops.xml
@@ -189,16 +189,21 @@ cloud.</para>
   </varlistentry>
 
   <varlistentry><term><envar>AZURE_SUBSCRIPTION_ID</envar></term>
-    <term><envar>AZURE_CERTIFICATE_PATH</envar></term>
+    <term><envar>AZURE_AUTHORITY_URL</envar></term>
+    <term><envar>AZURE_USER</envar></term>
+    <term><envar>AZURE_PASSWORD</envar></term>
 
-    <listitem><para>Azure subscription ID and path to the
-    corresponding certificate(*.pem) file.
+    <listitem><para>Azure subscription ID, authority URL, user and password.
     Used if not set in an Azure machine deployment configuration via
-    <varname>deployment.azure.subscriptionId</varname>
-    and <varname>deployment.azure.certificatePath</varname>,
+    <varname>deployment.azure.subscriptionId</varname>,
+    <varname>deployment.azure.authority</varname>,
+    <varname>deployment.azure.user</varname>
+    and <varname>deployment.azure.password</varname>,
     and if not set in an Azure resource specification via
-    <varname>resources.$TYPE.$NAME.subscriptionId</varname> and
-    <varname>resources.$TYPE.$NAME.certificatePath</varname>.
+    <varname>resources.$TYPE.$NAME.subscriptionId</varname>,
+    <varname>resources.$TYPE.$NAME.authority</varname>,
+    <varname>resources.$TYPE.$NAME.user</varname>
+    and <varname>resources.$TYPE.$NAME.password</varname>.
     </para></listitem>
   </varlistentry>
 
--- a/doc/manual/nixops.xml
+++ b/doc/manual/nixops.xml
@@ -191,18 +191,22 @@ cloud.</para>
   <varlistentry><term><envar>AZURE_SUBSCRIPTION_ID</envar></term>
     <term><envar>AZURE_AUTHORITY_URL</envar></term>
     <term><envar>AZURE_USER</envar></term>
+    <term><envar>AZURE_SERVICE_PRINCIPAL</envar></term>
     <term><envar>AZURE_PASSWORD</envar></term>
 
-    <listitem><para>Azure subscription ID, authority URL, user and password.
+    <listitem><para>Azure subscription ID, authority URL, user,
+    service principal and password.
     Used if not set in an Azure machine deployment configuration via
     <varname>deployment.azure.subscriptionId</varname>,
     <varname>deployment.azure.authority</varname>,
-    <varname>deployment.azure.user</varname>
+    <varname>deployment.azure.user</varname>,
+    <varname>deployment.azure.servicePrincipal</varname>
     and <varname>deployment.azure.password</varname>,
     and if not set in an Azure resource specification via
     <varname>resources.$TYPE.$NAME.subscriptionId</varname>,
     <varname>resources.$TYPE.$NAME.authority</varname>,
-    <varname>resources.$TYPE.$NAME.user</varname>
+    <varname>resources.$TYPE.$NAME.user</varname>,
+    <varname>resources.$TYPE.$NAME.servicePrincipal</varname>
     and <varname>resources.$TYPE.$NAME.password</varname>.
     </para></listitem>
   </varlistentry>
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -1075,6 +1075,64 @@ machine3.deployment.azure = {
   </para>
 </section>
 
+<section><title>Resource names and IDs</title>
+  <para>The best way to specify a reference to a resource that is managed by NixOps is
+  via <literal>resources.azure*.resourceName</literal>. However, if you need to reference
+  a resource not managed by NixOps, you can do so by resource name or ID.
+  </para>
+
+  <para>If the property description says "The name or resource of..." such
+  as for resource groups and storages, the reference is by name:
+<programlisting>
+  resources.azureResourceGroups.group = credentials // {
+    name = "my-test-group";
+    location = "West US";
+  };
+
+  resources.azureAvailabilitySets.set1 = {resources,...}: credentials // {
+    resourceGroup = resources.azureResourceGroups.group;
+    location = "West US";
+  };
+
+  resources.azureAvailabilitySets.set2 = credentials // {
+    resourceGroup = "my-test-group";
+    location = "West US";
+  };
+</programlisting>
+  </para>
+
+  <para>If the property description says "The Azure Resource Id or NixOps resource...",
+  the reference is by full Azure resource ID:
+<programlisting>
+  resources.azureVirtualNetworks.network = credentials // {
+    name = "test-network";
+    location = "West US";
+    addressSpace = [ "10.1.0.0/16" "10.4.0.0/16" ];
+    subnets = {
+      default.addressPrefix = "10.1.11.0/24";
+      GatewaySubnet.addressPrefix = "10.1.10.0/24";
+    };
+  };
+
+  resources.azureVirtualNetworkGateways.gateway1 = {resources,...}: credentials // {
+    location = "West US";
+    gatewayType = "RouteBased";
+    subnet.network = resources.azureVirtualNetworks.network;
+    subnet.name = "GatewaySubnet";
+  };
+
+  resources.azureVirtualNetworkGateways.gateway2 = credentials // {
+    location = "West US";
+    gatewayType = "RouteBased";
+    subnet.network = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups" +
+                     "/nixops-00000000-0000-0000-0000-000000000000-def-group/providers" +
+                     "/Microsoft.Network/virtualNetworks/test-network";
+    subnet.name = "GatewaySubnet";
+  };
+</programlisting>
+  </para>
+</section>
+
 <section><title>Azure limitations and quirks to be aware of:</title>
 
 <itemizedlist>
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -1101,7 +1101,7 @@ backendAddressPools = [ "default" ];
 
 <section><title>Managing virtual machines without public IP addresses</title>
   <para>If a virtual machine doesn't have a public IP address
-  (has obtainIP set to false), NixOps is unable to reach the
+  (has ip.obtain set to false), NixOps is unable to reach the
   SSH port of the machine and manage it.
   However, if you route the SSH port of the machine via an inboud NAT rule
   to a load balancer frontend interface that has a public IP address,
@@ -1125,7 +1125,7 @@ resources.azureLoadBalancers.lb = {resources,...}: credentials // {
 
 machine3.deployment.azure = {
   networkInterfaces.default ={
-    obtainIP = false;
+    ip.obtain = false;
     inboundNatRules = [{loadBalancer = resources.azureLoadBalancers.lb; name = "machine3-ssh";}];
   };
 };
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -835,6 +835,9 @@ in {
 <para>
   The deployment proceeds like this:
   <screen>
+$ nixops create ./load-balancer.nix ./load-balancer-azure.nix -d load-balancer-azure
+$ nixops deploy -d load-balancer-azure
+<replaceable>...</replaceable>
 def-group....................> creating Azure resource group 'nixops-71616e2e-c165-11e5-b910-b8e8560f8b58-def-group' in westus...
 dn-westus....................> creating Azure virtual network 'nixops-71616e2e-c165-11e5-b910-b8e8560f8b58-dn-westus' in westus...
 set..........................> creating Azure availability set 'nixops-71616e2e-c165-11e5-b910-b8e8560f8b58-set' in westus...
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -921,17 +921,61 @@ This terminates the Azure instances and deletes the alternative native load bala
 <section><title>Prequisites</title>
 <itemizedlist>
 
-  <listitem><para>You must supply the credentials(subscription ID, authority URL, user, password)
-  to your deployments via either <literal>*.subscriptionId</literal>,
-  <literal>*.authority</literal>, <literal>*.user</literal> and <literal>*.password</literal>
-  options or <literal>AZURE_SUBSCRIPTION_ID</literal>, <literal>AZURE_AUTHORITY_URL</literal>,
-  <literal>AZURE_USER</literal> and <literal>AZURE_PASSWORD</literal> environment variables.
-  Options take precedence over environment variables and are specified per resource/machine.
+  <listitem><para>You need Azure credentials to authenticate requests.
+  The authentication methods supported are user/password and service principal/password.
+  You need to ensure your Azure account has an Active Directory, and add a user or service principal to it.
+  Refer to these guides:
+  <link xlink:href="http://innerdot.com/azure/a-gaffers-guide-to-azure-introduction-and-getting-started">
+  creating an active directory and users</link> and
+  <link xlink:href="http://innerdot.com/azure/a-gaffers-guide-to-azure-service-principals-and-applications">
+  applications and service principals</link>.
   </para></listitem>
+
+  <listitem><para>You must supply the credentials(subscription ID, authority URL,
+  user or service principal, password) to your deployments via either
+  <literal>*.subscriptionId</literal>, <literal>*.authority</literal>, <literal>*.user</literal>,
+  <literal>*.servicePrincipal</literal> and <literal>*.password</literal>
+  options or <literal>AZURE_SUBSCRIPTION_ID</literal>, <literal>AZURE_AUTHORITY_URL</literal>,
+  <literal>AZURE_USER</literal>, <literal>AZURE_SERVICE_PRINCIPAL</literal>
+  and <literal>AZURE_PASSWORD</literal> environment variables.
+  Options take precedence over environment variables and are specified per resource/machine.
+  </para>
+
+  <para>You must specify either user or service principal but not both.
+  If either <literal>*.user</literal> or <literal>*.servicePrincipal</literal> is set,
+  both <literal>AZURE_USER</literal> and <literal>AZURE_SERVICE_PRINCIPAL</literal>
+  are ignored for the resource.</para>
+
+  <para>Example credentials for user/password authentication:
+<programlisting>
+credentials = {
+  subscriptionId = "00000000-0000-0000-0000-000000000000";
+  authority = "https://login.windows.net/YOURDIRECTORYNAME.onmicrosoft.com";
+  user = "user@YOURDIRECTORYNAME.onmicrosoft.com";
+  password = "**********";
+};
+</programlisting>
+  </para>
+
+  <para>Example credentials for service principal/password authentication:
+<programlisting>
+credentials = {
+  subscriptionId = "00000000-0000-0000-0000-000000000000";
+  authority = "https://login.windows.net/YOURDIRECTORYNAME.onmicrosoft.com";
+  servicePrincipal = "44444444-4444-4444-4444-444444444444";
+  password = "**********";
+};
+</programlisting>
+  </para>
+
+  <para>Authority URL can also be specified as
+  <literal>https://login.windows.net/TENANT_ID</literal>.</para>
+
+  </listitem>
 
   <listitem><para>You need to ensure that SSH ports of all machines are reachable
   either directly via machines' public IP addresses or via NAT rules on the public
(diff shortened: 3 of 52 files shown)


Phreedom:
Basic resources needed for bootstrapping and deployment are ready and well-tested.
Had lots of fun with locks, async operations and other stuff :)

Phreedom:
I tried to implement a fully-declarative workflow to build the
bootstrap image, upload, register it and spawn VMs, and almost succeded.
Unfortunately, Azure only supports "fixed" VHDs and there's no api to enlarge
the images which are already in the storage. So, it's very impractical since you
don't want an 1GB OS disk and you don't want to upload a 30GB blob. Of course
this isn't a problem if someone else creates an image, or you use the Azure client
which can convert sparse VHDs during the upload. Maybe I'll find a way to implement
something similar, but this is a low priority issue for now.

I suppose @rbvermaa is interested in this PR

Phreedom:
VMs are created and destroyed just fine. Data disks support is coming after a bit more testing.

Current blocker: it's impossible to create an empty azure deployment resource.
You can either create it with a VM inside, or you could provide some *.cspkg file which isn't
really documented and most tools are Windows-only and who knows azure might even have a check against empty deployments... I had spent countless hours looking for information.
Another great way to do it is create a deployment with a VM inside and then delete the VM :(
Or, VMs could create and delete it ad-hoc, but that's very ugly and breaks NixOps design.

Phreedom:
The current temporary approach is to create a deployment resource per VM, which is obviously broken. Additionally, this imposes a limit of 2 VMs max per Hosted Service resource.

The deployment resource is very important because it deals with load balancers and routing.

Routing is needed because Azure has very strict limits on the number of public IPs(and it costs $ too).
A good idea would be to use 1 IP per deployment with a port range mapped to ssh ports of all VMs.

I don't really expect anyone to chime in with a solution. Rather, I'm documenting the current state of affairs.

Phreedom:
rebasing

aycanirican:
I'll be happy to see this merged.

ericbmerritt:
@Phreedom whats the status here? Will it eventually get merged?

Phreedom:
yes

grahamc:
I'm also interested in this feature-set.

colemickens:
The example in the first post, is that using the Azure ASM endpoints? I'm trying to understand how you're using cert based auth (unless you provisioned an app + ServicePrincipal and that's the SP's certificate?)...

If it is ASM, do you have an example talking to the ARM endpoint? I thought I'd seen that stuff implemented in the PR when I looked previously.

Very interested to see this available as well. :+1:

Phreedom:
It's a very old example. ASM is 
gone :)
Just build the PR and take a 
look at the manual to get a 
more up-to-date example.

LiliDeng:
@Phreedom hi Phreedom, I got your update code for nixops deploy nixos against Azure. But hit one issue, could you please help on this? Thanks!
Below are my steps:

1.	Pull nixops from your github and check out nixops-master branch
2.	Run python setup.py install under the code path (need install many dependent python module)
3.	Run nixops create /root/nixops/examples/trivial-azure.nix -d azure
4.	Run nixops info -d azure to check status, hit below issue, I try to set $NIX_PATH to /root which I download the code into or /nix/var/nix/profiles/per-user/root/channels/nixos, both not work.
The file’s actual location is /root/nixops/nix/eval-machine-info.nix

warning: Nix search path entry ‘/nix/store/m3m0rrj9xjf5960cf04wl8w9xcah4p70-python-2.7.11/lib/python2.7/site-packages/nix’ does not exist, ignoring
error: file ‘nixops/eval-machine-info.nix’ was not found in the Nix search path (add it using $NIX_PATH or -I)
warning: evaluation of the deployment specification failed; status info may be incorrect

Network name: azure
Network UUID: 0b3dd0c2-e1e2-11e5-8274-00155d4c3e91
Network description: Unnamed NixOps network
Nix expressions: /root/nixops/examples/trivial-azure.nix

+------+--------+------+-------------+------------+
| Name | Status | Type | Resource Id | IP address |
+------+--------+------+-------------+------------+
+------+--------+------+-------------+------------+

Phreedom:
@LiliDeng

2: It's much easier to let nix build nixops rather than try doing it manually
3: the full "trivial" deployment example consists of both trivial.nix(makes the machine serve nix homepage via http) and trivial-azure.nix(describes how to deploy the machine to azure). using trivial-azure.nix alone would create a fully-functional machine which sits there doing nothing, which is ok too :)
```
export TEST_PATH=/some/place/to/store/stuff
mkdir -p $TEST_PATH
cd $TEST_PATH

# this is the branch that was used to build the azure images.
# using it greatly reduces the builds
git clone --branch revert-qemu https://github.com/Phreedom/nixpkgs.git 

# azure backend branch
git clone --branch nixops-master https://github.com/Phreedom/nixops.git

# build nixops
# creates "result" symlink that links to the fully-functional nixops build
cd nixops
NIX_PATH=$TEST_PATH nix-build release.nix -A build.x86_64-linux --exclude tarball

# create the "trivial" azure deployment example
NIX_PATH=$TEST_PATH result/bin/nixops create examples/trivial.nix examples/trivial-azure.nix -d azure

# now you can play with the deployment
# but this command and most others will fail unless you
# specify azure credentials(subscription ID etc)
# either in trivial-azure.nix or via env vars
NIX_PATH=$TEST_PATH result/bin/nixops info -d azure
```

Hope this helps!

LiliDeng:
@Phreedom It really did help me a lot, the issue is resolved, thank you. 
More question here about .nix files, since I want to create a NixOS based on nixops. I already use nixpkgs/nixos/maintainers/scripts/azure/create-azure.sh to build, upload vhd and create an nixos image. 

#1 Currently we only can deploy against ARM mode? or both ASM and ARM mode?
#2 For the authority, user and password, should I create an azure active domain under ARM mode? Why one use certificate doesn't work? 

Appreciate your help!

Phreedom:
@LiliDeng 

You don't have to build your custom image anymore. nixops creates the necessary resources(resource group, storage, container, image copy) to deploy vms automatically. Or maybe I misunderstood you.

1: Yes. ASM is gone for good. While I did manage to make it work well enough, nevertheless it was slow and supposedly ARM is the future.
2: Yes, the user/password combination should be created under ARM mode. User/password instead of certificates looked like it would be easier for users to handle and comments in the ADAL library code that I'm using for authentication says that certificate-based auth code is not well-tested.

LiliDeng:
@Phreedom Thanks again! Based on your first comment 'Currently you have to build and register a nixos bootstrap image by hand.' so I thought I need to upload the image by myself, so currently no need.

To create a NixOS VM in azure:
First I create Azure Active Domain under ARM mode?  Then create a user in the domain for filling below content.

let
  # change this as necessary or wipe and use ENV vars
  credentials = {
    subscriptionId = "00000000-0000-0000-0000-000000000000";
    authority = "https://login.windows.net/AUTHORITY.onmicrosoft.com";
    user = "user@AUTHORITY.onmicrosoft.com";
password = "**********"; 
};

Then I use your deployment sample in the first comment, can I deploy a new fresh NixOS in azure?
Please forgive me having so many questions since I am a newbie for NixOS.

Phreedom:
@LiliDeng Don't worry, you're doing just fine.

That comment was written 1 year ago, please ignore it :)

To deploy a simple vm, after following the steps in my previous comment
https://github.com/NixOS/nixops/pull/269#issuecomment-192525015
you need to add the credentials(subscriptionid, authority etc) to examples/trivial-azure.nix or use ENV vars. After that, deploying the vm should be as easy as
```
# deploy. takes some time
NIX_PATH=$TEST_PATH result/bin/nixops deploy -d azure

# see what IP was assigned to the machine
NIX_PATH=$TEST_PATH result/bin/nixops info -d azure

# check that the machine serves pages over HTTP
curl $MACHINE_IP

# ssh into the machine
NIX_PATH=$TEST_PATH result/bin/nixops ssh -d azure machine
```

LiliDeng:
@Phreedom I will have a try, maybe will bother you again if hit any blocker, thanks so much, you are so nice!

colemickens:
```
error: Package ‘python2.7-pyjwt-1.4.0’ in ‘/nixpkgs/pkgs/top-level/python-packages.nix:16788’ is not supported on ‘x86_64-darwin’, refusing to evaluate.
```

Is NixOps already unbuildable on Darwin, or is this unique to your branch, @phreedom ?

colemickens:
`nix-build release.nix -A build.x86_64-linux` failed as well:

https://gist.github.com/e6dec7c8da5209fb3447

Phreedom:
On Monday, March 07, 2016 16:36:32 Cole Mickens wrote:
> ```
> error: Package ‘python2.7-pyjwt-1.4.0’ in
> ‘/nixpkgs/pkgs/top-level/python-packages.nix:16788’ is not 
supported on
> ‘x86_64-darwin’, refusing to evaluate. ```
> 
> Is NixOps already unbuildable on Darwin, or is this unique to 
your branch,
> @phreedom ?

This is a dependency of ADAL. Didn't expect a simple auth 
code to be such a PITA. I have no way to debug and fix this, 
and this is how Apple wants it to be :(

Phreedom:
On Monday, March 07, 2016 17:04:18 Cole Mickens 
wrote:
> `nix-build release.nix -A build.x86_64-linux` failed 
as well:
> 
> https://gist.github.com/e6dec7c8da5209fb3447

Are you trying to cross-build on darwin? This error 
doesn't look like it's something that would be hard 
to fix, but again I have no way to do it.

colemickens:
No, that's on my NixOS box. First error was when I was doing `nix-build release.nix -A build`. Second error was when building `x86_64-linux` from my 64bit NixOS machine.

When I pulled your branch a few weeks ago, I encountered the first error. The second error, however, is new.

LiliDeng:
I didn't hit this issue when execute below command in my NixOS VM. (NixOS 15.09.784.7b85b8a)
export TEST_PATH=/some/place/to/store/stuff
mkdir -p $TEST_PATH
cd $TEST_PATH

# this is the branch that was used to build the azure images.
# using it greatly reduces the builds
git clone --branch revert-qemu https://github.com/Phreedom/nixpkgs.git 

# azure backend branch
git clone --branch nixops-master https://github.com/Phreedom/nixops.git

# build nixops
# creates "result" symlink that links to the fully-functional nixops build
cd nixops
NIX_PATH=$TEST_PATH nix-build release.nix -A build.x86_64-linux --exclude tarball

LiliDeng:
I didn't hit issue when execute below commands in my NixOS VM (NixOS 15.09.784.7b85b8a)

export TEST_PATH=/some/place/to/store/stuff
mkdir -p $TEST_PATH
cd $TEST_PATH
git clone --branch revert-qemu https://github.com/Phreedom/nixpkgs.git 
git clone --branch nixops-master https://github.com/Phreedom/nixops.git
cd nixops
NIX_PATH=$TEST_PATH nix-build release.nix -A build.x86_64-linux --exclude tarball

LiliDeng:
@Phreedom Bother you again, I create an azure active directory in our subscription, and add a new user in the organization, use this user to login in the portal, it will need to change the temporary password into a fix one. 
Fill these values into trivial-azure.nix, the run nixops deploy command hit 400 error, when I use the new created account to login the portal, there is not any subscription associated with it. Is it possible you can figure out the reason from my description?

segment in trivial-azure.nix
       authority = "https://login.windows.net/name.onmicrosoft.com";
       user = "lilideng@name.onmicrosoft.com";
       password = "password";

error: Auth failure: Error:Error:Error:Get Token request returned http error: 400 and server response: {"error":"invalid_grant","error_description":"AADSTS70002: Error validating credentials. AADSTS50126: Invalid username or password\r\nTrace ID: 89e69d84-6989-443c-b329-580530dcf00c\r\nCorrelation ID: a772446c-0c76-4c80-9495-8b17472fd46a\r\nTimestamp: 2016-03-08 01:39:05Z","error_codes":[70002,50126],"timestamp":"2016-03-08 01:39:05Z","trace_id":"89e69d84-6989-443c-b329-580530dcf00c","correlation_id":"a772446c-0c76-4c80-9495-8b17472fd46a"} token_response:{u'timestamp': u'2016-03-08 01:39:05Z', u'trace_id': u'89e69d84-6989-443c-b329-580530dcf00c', u'correlation_id': u'a772446c-0c76-4c80-9495-8b17472fd46a', u'error_description': u'AADSTS70002: Error validating credentials. AADSTS50126: Invalid username or password\r\nTrace ID: 89e69d84-6989-443c-b329-580530dcf00c\r\nCorrelation ID: a772446c-0c76-4c80-9495-8b17472fd46a\r\nTimestamp: 2016-03-08 01:39:05Z', u'error': u'invalid_grant', u'error_codes': [70002, 50126]} token_response:None token_response:None

Phreedom:
On Monday, March 07, 2016 17:25:50 Cole Mickens wrote:
> No, that's on my NixOS box. First error was when I was doing `nix-build
> release.nix -A build`. 

Ok. So you tried building for all architectures, but it doesn't change the situation: 
darwin build is broken. I don't know why the pyjwt has only linux in the supported 
platforms in NixOS. Most likely, it's simply because the packager was careful and 
honest, and listed only the arch they tested on. But I still have no way to test what 
happens if I lift the restriction.

> Second error was when building `x86_64-linux` from
> my 64bit NixOS machine.

Is it running an older stable revision? It's quite surprising since the second error is 
due to gce/libcloud and not my azure work, and gce backend was supposedly ported 
and tested on all NixOS branches.

> When I pulled your branch a few weeks ago, I encountered the first error.
> The second error, however, is new.

colemickens:
Sure, I understand the darwin failure -- I'm no longer worried about it (though if NixOps no longer works on Darwin, then it seems like the platform should be removed from the build definitions in `release.nix`).

The second error was with revision 22b3bb862339f708c1fed02b334424644f3ac547 of your branch of `nixops-master` the latest from nixos-channel's `nixos-unstable`. I'll try to dig into this more this evening. (I'm `meloc` on #nixos if you want to ping me.)

Phreedom:
@LiliDeng

> @Phreedom Bother you again, I create an azure active directory in our
> subscription, and add a new user in the organization, use this user to
> login in the portal, it will need to change the temporary password into a
> fix one. Fill these values into trivial-azure.nix, the run nixops deploy
> command hit 400 error, when I use the new created account to login the
> portal, there is not any subscription associated with it. Is it possible
> you can figure out the reason from my description?

I'm assuming you're using the new portal: https://portal.azure.com/

Subscriptions->pick the subscription->Manage->Users-> You should be able to see your 
newly-created user and see the role it's assigned. The role should be something "powerful" 
like owner or contributor.

Also it should be possible to see the user list and subscription if you login via azure-cli:
```
azure account list
# shoud show something like:
# Visual Studio Ultimate with MSDN  SUBSCRIPTION_ID  true     Enabled

azure ad user list
# should include in its output
# data:    Object Id:       *****
# data:    Principal Name:  lilideng@name.onmicrosoft.com
# data:    Display Name:    lilideng
# data:    E-Mail:          lilideng@name.onmicrosoft.com
```

> segment in trivial-azure.nix
>        authority = "https://login.windows.net/name.onmicrosoft.com";
>        user = "lilideng@name.onmicrosoft.com";
>        password = "password";
> 
> error: Auth failure: Error:Error:Error:Get Token request returned http
> error: 400 and server response:
> {"error":"invalid_grant","error_description":"AADSTS70002: Error validating
> credentials. AADSTS50126: Invalid username or password\r\nTrace ID:
> 89e69d84-6989-443c-b329-580530dcf00c\r\nCorrelation ID:
> a772446c-0c76-4c80-9495-8b17472fd46a\r\nTimestamp: 2016-03-08
> 01:39:05Z","error_codes":[70002,50126],"timestamp":"2016-03-08
> 01:39:05Z","trace_id":"89e69d84-6989-443c-b329-580530dcf00c","correlation_i
> d":"a772446c-0c76-4c80-9495-8b17472fd46a"} token_response:{u'timestamp':
> u'2016-03-08 01:39:05Z', u'trace_id':
> u'89e69d84-6989-443c-b329-580530dcf00c', u'correlation_id':
> u'a772446c-0c76-4c80-9495-8b17472fd46a', u'error_description':
> u'AADSTS70002: Error validating credentials. AADSTS50126: Invalid username
> or password\r\nTrace ID:
> 89e69d84-6989-443c-b329-580530dcf00c\r\nCorrelation ID:
> a772446c-0c76-4c80-9495-8b17472fd46a\r\nTimestamp: 2016-03-08 01:39:05Z',
> u'error': u'invalid_grant', u'e rror_codes': [70002, 50126]}
> token_response:None token_response:None

This error message means that both name.onmicrosoft.com AD and lilideng user exist. You 
can only get this error if you specify a wrong password.

Phreedom:
On Monday, March 07, 2016 18:39:29 Cole Mickens wrote:
> Sure, I understand the darwin failure -- I'm no longer worried about it
> (though if NixOps no longer works on Darwin, then it seems like the
> platform should be removed from the build definitions in `release.nix`).

I'd rather lift the arch restriction on pyjwt, assuming someone ever tested 
nixops on darwin, if anything is broken it's darwin+azure combo and not 
all of nixops and we should wait for a tester to show up...

> The second error was with revision 
22b3bb862339f708c1fed02b334424644f3ac547
> of your branch of `nixops-master` the latest from nixos-channel's
> `nixos-unstable`. I'll try to dig into this more this evening. (I'm `meloc`
> on #nixos if you want to ping me.)

I managed to replicate this by building nixops against the current master 
instead of the base branch for azure images. It's a recent "improvement" :)

LiliDeng:
@Phreedom I need add this account as co-adminstrator for the subscription, right? Seems our test sub only support microsoft account

Phreedom:
On Monday, March 07, 2016 21:42:36 LiliDeng 
wrote:
> @Phreedom I need add this account as co-
adminstrator for the subscription,
> right? Seems our test sub only support 
microsoft account

Co-admin definitely should have enough 
permissions

LiliDeng:
@Phreedom do we support Service Principals auth? I follow below steps, only difference is I can't add the new created account into co-admin.
http://innerdot.com/azure/a-gaffers-guide-to-azure-introduction-and-getting-started
http://innerdot.com/azure/a-gaffers-guide-to-azure-service-principals-and-applications

Phreedom:
On Monday, March 07, 2016 22:15:44 LiliDeng wrote:
> @Phreedom do we support Service Principals auth? I follow below steps, only
> difference is I can't add the new created account into co-admin.
> http://innerdot.com/azure/a-gaffers-guide-to-azure-introduction-and-getting
> -started
> http://innerdot.com/azure/a-gaffers-guide-to-azure-service-principals-and-a
> pplications

Ok. So it turns out that the user/pass authentication for users requires calling a different 
ADAL function than for the user/pass authentication for service principals :/

So the code as committed only supports users, although I have local changes which 
break users but allow service principal auth.

The fix is trivial, but I'm wondering how to make it user friendly. 4 credentials is already 
intimidating, and adding 5th along the lines of isServicePrincipal = true/false would be 
quite an overkill.

Just how good/bad idea is it to tell apart users from service principals by the presense of 
'@'? Users supposedly always look like "user@tenant.onmicrosoft.com" while service 
principals are in GUID format. Any Microsofties can comment on this?

I will be committing this approach as a temporary fix in a few minutes.

And it looks like adding a section on authentication to the manual is a good idea.

Phreedom:
OK just checked and "user name" for 
application works just fine in 
"app_id@tenant.onmicrosoft.com" format, so 
you can't tell them apart :(

You really have to know whether you are 
dealing with a user or application to invoke 
the right ADAL function :/

LiliDeng:
@Phreedom Thanks much, let me check then~

LiliDeng:
@Phreedom Does it mean I have to add the user into co-admin?

Phreedom:
I have committed the patch 
I promised but it doesn't 
look like a correct solution 
to me ;(

Phreedom:
On Tuesday, March 08, 2016 00:01:01 
LiliDeng wrote:
> @Phreedom Does it mean I have to add 
the user into co-admin?

 Contributor role should be sufficient but 
co-admin of course works too

LiliDeng:
@Phreedom finnally, I am seeing the resource is created... Thanks very much!!! I use below credentials.

  credentials = {
    subscriptionId = "00000000000000000000000";
    authority = "https://login.windows.net/tenantid";
    user = "appid";
    password = "password";
  };

LiliDeng:
@Phreedom finally, I am seeing the resource is created... Thanks very much!!! I use below credentials.
credentials = {
subscriptionId = "00000000000000000000000";
authority = "https://login.windows.net/tenantid";
user = "appid";
password = "password";
};

LiliDeng:
@Phreedom The VM is created successfully, if not set the username and password what is the default credential to login to the VM? If want to set it by myself, what section should I add?

Phreedom:
@LiliDeng 
NixOps automatically manages root ssh keys, so you can log into a machine "machine" using
```
result/bin/nixops ssh -d azure machine
```

But if you want to add another user, you add something like this to the machine definition:
```
    users.extraUsers.mytestuser2 = {
      initialPassword = "passsssss";
      useDefaultShell = true;
    };
```

LiliDeng:
@Phreedom amazing.... I can connect it now. Thanks!

LiliDeng:
@Phreedom  by the way, is there any document for the nix configuration files (and the sections for azure)?

colemickens:
@LiliDeng If you `nix-build release.nix` then you'll get a result (or result-2) directory that contains the generated manual. It includes documentation for Azure.

@Phreedom How do I change the base disk used? The 2GB VHD is too small for me and I have found a way to upload large fixed-size VHDs without having to upload every byte...

Phreedom:
@LiliDeng http://nixos.org/nixos/manual and result/share/doc/nixops/manual.html

LiliDeng:
@Phreedom thanks, hit one issue when run nixops deploy, I can ssh to the VM and found the waagent is running version is 2.0.16, what is the impact of the below error?

fatal: reference is not a tree: 9dba81c7b1239c7971ec96e405e403c7cd224e6b
Unable to checkout 9dba81c7b1239c7971ec96e405e403c7cd224e6b from https://github.com/Phreedom/WALinuxAgent.git.
builder for ‘/nix/store/f5cizfqx02hn16gilgaazd6np6yg38j5-WALinuxAgent-9dba81c.drv’ failed with exit code 1
fetching path ‘/nix/store/k1kab609hn757bxswvrgw72sbs28jhl1-libxcb-1.11.1’...
cannot build derivation ‘/nix/store/gfnkp6r0jnknyn243fxfa1paknwgbd3a-waagent-2.0.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/0l6yi1irwsjgxsw8rlmyc0y8zxpnzs1d-udev-rules.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/iq7s2r5v2a1kqz7iv04m5g8xbnpmf946-unit-waagent.service.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/43xpfc7jfpq0srcfbpwi19vfvq8s8zl0-etc.drv’: 1 dependencies couldn't be built
killing process 4781
cannot build derivation ‘/nix/store/g5hij69jiqnp41z8d8qdhairxqg62v12-nixos-system-machine-16.03pre-git.drv’: 1 dependencies couldn't be built
cannot build derivation ‘/nix/store/qxmq1j1jx19avzm6rsxklgfs69zz092x-nixops-machines.drv’: 1 dependencies couldn't be built
error: build of ‘/nix/store/qxmq1j1jx19avzm6rsxklgfs69zz092x-nixops-machines.drv’ failed
error: unable to build all machine configurations

Phreedom:
@colemickens This is a temporary situation. As soon as @rbvermaa find some time to finalize this, I'll transfer nixos images storage to him and the default image could be made larger. It was totally impractical to debug 30GB images.

If you want to run your own images, you could deploy the default resource group, storage and container,  upload your custom image into the container and specify the vhd file name(name only, not path or blob url) via machine.deployment.azure.ephemeralDiskContainer

If you have a dedicated storage for it:
```
resources.azureBlobs.boot-test =  {resources, nodes, ...}: {
    # steals the default container resource automatically created by NixOps
    container = nodes.$MACHINE_NAME.azure.ephemeralDiskContainer;
    blobType = "PageBlob";
    # this is the "official" nixos image blob as of today, replace with your own
    copyFromBlob = "https://nixos.blob.core.windows.net/images/nixos-unstable-nixops.vhd";
};

MACHINE_NAME.deployment.azure.rootDiskImageBlob = resources.azureBlobs.boot-test;
```

Phreedom:
@colemickens actually, I've just implemented a much better approach, which works seamlessly even for more complex examples like geodistributed-azure.nix:
```
nixops set-args --argstr azure-image-url BOOT://IMAGE/URL
```

Phreedom:
@LiliDeng ah I forgot I rebased 
my agent branch before merging. 
I have pushed the fix to my 
revert-qemu branch.

LiliDeng:
@Phreedom let me try again, thanks!

LiliDeng:
@Phreedom activation finished successfully
But during deploy,it need me to input 'yes' to ssh the machine, and finially it ssh successfully or not? I can't tell from the log.

The authenticity of host '104.40.85.X (104.40.85.X)' can't be established.
ED25519 key fingerprint is SHA256:goXDq2RKYPUIsF4svziX5PstUd1KPUwENDgXVeWf2Lk.
Are you sure you want to continue connecting (yes/no)? y
Please type 'yes' or 'no': yes
Warning: Permanently added '104.40.85.X' (ED25519) to the list of known hosts.
ssh_packet_read: Connection closed
machine......................> could not connect to ‘root@104.40.85.195’, retrying in 4 seconds...
machine......................> setting state version to 16.03
machine......................> waiting for SSH...

Phreedom:
@LiliDeng

Transient SSH failures are normal since waagent restarts ssh, timeouts happen etc. On 
the other hand, SSH keys should have been added to .ssh/known_keys automatically. Did 
you use your own image instead of the default one? Can you check if waagent log says 
anything about waagent regenerating host keys? Or anything else happened than a 
straight process from nixops deploy to finish?

LiliDeng:
@Phreedom I didn't use my own image, seems it used the latest one from unstable build start from 16.X version; waagent didn't have the issue about host keys; finally it deploy successfully and I can use result/bin/nixops ssh -d azure machine to ssh into the VM

Phreedom:
@LiliDeng It's nearly impossible to abort 
nixops deployment exactly between the 
moment NixOps marks the VM as created 
and the moment it adds the key to 
.ssh/known_keys and writes "Got IP: 
xxxxx" which is the only way the keys can 
end up not getting added to 
known_keys...

LiliDeng:
@Phreedom only error in waagent.log is 

2016/03/09 09:36:27 ERROR:unicode argument without an encoding occured creating /var/lib/waagent/CustomData

LiliDeng:
@Phreedom 
1 Only error in waagent.log is 
   2016/03/09 09:36:27 ERROR:unicode argument without an encoding occured                  creating /var/lib/waagent/CustomData

2 Suggest that increase the /dev/sda size, it only has 2 GB, I got below error when install vim, I need to update some nix file to reset it?
   error: unable to fork: Cannot allocate memory

3 When destroy the resrouce, must I interact with it by type 'y'? Actually if delete the resource group, all of the resource in it will be deleted.

Phreedom:
@LiliDeng
> 1 Only error in waagent.log is
>    2016/03/09 09:36:27 ERROR:unicode argument without an encoding occured   
>               creating /var/lib/waagent/CustomData

This doesn't seem to be a problem.

> 2 Suggest that increase the /dev/sda size, it only has 2 GB,

Yes, as soon as this is merged, it makes sense to increase the root disk image size to 
the standard 30GB. But it's very impractical to debug images this large if there's no 
way to efficiently upload sparse VHDs

> I got below
> error when install vim, I need to update some nix file to reset it? error:
> unable to fork: Cannot allocate memory

Looks like your instance is out of RAM.


> 3 When destroy the resrouce, must I interact with it by type 'y'?

You can pass use --confirm option instead

> Actually if delete the resource group, all of the resource in it will be deleted.

Yes, but this is azure-specific.

LiliDeng:
@Phreedom thanks for your patience to solve all of my issues with technical detail.

colemickens:
With a nixos base image built from current `nixos-unstable`, I'm seeing two issues:

1) Host SSH Key Errors

I'm getting host ssh key errors on all (seven out of seven) deployment attempts. See below, I link to logs.

2) Machine does not return after shutdown/startup from Portal
I rebooted the instance via ssh and it remains accessible. However, if you use the Portal to "Stop" the machine, and then "Start" using the Portal, the machine does not come back up properly. Or at least wasn't reachable for me. This is because when it is Stopped and Started in the portal, the public ip address changes.

The bug here is that the public ip you're creating is dynamic (make it static and this will fix), and doesn't have a DNS name. If it had a DNS name and you pointed at that, you would be resilient to the public ip address being dynamic. (But will be at the mercy of DNS TTLs.) I recommend doing both - making the IP address be fixed and setting a dns name on the public ip address for convenience.

---

I have a couple of scripts here that nearly fully automate the creation and upload of the VHD, as well as a script to do the actual nixops deployment. Also contained in there are the logs from the deployment attempts. https://github.com/colemickens/azure-toolkit/tree/master/nixos/

(Also, it uses `azure-vhd-utils-for-go` which does a minimal upload, even for a large disk. I am uploading 30GB images, but it's only uploading the 800MB or so that is actually occupied.)

colemickens:
Here's `journalctl` from one of them: https://gist.githubusercontent.com/colemickens/a43c5d8a0ae35d3e0e34/raw/685d960472dc377ff7f1a1d26b8e8bc36aa15680/gistfile1.txt

Ctrl-F: "Generating public/private rsa key pair.". You can see it appears multiple times - it looks like it's restarting `sshd` and then it regenerates the pair again. Maybe the sshd-pre-start wrapper has changed?

colemickens:
I'm guessing it's this: https://github.com/Azure/WALinuxAgent/blob/2.0/waagent#L6410

LiliDeng:
the same issue happen can't ssh vm after shutdown, then start from portal.

colemickens:
Fix for the first issue. This is the issue of the public ip not being static. This causes the VM to not be ssh-able after Stop and Start from Portal. Suggested fix: https://github.com/colemickens/nixops/commit/d7df58cdfc88e5c6a2ea3e4c52f6553c11803bff

Fix for the second issue. This is the issue that causes the ssh host key errors that  cause deployments or sshing to fail, depending on when the agent finishes provisioning. The fix is to disable ssh key regeneration. The [real issue](https://github.com/Azure/WALinuxAgent/issues/162) is the agent is overzealous about deleting keys.  Suggested fix: https://github.com/colemickens/nixpkgs/commit/38166cffc55112cfe52d398202b2346448d79681 

Things seems to work as expected with these changes.

Phreedom:
On Thursday, March 10, 2016 13:28:35 Cole Mickens wrote:
> With a nixos base image built from current `nixos-unstable`, I'm seeing two
> issues:
> 
> 1) Host SSH Key Errors
> 
> I'm getting host ssh key errors on all (seven out of seven) deployment
> attempts. See below, I link to logs.

This happens because of Provisioning.RegenerateSshHostKeyPair=y  in 
nixos/modules/virtualisation/azure-agent.nix

The default nixops image is generated with n, but for "standalone" version y is more 
appropriate. I don't really have a pretty solution to this.

Most likely, I'm going to move entropy consumption code out of WALA and into a 
separate systemd job and run it before sshd starts, and disallow key regeneration.


> 2) Machine does not return after shutdown/startup from Portal
> I rebooted the instance via ssh and it remains accessible. However, if you
> use the Portal to "Stop" the machine, and then "Start" using the Portal,
> the machine does not come back up properly. Or at least wasn't reachable
> for me. This is because when it is Stopped and Started in the portal, the
> public ip address changes.
> 
> The bug here is that the public ip you're creating is dynamic (make it
> static and this will fix), and doesn't have a DNS name. If it had a DNS
> name and you pointed at that, you would be resilient to the public ip
> address being dynamic. (But will be at the mercy of DNS TTLs.) I recommend
> doing both - making the IP address be fixed and setting a dns name on the
> public ip address for convenience.

This is strange. Nixops stop and start don't cause stuff like this. Is this because the portal 
stop and deallocates the vm instead of only stopping it?

Either way, if VM IP changes, you can re-fetch it with deploy --check

> 
> ---
> 
> I have a couple of scripts here that nearly fully automate the creation and
> upload of the VHD, as well as a script to do the actual nixops deployment.
> Also contained in there are the logs from the deployment attempts.
> https://github.com/colemickens/azure-toolkit/tree/master/nixos/
> 
> (Also, it uses `azure-vhd-utils-for-go` which does a minimal upload, even
> for a large disk. I am uploading 30GB images, but it's only uploading the
> 800MB or so that is actually occupied.)

I'll take a look, thanks!

Phreedom:
On Thursday, March 10, 2016 13:51:37 Cole Mickens wrote:
> Here's `journalctl` from one of them:
> 
https://gist.githubusercontent.com/colemickens/a43c5d8a0ae35d3e0e34/raw/6
85
> d960472dc377ff7f1a1d26b8e8bc36aa15680/gistfile1.txt
> 
> Ctrl-F: "Generating public/private rsa key pair.". You can see it appears
> multiple times - it looks like it's restarting `sshd` and then it
> regenerates the pair again. Maybe the sshd-pre-start wrapper has changed?

It generates the RSA key, but the key type used by nixops and configured for 
wala is ed25519. SSHD is configured to have both key types. So probably the 
1st key generation happens on SSHD start since nixops doesn't provision the 
RSA key, then WALA wipes all host keys and generates only the ed25519 one, 
the pre-start script fixes the lack of RSA host key again.

Phreedom:
@LiliDeng

> the same issue happen can't ssh vm 
after shutdown, then start from portal.

Run deploy --check to fetch the new IP.

Phreedom:
On Friday, March 11, 2016 00:22:33 Cole Mickens wrote:
> Fix for the first issue. This is the issue of the public ip not being
> static. This causes the VM to not be ssh-able after Stop and Start from
> Portal. Suggested fix:
> https://github.com/colemickens/nixops/commit/d7df58cdfc88e5c6a2ea3e4c52f655
> 3c11803bff

Probably it's going to have some billing consequences... I think I'll make it configurable 
with static as default.

> Fix for the second issue. This is the issue that causes the ssh host key
> errors that  cause deployments or sshing to fail, depending on when the
> agent finishes provisioning. The fix is to disable ssh key regeneration.
> The [real issue](https://github.com/Azure/WALinuxAgent/issues/162) is the
> agent is overzealous about deleting keys.  Suggested fix:
> https://github.com/colemickens/nixpkgs/commit/38166cffc55112cfe52d398202b23
> 46448d79681
> 
> Things seems to work as expected with these changes.

But as explained above, this image is not a good candidate for "standalone" mode, that 
is when you use azure-cli or such to create an image and play with it.

LiliDeng:
@Phreedom  ssh worked after run nixops check, shutdown -> start, the VM will provision again. 

NIX_PATH=$TEST_PATH result/bin/nixops check -d azure
machine......................> warning: Azure machine 'nixops-5b349c45-e5d9-11e5-8858-00155d4c3e91-machine' public ipv4 has changed to '137.135.43.240'; expected it to be '40.118.131.196'
+---------+--------+-----+-----------+----------+----------------+-------+-------+
| Name    | Exists | Up  | Reachable | Disks OK | Load avg.      | Units | Notes |
+---------+--------+-----+-----------+----------+----------------+-------+-------+
| machine | Yes    | Yes | Yes       | Yes      | 0.12 0.11 0.08 |       |       |
+---------+--------+-----+-----------+----------+----------------+-------+-------+

Phreedom:
@LiliDeng

> @Phreedom  ssh worked after run nixops check, 
shutdown -> start, the VM will
> provision again.

Good, so I'll add an IP check on VM start which will 
make things just work if started from NixOps. Stop & 
deallocate + start from panel will still require deploy --
check which is ok. Need to document this too.

colemickens:
Hi @Phreedom thanks for the replies! One question - why is it desirable to have the agent regenerate the keys in "standalone" mode? Since it just blindly calls openssl the same way the sshd-pre-start does, why not just fully let sshd-pre-start take care of it?

Phreedom:
On Friday, March 11, 2016 11:22:52 Cole Mickens wrote:
> Hi @Phreedom thanks for the replies! One question - why is it 
desirable to
> have the agent regenerate the keys in "standalone" mode? Since it 
just
> blindly calls openssl the same way the sshd-pre-start does, why not 
just
> fully let sshd-pre-start take care of it?

VMs tend to not have enough entropy available and this affects the 
strength of the keys generated. This is one of the reasons why NixOps 
generates the key on your local physical(entropy-rich) machine and 
why WALA fetches some entropy from the hypervisor before 
regenerating the keys. My plan is to move the entropy step out of 
WALA and execute it before sshd has the chance to generate the keys.

Phreedom:
I have removed the distinction between nixops and "standalone" images, packaged azure-vhd-tools-for-go, updated image generation and upload scripts for maintainers, pushed this into nixos master, regenerated the default image and tested it. The base commit for the current image is 158d723812e7364d8cadd7802d0b8d239fb227d8 .

colemickens:
This has been working for me in my limited testing. Thanks @Phreedom!

Phreedom:
*@colemickens [1]*Thanks for your feedback. 
It was really helpful. Hopefully, this backend 
is going to be even better than the GCE one :)

--------
[1] https://github.com/colemickens

edolstra:
@Phreedom This broke NixOps eval on Nixpkgs 15.09:
```
$ nix-instantiate release.nix -A build.x86_64-linux -I nixpkgs=https://nixos.org/channels/nixos-15.09/nixexprs.tar.xz
error: cannot coerce null to a string, at /nix/store/hkmhb6ifd76jpkm8gs9xvpykqmjcb9qf-nixexprs.tar.xz/lib/types.nix:217:24
```
It works with 16.03. Any idea where that error comes from?

Phreedom:
On Wednesday, April 06, 2016 09:55:09 Eelco Dolstra wrote:
> @Phreedom This broke NixOps eval on Nixpkgs 15.09:
> ```
> $ nix-instantiate release.nix -A build.x86_64-linux -I
> nixpkgs=https://nixos.org/channels/nixos-15.09/nixexprs.tar.xz error:
> cannot coerce null to a string, at
> /nix/store/hkmhb6ifd76jpkm8gs9xvpykqmjcb9qf-
nixexprs.tar.xz/lib/types.nix:2
> 17:24 (use ‘--show-trace’ to show detailed location information)
> ```
> It works with 16.03. Any idea where that error comes from?
> 

nix/azure-blob-container.nix: type = types.enum [ null "container" "blob" ] ;

Apparently enum was generalized to handle non-string values after 15.09.

I didn't port/cherry-pick anything into 15.09.

Issue #221, opened 2014-08-20
Option to use private ip address with EC2


joamaki:
Often one wants to block SSH on the public IP address of the EC2 instance and only
expose SSH on the internal IP. It would be great if NixOps could fallback to, or prefer, the private IP
of the target when deploying.

cwp:
If this isn't implemented, why does the deployment.ec2.associatePublicIpAddress option exist? If it's set to false (which is the default), nixops can't connect to the instance to configure it.

mbbx6spp:
@cwp I can be on a VPN that is connected to a igw subnet and want to provision and deploy to instances in there. You shouldn't need to have a public IP address to deploy to those instances if you can connect to it. HTH.

mbbx6spp:
If this is still a work in progress, I would like to attempt it as I have this very need right now. Let me know if anyone else is actively working on it. Thanks.

mbbx6spp:
Sorry @cwp I think you were asking a slightly different question. You should use the `deployment.ec2.associatePublicIpAddress` option if the instances in your VPC subnet that you are deploying needed a public IP to be associated with it in a `vgw` subnet. Hope that makes sense as to the difference in request.

mbbx6spp:
I just looked into doing this and it seems we might need a slight change to the backend API to support this. Honestly, it seems like it would be a useful API change for a number of things anyway. But I wanted to run it by people here to see what they think.

I added a `deployment.ec2.usePrivateIpAddress` option in Nix and read into the Python EC2Definition class and then in the `EC2State` we have the method `get_ssh_name` which currently only takes `self` as an argument. It would need the definition passed into it to be able to do what we need above. This requires a change to the backend API as a whole (across all backends). I didn't want to go in and just do this without gathering thoughts from others on this.

Cheers.

cwp:
@mbbx6spp The problem is that nixops will only connect to the public IP address via SSH. If deployment.ec2.associatePublicIpAddress is false, then it will create an instance and then sit there endless printing out "waiting for IP address...". It's waiting for the instance metadata to have a public IP address, which will never happen. 

AFAICT, deployment.ec2.associatePublicIpAddress must be set to true, or deployment will always fail.

mbbx6spp:
[...]

PR #287, opened 2015-04-24
Add support for ssh connections via a private ip.

andrewlmurray:
Edit: 

It seems there has been some discussion on this issue in NixOS/nixops#221. I seem to have implemented it as was being discussed there. 

This fixes the behavior of ec2.associatePublicIpAddress which
cannot safely be set to false when deploying inside a subnet before this commit.

Add a new option ec2.usePrivateIpAddress which will default to
true in the case that associatePublicIpAddress is false and you are
deploying to a subnet.

Add supporting code in the ec2 backend to support this option. This required
storing some state from the defn in the machine state, but it seems that's already
being done for other things like the spot instance price so this seems consistent.

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -27,6 +27,16 @@ let
     let props = config.deployment.ec2.physicalProperties;
     in if props == null then false else (props.allowsEbsOptimized or false);
 
+  defaultUsePrivateIpAddress =
+    let
+      assocPublicIp = config.deployment.ec2.associatePublicIpAddress;
+      subnetId = config.deployment.ec2.subnetId;
+    in
+      if assocPublicIp == false && subnetId != "" then
+        true
+      else
+        false;
+
   commonEC2Options = import ./common-ec2-options.nix { inherit lib; };
 
   ec2DiskOptions = { config, ... }: {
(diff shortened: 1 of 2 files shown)


andrewlmurray:
Err hold off on pulling - it seems that while this works for creation I'm getting some key errors post creation. Investigating.

andrewlmurray:
Ok - sorry I missed that the boxes were being re-keyed later after host creation so I assumed that nixops ssh would work after creation. 

I've pushed a commit which fixes this problem.

andrewlmurray:
@edolstra @rbvermaa 
Is there anything in particular I need to do to get this merged in?

rbvermaa:
@andrewlmurray I haven't looked at this yet as I just got back from holiday. Will look at it tomorrow.

andrewlmurray:
Well there is the reason they are named differently then there is the issue of what name I should use for the public one. 

The reason they are different is that you often will want a public ip address assigned but to use the private ip address for provisioning. (very typically you don't want to expose your ssh port to the public internet at all, but the boxes you are deploying to want public ip addresses as other ports (80 for example) may be open to the public internet. 

Does this make sense?

That being said there is a different name for the internal state var and nix attr-key which sets it. (associate_public_ip_address vs wants_public_ip_address) It would probably be more clear just to use the same name in both places. I'll change that.

andrewlmurray:
@rbvermaa ok updated

rbvermaa:
Thanks for your explanation.

PR #292, opened 2015-05-07
Use the right SSH port when setting up ssh tunnels

basvandijk:
This commit allows to use `nixops ssh` for machines which run sshd on a non-standard port.


basvandijk:
Closing this PR since this seems to work now in master.

Issue #289, opened 2015-04-27
/etc/hosts sets up the local "-encrypted" name as 127.0.0.1


bruceadams:
This is a problem for software that binds to a specific interface. I want the local "-encrypted" name to be bound to the local encrypted tunnel network interface, not localhost. I don't see any host name setup for the local encrypted tunnel network interface.

My specific example is `zookeeper`. On a machine `zk0`, `nixops` sets up the host name `zk0-encrypted` as 127.0.0.1. The problem comes from `zookeeper` binding to `zk0-encrypted`, which is one of the hosts named in `zoo.cfg`. Since `zookeeper` binds to 127.0.0.1, remote hosts cannot connect over the encrypted tunnel. In this specific example, `zookeeper` needs to bind to 192.168.105.0 to accept the remote connection. I don't see _any_ local name set to this IP address.

    [root@zk0:~]# cat /etc/hosts
    127.0.0.1 localhost
    ::1 localhost

    10.233.4.2 zk0 zk0-unencrypted
    127.0.0.1 zk0-encrypted
    192.168.105.1 zk1 zk1-encrypted
    10.233.2.2 zk1 zk1-unencrypted
    192.168.105.2 zk2 zk2-encrypted
    10.233.3.2 zk2 zk2-unencrypted
    [root@zk0:~]# netstat -anp | grep 3888
    tcp        0      0 127.0.0.1:3888          0.0.0.0:*               LISTEN      700/java

Attempting to connect from `zk1` fails:

    [root@zk1:~]# nc -v zk0-encrypted 3888
    zk0-encrypted [192.168.105.0] 3888 (ciphire-serv): Connection refused

If we got `zookeeper` to bind to all interfaces, 0.0.0.0, connections would work, but this not what `zookeeper` is designed to do and is arguably bad practice.

PR #294, opened 2015-05-09
Set local ...-encrypted name to the tunnel IP #289

bruceadams:
This fixes the problem I describe in #289. I tested this change by deploying a Zookeeper cluster to nixos-containers and then to ec2.


rbvermaa:
Thanks!

PR #300, opened 2015-06-01
Fix container's deploy when used with keys.

jraygauthier:
Fundamentally there was this missing `self.set_common_state(defn)`
that wasn't called on `create` which is essential to retrieve
keys.

This should also fix an issue I had when `storeKeysOnMachine`
was not set explicitly which caused the keys not to be sent.

Attempted to fix the use of `stop`/`start` when there are keys.
There remain some underlying problem I couldn't understand yet.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -236,6 +236,10 @@ class MachineState(nixops.resources.ResourceState):
         else:
             return ["-p", str(self.ssh_port)]
 
+    def adapt_ssh_flag_for_use_with_scp(self, flags):
+      # For scp, the "-P" flag is used to specify the port. "-p"
+      # attempt to preserve file attributes.
+      return (map (lambda x: '-P' if x == '-p' else x, flags))
 
     def get_ssh_password(self):
         return None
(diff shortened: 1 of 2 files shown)


jraygauthier:
Can somebody tell me what's wrong with this patch? Does nobody use containers with secret
keys (e.g: pulling from a repository for example)? Anyway, if anybody uses secrets keys for
their amazon cloud or any other cloud, don't they need to test locally?

hrdinka:
I can confirm this bug. Here is a minimal test:

```
{
  test =
    { ... }:
    {
      deployment.targetEnv = "container";
      deployment.container.host = "localhost";
      deployment.keys.password.text = "foobar";
    }; 
}
```

Nixops does never send the keys and `/run/keys` stays empty. When `keys.target` gets started on the machine its activation will freeze forever cause it waits for `/run/keys/done` to be existent. I'll test this patch ASAP.

jraygauthier:
Re-applied and re-tested on top of latest.

ip1981:
Maybe related to https://github.com/NixOS/nixops/pull/650#issuecomment-299140107

domenkozar:
I've just merged https://github.com/NixOS/nixops/pull/804 that should fix this.

jraygauthier:
@domenkozar I just tested and confirm #804 fixes the problem.

jraygauthier:
Thanks for the merge!

PR #314, opened 2015-07-06
Do not ask if IP not in use

dansteeves68:
Do not ask if elastic IP is in use. I don't think I should have to use --confirm to get around this.


rbvermaa:
Thanks!

Issue #309, opened 2015-06-25
GCE needs to support privateIpv4


oconnorr:
In order for deploying a network of private servers, private IP addresses need to be supported.

I believe the IP address is available in node.private_ips[0] and so something similar to https://github.com/NixOS/nixops/blob/a932a9ecec15336a55f2aa2c2a20f08264adb837/nixops/backends/gce.py#L388 needs to be added for private IPs, plus all the support for a new value in the GCE state.

PR #322, opened 2015-07-27
Add support for networking.privateIPv4 for GCE machines.

oconnorr:
This addresses issue https://github.com/NixOS/nixops/issues/309.

I've tested this on one of my deployment.  After deployment of a GCE network, `nixops show-physical` now lists privateIPv4 for machines, while before this patch it did not.  Also I've successfully used networking.privateIPv4 in my GCE network configuration.

That said, I haven't really tested these corner cases of when the private IP address has unexpected changed.  Let me know if you want further testing, preferably with some hints on how to do further testing.


rbvermaa:
Merged in b79f208

rbvermaa:
Thanks!

PR #323, opened 2015-08-03
ec2: Assign public ip by default

offlinehacker:
How else is instance able to access internet, except maybe if you are running deticated NAT instance.

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -325,7 +325,7 @@ in
     };
 
     deployment.ec2.associatePublicIpAddress = mkOption {
-      default = false;
+      default = true;
       type = types.bool;
       description = ''
         If instance in a subnet/VPC, whether to associate a public


rbvermaa:
I am not sure if this is a good idea, what is the default behaviour in EC2 with regards to public IP assignment when starting instance in a VPC?

 Any other VPC users that have an option?

danbst:
You can specify in Amazon VPC dashboard default behavior for public IPs.

spinus:
I would vote -1 for public IP as default thing.

domenkozar:
Seems like this one is a no-no.

spinus:
@offlinehacker side comment, in AWS for example you have solution that you described, AWS has dedicated service to do NAT over private instances (or you can use VPN to your private network).

offlinehacker:
Tnx, I know this now, I didn't know a year ago.

PR #324, opened 2015-08-03
Add EC2 RDS resource to Nixops resources.

trentonstrong:
This is a base implementation for an RDS resource.  There are a few options that boto supports that could be added incrementally as needed (rds security groups, nix dependency checking, restoring from snapshots).  Perhaps should document that in the module docstring.

### DIFF
--- a/nix/ec2-rds-dbinstance.nix
+++ b/nix/ec2-rds-dbinstance.nix
@@ -1,0 +1,79 @@
+{ config, lib, uuid, name, ... }:
+
+with lib;
+
+{
+
+  options = {
+
+    id = mkOption {
+      type = types.str;
+      default = "nixops-${uuid}-${name}";
+      description = "Identifier for RDS database instance";
+    };
+
+    region = mkOption {
+      type = types.str;
+      description = "Amazon RDS region.";
+    };
+
+    accessKeyId = mkOption {
+      default = "";
+      type = types.str;
+      description = "The AWS Access Key ID.";
+    };
+
+    allocatedStorage = mkOption {
+      type = types.int;
+      description = "Allocated storage in GB";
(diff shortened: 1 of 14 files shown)


shlevy:
@rbvermaa any objections?

rbvermaa:
@shlevy No, just want to try it before merging and hadn't gotten around to it. Looking forward to the feature.

nix/ec2-rds-dbinstance.nix
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    port = mkOption {
+      type = types.int;
+      description = "Port for database instance connections.";
+    };
+
+    engine = mkOption {
+      type = types.str;
+      description = ''Database engine See <link
```
edolstra: Period missing here.

rbvermaa:
Thanks!

nixops/resources/ec2_rds_dbinstance.py
```diff
@@ -0,0 +1,291 @@
+# -*- coding: utf-8 -*-
+
+# Automatic provisioning of AWS RDS Database Instances.
+
+import boto.rds
+import nixops.resources
+import nixops.util
+import nixops.ec2_utils
+import time
+from uuid import uuid4
```
edolstra: Please load the `uuid` module on demand (like here: https://github.com/NixOS/nixops/blob/master/nixops/statefile.py#L153). Otherwise it adds about 50% to NixOps load time, because the uuid module does crazy stuff on load like invoking `gcc` and `objdump`.

trentonstrong:
Ah, did not know that!  Will submit a new PR with the changes you've requested @edolstra

Issue #312, opened 2015-06-29
Make GCE forwarding rules ephemeral IP addresses available for configuration


oconnorr:
There are two ways of making the IP address of a forwarding rule accessable, and probably both should be done.

(1) add an entry to the /etc/hosts files under the name of the forwarding rule
(2) make the publicIPv4 available as a configuration option like is done for machines.

PR #326, opened 2015-08-05
Add support for publicIPv4 for GCE resources.

oconnorr:
Configuration option publicIPv4 are added to both resources.gceForwardingRules and resources.gceStaticIPs.
These options are filled in by NixOps with the IP addresses assigned to these resources.

### DIFF
--- a/nix/gce-forwarding-rule.nix
+++ b/nix/gce-forwarding-rule.nix
@@ -31,6 +31,16 @@ with import ./lib.nix lib;
       '';
     };
 
+    publicIPv4 = mkOption {
+      default = null;
+      type = types.uniq (types.nullOr types.str);
+      description = ''
+        The assigned IP address of this forwarding rule.
+        This is set by NixOps to the ephemeral IP address of the resource if
+        ipAddress wasn't set, otherwise it should be the same as ipAddress.
+      '';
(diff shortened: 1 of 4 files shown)


oconnorr:
This addresses part (b) of issue https://github.com/NixOS/nixops/issues/312.

I've tested this on one of my deployment. After deployment of a GCE network, nixops show-physical now lists `publicIPv4` for forwarding rules and static ips, while before this patch it did not. Also I've successfully used forwarding rules's `publicIPv4` in my GCE network configuration.

An alternative to this implementation would be to have NixOps set the ipAddress configuration options of forwarding rules and static IPs instead of creating a new configuration option.  I rejected that approach because the `ipAddress` of a forwarding rule can be of type `string` or `gce-static-ip` and merging these two types would be complex.  In principle we could set the `ipAddress` option of static IPs because either it is the default value of null, in which case the physical IP address can override it, or it is a given value, in which case the physical IP address should match it, and they should be able to mergeEqual.  However, for consistency, I think it better to have a separate `publicIPv4` for static IPs as well.  It is probably also a better not to mix user specified options and NixOps specified options.

rbvermaa:
6bb8189

PR #335, opened 2015-09-08
Use the correct port when setting up an SSH tunnel

basvandijk:
This is needed for targets that use a non-standard targetPort.


rbvermaa:
Sorry for not looking at your PR earlier. I will try to act sooner next time.

basvandijk:
No problem. Thanks for the merge!

PR #336, opened 2015-09-14
Add support for non-machine IP resources in /etc/hosts

oconnorr:
There exist non-machine resouces, such as gce_forwarding_rule and gce_static_ip, that have IP addresses associated with them and are routable.
This change extends the generation of the hosts files to, by default, all resources that have a public_ipv4 attribute.

This addresses part (1) of issue #312 and would close the issue.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -242,9 +242,9 @@ class MachineState(nixops.resources.ResourceState):
     def private_ipv4(self):
         return None
 
-    def address_to(self, m):
-        """Return the IP address to be used to access machone "m" from this machine."""
-        ip = m.public_ipv4
+    def address_to(self, r):
+        """Return the IP address to be used to access resource "r" from this machine."""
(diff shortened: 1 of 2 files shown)


rbvermaa:
Sorry for the slow merge, I will try to merge sooner next time.

nixops/backends/__init__.py
```diff
@@ -242,9 +242,9 @@ def public_ipv4(self):
     def private_ipv4(self):
         return None
 
-    def address_to(self, m):
-        """Return the IP address to be used to access machone "m" from this machine."""
-        ip = m.public_ipv4
+    def address_to(self, r):
+        """Return the IP address to be used to access resource "r" from this machine."""
+        ip = getattr(r, 'public_ipv4', None)
```
edolstra: Why not move the public_ipv4 property to the resource class? Calling `getattr` from the outside seems ugly.

rbvermaa: Moved the property in 98b4edd

PR #355, opened 2015-11-07
Use IPv6 capable socket to check if SSH is available

fpletz:
Use an AF_INET6 instead of an AF_INET socket to probe the SSH TCP port. This also works for IPv4.

Previously, nixops would wait forever for SSH to become available on an IPv6-only machine.

### DIFF
--- a/nixops/util.py
+++ b/nixops/util.py
@@ -149,7 +149,7 @@ def make_non_blocking(fd):
 
 
 def ping_tcp_port(ip, port, timeout=1, ensure_timeout=False):
-    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
     s.settimeout(timeout)
     s.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack('ii', 1, 0))
     try:


edolstra:
This does not work for me. And unless Python does some magic mapping AF_INET6 to AF_INET on systems without IPv6 connectivity, I wouldn't expect it to work...

fpletz:
Hmm, you're right. It also works for me now. Not sure what was broken in my setup.

wizeman:
I have run into this issue as well, which happens to me when the target only has an IPv6 address (I have filed #601).

I am very far from being an IPv6 expert, but my understanding is that this patch wouldn't work only on systems which have IPv6 completely disabled in the kernel (which should be almost none these days?). If IPv6 is enabled, but the system doesn't have IPv6 addresses, then I think it should work (at least in theory).

However, it may be necessary to specify an IPv4 address using IPv6 notation, i.e., `::ffff:192.168.x.x`, which could be done automatically by `nixops`.

wizeman:
Alternatively, maybe `socket.create_connection()` could be used instead of `socket.connect()`.

According to https://docs.python.org/2/library/socket.html#socket.create_connection:

> ```
> socket.create_connection(address[, timeout[, source_address]])
> ```
> Connect to a TCP service listening on the Internet address (a 2-tuple (host, port)), and return the socket object. This is a higher-level function than socket.connect(): if host is a non-numeric hostname, it will try to resolve it for both AF_INET and AF_INET6, and then try to connect to all possible addresses in turn until a connection succeeds. **This makes it easy to write clients that are compatible to both IPv4 and IPv6**.

edolstra:
Yes, that sounds like the way to go.

wizeman:
Actually, #486 seems like a better, more generic fix.

As it stands it doesn't work yet, but I've added a comment indicating some additional fixes which are likely to be necessary.

PR #359, opened 2015-11-17
Change sbin/mount.vboxsf to bin/mount.vboxsf

ninegua:
The virtualbox package no longer has mount.vboxsf under sbin.


domenkozar:
cc @rbvermaa

PR #364, opened 2015-11-23
backends/none: don't use _ssh_private_key if its corresponding public key hasn't been deployed yet

### LINKED ISSUE #361: Can't deploy none backend due to public key not being present
@aszlig I think that because of your commit https://github.com/NixOS/nixops/commit/79857da39416627aaff214f5bfaca3e8a2867461 I'm no longer able to deploy to a bunch of existing `none` machines that I was previously able to deploy to.

The problem is that the `create` method of the `none` backend is [setting](https://github.com/NixOS/nixops/blob/6b1ff732e25aebd286ec19525785fa510c2b3607/nixops/backends/none.py#L55) a SSH keypair if the `self.vm_id` attribute is not set. Since my machines haven't been deployed yet using a nixops with your patch they don't have the `vm_id` attribute. After setting the keypair the `get_ssh_flags` method will add the `"-i", self.get_ssh_private_key_file()` flag to the SSH options. This prevents nixops from authenticating with the machine since the public key is not yet present on the machine.

If I remove the  `"-i", self.get_ssh_private_key_file()` flag from the  `get_ssh_flags` method I am able to deploy. However, we need a backward compatible way of using the latest nixops with machines deployed with an older version of nixops.

basvandijk:
@aszlig I think this should fix #361. I haven't tested it yet but will do that tomorrow.

The boolean attribute `none.sshPublicKeyDeployed` is added to the
NoneState and defaults to False. It indicates whether the SSH public key
has been deployed to the target. If it's False the `get_ssh_flags()`
function will not set the corresponding private key as the identity.

`none.sshPublicKeyDeployed` will be set to True after activating the
configuration. From then on `get_ssh_flags()` will set the identity to the
corresponding private key allowing nixops to authenticate itself using
the keypair stored in its DB.

### DIFF
--- a/nixops/backends/none.py
+++ b/nixops/backends/none.py
@@ -29,6 +29,7 @@ class NoneState(MachineState):
     target_host = nixops.util.attr_property("targetHost", None)
     _ssh_private_key = attr_property("none.sshPrivateKey", None)
     _ssh_public_key = attr_property("none.sshPublicKey", None)
+    _ssh_public_key_deployed = attr_property("none.sshPublicKeyDeployed", False, type = bool)
 
     def __init__(self, depl, name, id):
         MachineState.__init__(self, depl, name, id)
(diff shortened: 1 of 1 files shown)


basvandijk:
I've now tested this and it works perfectly.

domenkozar:
it's very unfortunate we track this state. I don't really understand why it's needed, what does `-i` prevent really?

basvandijk:
@domenkozar if `-i` is set and the corresponding public key hasn't been deployed yet SSH won't be able to sign in.

PR #369, opened 2015-12-05
Fix #362: chown the key using the uid and gid

### LINKED ISSUE #362: chmod of keys can occur before group is created.
Occasionally when provisioning a new node I get the following error:

```
s10> uploading key ‘ssl_www.kevincox.ca.crt’...
s10> chown: invalid group: ‘root:ssl_www.kevincox.ca’
```

This can be worked around by:
1. Commenting out the key.
2. Deploying (creating the group).
3. Uncommenting the key.
4. Deploying.

basvandijk:
Previously the owner of the uploaded key would be set to the specified user name and group name. However, when the configuration which defines this user and group is not yet activated the chown operation will fail because the user or group might not be present.

What we now do instead is set the owner of the key using the UID and GID computed from the configuration.

This does enable a small attack window after the chown UID:GID and before the activation of the new configuration which adds the new user and group. If there's an existing user or group with the specified UID or GID respectively he might be able to read the key during that short window.

I also realise this doesn't yet work when `users.mutableUsers` is set the `true`. Ideally we would fall back to changing the ownership based on the user and group name when `users.mutableUsers` is set to`true`.

### DIFF
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -32,6 +32,22 @@ let
       '';
     };
 
+    options.uid = mkOption {
+      internal = true;
+      type = types.int;
+      description = ''
+        Internal read-only option which should be the UID of the user.
+      '';
+    };
+
+    options.gid = mkOption {
+      internal = true;
+      type = types.int;
+      description = ''
+        Internal read-only option which should be the GID of the group.
+      '';
+    };
+
     options.permissions = mkOption {
       default = "0600";
       example = "0640";
@@ -50,6 +66,11 @@ let
             + " \"<value>\"` instead of `deployment.keys.${key} ="
             + " \"<value>\"`.";
   in if isString val then builtins.trace warning { text = val; } else val;
+
+  setIds = keyOptions: keyOptions // {
+             inherit (config.users.users.${keyOptions.user}) uid;
+             inherit (config.users.groups.${keyOptions.group}) gid;
+           };
 
   keyType = mkOptionType {
     name = "string or key options";
@@ -88,8 +109,7 @@ in
       default = {};
       example = { password.text = "foobar"; };
       type = types.attrsOf keyType;
-      apply = mapAttrs convertOldKeyType;
-
+      apply = mapAttrs (key: val: setIds (convertOldKeyType key val));
       description = ''
         The set of keys to be deployed to the machine.  Each attribute
         maps a key name to a file that can be accessed as
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -26,6 +26,10 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
                 elem = x.find("attrs/attr[@name='{0}']/string".format(key))
                 if elem is not None:
                     opts[key] = elem.get("value")
+            for key in ('uid', 'gid'):
+                elem = x.find("attrs/attr[@name='{0}']/int".format(key))
+                if elem is not None:
+                    opts[key] = elem.get("value")
             return opts
 
         self.keys = {k.get("name"): _extract_key_options(k) for k in
@@ -209,7 +213,7 @@ class MachineState(nixops.resources.ResourceState):
             chmod = "chmod '{0}' " + outfile_esc
             chown = "chown '{0}:{1}' " + outfile_esc
             self.run_command(' && '.join([
-                chown.format(opts['user'], opts['group']),
+                chown.format(opts['uid'], opts['gid']),
                 chmod.format(opts['permissions'])
             ]))
             os.remove(tmp)


kevincox:
Running this version I get the following error:

``
s1......> copying closure...
s6......> copying closure...
s6......> uploading key ‘ssl_api-o.kevincox.ca.key’...
s6......> error: Traceback (most recent call last):
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 665, in worker
    m.send_keys()
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 208, in send_keys
    self.upload_file(tmp, outfile)
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 347, in upload_file
    cmdline = ["scp"] + self.get_ssh_flags(True) + master.opts
TypeError: get_ssh_flags() takes exactly 1 argument (2 given)

s1......> uploading key ‘ssl_www.kevincox.ca.key’...
s1......> error: Traceback (most recent call last):
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 665, in worker
    m.send_keys()
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 208, in send_keys
    self.upload_file(tmp, outfile)
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 347, in upload_file
    cmdline = ["scp"] + self.get_ssh_flags(True) + master.opts
TypeError: get_ssh_flags() takes exactly 1 argument (2 given)

  File "/home/kevincox/p/nixops/scripts/nixops", line 921, in <module>
    args.op()
  File "/home/kevincox/p/nixops/scripts/nixops", line 372, in op_deploy
    repair=args.repair)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 941, in deploy
    self._deploy(**kwargs)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 926, in _deploy
    sync=sync, always_activate=always_activate)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 709, in activate_configs
    .format(len(failed), len(res), ", ".join(["‘{0}’".format(x) for x in failed])))
Exception: activation of 2 of 2 machines failed (namely on ‘s6’, ‘s1’)
```

**Comment by basvandijk** [comment]
This doesn't have anything to do with my pull request but you did found a bug. The problem is that the `get_ssh_flags` function in the [none](https://github.com/NixOS/nixops/blob/98bbc07670e26a9d03081af10710aeeff5c9ba58/nixops/backends/none.py#L69) and [container](https://github.com/NixOS/nixops/blob/ce9c8d3670de26175286741ef95b7f9736269adc/nixops/backends/container.py#L60) backends don't specify the required `scp` keyword option as is defined in their [super class](https://github.com/NixOS/nixops/blob/98b4edd519355c5ac2ed7d8ada0fb0c61a9509ca/nixops/backends/__init__.py#L224). Care to submit a pull request with a fix?

**Comment by kevincox** [comment]
I'll try to find time to get an environment set up.

**Comment by aszlig** [comment]
@kevincox: Your issue should be fixed by 880fc520838d0790a8145d961f9f8b19d3f43cbc.

@basvandijk: The race condition is a bit ugly, so I'd rather flip order of activation/sendkeys instead. Of course, services not having `after = [ "keys.target" ]` will also run into race conditions, so a better idea probably would be to use (bind) mount units directly on each key (or maybe on `/run/keys` in the short run), so services depending on a particular key will block until it is available. That at least would prevent races and the need to explicitly depend on `keys.target`.

**Comment by kevincox** [comment]
Thanks @aszlig however I had to extend your patch a bit. Please see #371.

After cherry-picking your commit on top I get the following error which sounds more related an I suspect is because I'm not explicit assigning GIDs.

```
s6......> uploading key ‘ssl_api-o.kevincox.ca.key’...
s6......> error: Traceback (most recent call last):
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 665, in worker
    m.send_keys()
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 216, in send_keys
    chown.format(opts['uid'], opts['gid']),
KeyError: 'gid'

s1......> uploading key ‘ssl_www.kevincox.ca.key’...
s1......> error: Traceback (most recent call last):
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 665, in worker
    m.send_keys()
  File "/home/kevincox/p/nixops/nixops/backends/__init__.py", line 216, in send_keys
    chown.format(opts['uid'], opts['gid']),
KeyError: 'gid'

  File "/home/kevincox/p/nixops/scripts/nixops", line 921, in <module>
    args.op()
  File "/home/kevincox/p/nixops/scripts/nixops", line 372, in op_deploy
    repair=args.repair)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 941, in deploy
    self._deploy(**kwargs)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 926, in _deploy
    sync=sync, always_activate=always_activate)
  File "/home/kevincox/p/nixops/nixops/deployment.py", line 709, in activate_configs
    .format(len(failed), len(res), ", ".join(["‘{0}’".format(x) for x in failed])))
Exception: activation of 2 of 2 machines failed (namely on ‘s6’, ‘s1’)
```

kevincox:
Also switching the order does sound more correct and robust to me as well. I think it should be safe as I see no way that a user or group could have a dependency on a key.

aszlig:
@kevincox: A user/group shouldn't have a dependency on a key but a service could have.

kevincox:
On Dec 7, 2015 23:58, "aszlig" <notifications@github.com> wrote:
>
> @kevincox: A user/group shouldn't have a dependency on a key but a
service could have.
>

I really don't have any understanding of the internal architecture but
couldn't it run between the two? Or are services and groups managed
"atomically" by nix

basvandijk:
> a better idea probably would be to use (bind) mount units directly on each key

@aszlig that's indeed a way better idea than my uid:gid trick.

I'll close this PR and add your idea to #362.

nh2:
Ah @basvandijk @aszlig and @kevincox, looks like we're working on the same thing at the same time again :)

Probably relevant to you: #803

nh2:
> working on the same thing at the same time again

I just realised that it's exactly the same day, but 2 years later :)

PR #387, opened 2016-01-24
Add disks option to libvirt backend. Similar to VirtualBox option.

coreyoconnor:
This uses nix/generate-ext4-image to generate an appropriately sized ext4 image.
Unless a baseImage path is provided as part of the network.
This is performed during create by the libvirtd backend python code.
libvirtd-image.nix was renamed libvirtd-root-image to clarify the image build
performed is only for the root filesystem with the base nix install.


domenkozar:
In general I  think it's a good step forward, but the code is too fragmented. Both of the FS generation images should use `<nixpkgs/nixos/lib/make-disk-image.nix>`

coreyoconnor:
The root file system definitely should use make-disk-image. Any reason the original disk image code did not? I don't see anything obvious.

The alternate disks should be images without content. The make-disk-image function installs a nixos configuration onto the generated disk image. Is that appropriate in this case?

I'll change the root filesystem disk image generation. Let me know about the alternate disks. Any other aspect need revision?

coreyoconnor:
clearing old PRs. I'm no longer developing on nixops.

PR #393, opened 2016-02-12
Add m4 & c4 instance type physical properties

mbbx6spp:
The problem this attempts to solve is to get the default EBS optimized EC2
attribute set appropriately for m4 instances. Currently despite m4 instance
types supposedly being EBS-optimized by default, without these entries nixops
will create them unoptimized.


rbvermaa:
Thanks!

PR #400, opened 2016-02-25
Allow new keys & users/groups to be created in the same deploy

ryanartecona:
### Delay chown of keys until user/group both exist

Instead of chowning keys to their user/group every time they are sent,
only attempt the chown during send-keys if the user and group both
exist, and again do a chown during activation after the users and groups
have been created.

One result is that if a key and its user and/or group are to be created
in the same `nixops deploy`, the key will first be uploaded and owned
by root:root, then chmod'd, then late in activation the key will be
chowned to the newly created user/group. This includes a node's first
deploy, when it has neither keys nor users/groups.

Another result is that between send-keys and the next deploy (often,
but not necessarily, in the same `nixops deploy`), a key may
have its permissions set as configured, but _not_ be owned by the
configured user/group (instead root:root), which is presumed safe 
(**is it really though?**).
### Add service per key to track key dependencies

If a user service wants to block until `deployment.keys.mysecret` exists
to start up, it can now do so by adding `"mysecret-key.service"` to
its `requires` and `after` lists.

I tried doing this first via a systemd mount unit and a systemd path unit.
I couldn't get the mount to work (I have little experience configuring
filesystems). The path unit only supports unit _activation_, and only for
a single unit, and besides that it's extra fiddly to configure. 

For the per-key services, the `preStart` script will block until the key exists, 
and then the `script` will block until the key no longer exists. Permissions
aren't checked, only existence, as permissions are assumed to be correctly
set during activation.

### DIFF
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -1562,6 +1562,88 @@ deploy the new configuration.
 
 </section>
 
+<section>
+  <title>Managing keys</title>
+
+  <para>
+    Files in <filename>/nix/store/</filename> are readable by every
+    user on that host, so storing secret keys embedded in nix derivations
+    is insecure. To address this, nixops provides the configuration
+    option <varname>deployment.keys</varname>, which nixops manages
+    separately from the main configuration derivation for each machine.
+  </para>
+
+  <para>
+    Add a key to a machine like so.
+
+    <screen>
+{
+  machine =
+    { config, pkgs, ... }:
+    {
+      deployment.keys.my-secret.text = "shhh this is a secret";
+      deployment.keys.my-secret.user = "myuser";
+      deployment.keys.my-secret.group = "wheel";
+      deployment.keys.my-secret.permissions = "0640";
+    };
+}
+    </screen>
+
+    This will create a file <filename>/run/keys/my-secret</filename>
+    with the specified contents, ownership, and permissions.
+  </para>
+
+  <para>
+    Among the key options, only <varname>text</varname> is required. The
+    <varname>user</varname> and <varname>group</varname> options both default
+    to <literal>"root"</literal>, and <varname>permissions</varname> defaults
+    to <literal>"0600"</literal>.
+  </para>
+
+  <para>
+    Keys from <varname>deployment.keys</varname> are stored under <filename>/run/</filename>
+    on a temporary filesystem and will not persist across a reboot.
+    To send a rebooted machine its keys, use <command>nixops send-keys</command>. Note that all
+    <command>nixops</command> commands implicitly upload keys when appropriate,
+    so manually sending keys should only be necessary after an unattended reboot.
+  </para>
+
+  <para>
+    If you have a custom service that depends on a key from <varname>deployment.keys</varname>,
+    you can opt to let systemd track that dependency. Each key gets a corresponding
+    systemd service <literal>"${keyname}-key.service"</literal> which is active
+    while the key is present, and otherwise inactive when the key is absent. See
+    <xref linkend="key-dependency.nix" /> for how to set this up.
+
+    <example xml:id="key-dependency.nix">
+      <title><filename>key-dependency.nix</filename>: track key dependence with systemd</title>
+      <programlisting>
+{
+  machine =
+    { config, pkgs, ... }:
+    {
+      deployment.keys.my-secret.text = "shhh this is a secret";
+
+      systemd.services.my-service = {
+        after = [ "my-secret-key.service" ];
+        wants = [ "my-secret-key.service" ];
+        script = ''
+          export MY_SECRET=$(cat /run/keys/my-secret)
+          run-my-program
+        '';
+      };
+    };
+}
+      </programlisting>
+    </example>
+
+    These dependencies will ensure that the service is only started when the keys it
+    requires are present. For example, after a reboot, the services will be delayed
+    until the keys are available, and <command>systemctl status</command> and friends
+    will lead you to the cause.
+  </para>
+</section>
+
 <!--
 
 <para>EC2 logical.nix</para>
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -91,15 +91,20 @@ in
       apply = mapAttrs convertOldKeyType;
 
       description = ''
-        The set of keys to be deployed to the machine.  Each attribute
+        <para>The set of keys to be deployed to the machine.  Each attribute
         maps a key name to a file that can be accessed as
         <filename>/run/keys/<replaceable>name</replaceable></filename>.
         Thus, <literal>{ password.text = "foobar"; }</literal> causes a
         file <filename>/run/keys/password</filename> to be created
         with contents <literal>foobar</literal>.  The directory
         <filename>/run/keys</filename> is only accessible to root and
-        the <literal>keys</literal> group.  So keep in mind to add any
-        users that need to have access to a particular key to this group.
+        the <literal>keys</literal> group, so keep in mind to add any
+        users that need to have access to a particular key to this group.</para>
+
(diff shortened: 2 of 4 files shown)


basvandijk:
+1 I like it.

nix/keys.nix
```diff
@@ -136,23 +136,64 @@ in
...
-        before = [ "keys.target" ];
-        unitConfig.DefaultDependencies = false; # needed to prevent a cycle
-        serviceConfig.Type = "oneshot";
-        serviceConfig.RemainAfterExit = true;
-        script =
+        ${concatStringsSep "\n" (flip mapAttrsToList config.deployment.keys (name: value:
+          # Make sure each key has correct ownership, since the configured owning
+          # user or group may not have existed when first uploaded.
+          ''
+            [[ -f "/run/keys/${name}" ]] && chown '${value.user}:${value.group}' "/run/keys/${name}"
```
ip1981: what if the file does not exist?
I'd use single quotes.

ip1981: Maybe keep only 
```
chown '${value.user}:${value.group}' '/run/keys/${name}'
```
to fail noisily

kevincox:
I'll test this out today.

kevincox:
This is working perfectly for me! Thanks so much, this will save me a lot of manual fiddling in the near future.

ip1981:
Using numbers instead of names works right now :-)

kevincox:
I know, but I generate my groups programmatically and assigning unique numbers would be a non-trivial task. So I'm putting up with it for now. I like not having to think about group numbers if I can avoid it.

ip1981:
Ah, yeah. I did assigning unique IDs trivial: https://marc.info/?l=nix-dev&m=145046990923112&w=2

kevincox:
On 08/06/16 10:44, Igor Pashev wrote:
> Ah, yeah. I did assigning unique IDs trivial:
> https://marc.info/?l=nix-dev&m=145046990923112&w=2

Clever. If this doesn't ship soon I might try that out.

nix/keys.nix
ryanartecona: Perhaps. I was unsure whether to fail noisily, or to treat this as best-effort chowning of keys that both (1) we know about at this point in activation and (2) exist, having been placed here for us before activation even started.

I opted for the latter, so that the absence of one key doesn't screw up ownership of other keys or fail an activation. Also, if a service depends on a specific key, they can now do so via a systemd service dependency, since each key gets a service which is running as long as the key exists (added below).

ryanartecona:
Any big reason why this can't be merged? Is it just a lack of testing?

kevincox:
Yeah, I was going to ping this today as well because I had to provision a
new machine.

ryanartecona:
Just rebased onto current master.

domenkozar:
@ryanartecona two things I think would make it easier to merge this:

- document how keys are now used in deployments to nixops manual
- provide a [simple example](sscce.org) how to test this PR with nixops

I know this is effort, but I've seen such PRs rot exactly for these reasons, reviewers don't have much time if they're not particularly interested into the feature/bug.

kevincox:
@ryanartecona Do you have any plans to add the documentation. I think only a couple of lines would be required.

@domenkozar For a simple example see the following. Either of the clauses will fail (when deploying to a machine where that user/group did not previously exist) where as this patch should make the deploy successful.

```nix
{
	users.users.a-user = {};
	deployment.keys.example = {
		user = "a-user";
		text = "blah";
	};
	
	# or
	
	users.groups.a-group = {};
	deployment.keys.example = {
		group = "a-group";
		text = "group blah";
	};
}
```

ryanartecona:
Sorry for the radio silence.

I just rebased this branch onto current master and added commit aeb5b1e which documents `deployment.keys`,  and `nixops send-keys`. I would definitely like a second pair of eyes on the documentation. It seems to build and render as intended on my computer, but I have never worked with docbook before. 

Tangentially, I noticed that the "Command Reference" section of the manual is fairly thorough, but I had never realized it because none of its subsections are in the table of contents. I made a feeble effort to get those `<refsection>`s into the table of contents, but my docbook-fu isn't nearly strong enough.

Anything else I should provide so this can get merged?

doc/manual/overview.xml
```diff
@@ -1562,6 +1562,94 @@ deploy the new configuration.
...
+
+  <note>
+    <para>
+      Keys in <literal>/run/keys/</literal> are not stored on disk by default.
+    </para>
+  </note>
+
+  <para>
+    Since the <filename>/run</filename> directory is mounted as a tmpfs filesystem,
+    keys stored there from <varname>deployment.keys</varname> are stored in RAM
```
kevincox: I prefer:

> Keys from <varname>deployment.keys</varname> are stored user <filename>/run/</filename> on a temporary filesystem and will not be persisted across a reboot. To send a rebooted machine it's keys, use <command>nixops send-keys</command>. Note that all  <command>nixops</command> commands implicitly upload keys when appropriate,  so manually sending keys should only be necessary after an unattended reboot.

It is a slightly rearranged version that
- Removes a couple implementation details that I didn't think were relevant here.
- Made `nixops send-keys` sentence a little more clear and consistent with terminology.

What do you think?

doc/manual/overview.xml
```diff
@@ -1562,6 +1562,94 @@ deploy the new configuration.
...
+  <para>
+    Among the key options, only <varname>text</varname> is required. The
+    <varname>user</varname> and <varname>group</varname> options both default
+    to <literal>"root"</literal>, and <varname>permissions</varname> defaults
+    to <literal>"0600"</literal>.
+  </para>
+
+  <note>
+    <para>
+      Keys in <literal>/run/keys/</literal> are not stored on disk by default.
```
kevincox: I think this paragraph is unnecessary because the following one covers this.

doc/manual/overview.xml
```diff
@@ -1562,6 +1562,94 @@ deploy the new configuration.
...
+        '';
+      };
+    };
+}
+      </programlisting>
+    </example>
+
+    A service with a key dependency set up in this way will be prevented by systemd
+    from being started after an unattended reboot, and vanilla <command>systemctl status</command>
+    and friends will lead you to the cause.
```
kevincox: How about:

> These decencies will ensure that the service is started when the keys it requires are present. For example, after a reboot the services will be delayed until the keys are available and <command>systemctl status</command>  and friends will lead you to the cause.

- avoids the ambiguous phrase "...a key dependency...
- Changes "prevents starting" to "starts when the keys are present" (sounds more positive I think)

doc/manual/overview.xml
ryanartecona: I was on the fence here myself. I wasn't sure if this was important enough to give a little highlighted callout, or just to trust that whoever it concerns will read through this whole new keys section. Thinking more now, the following paragraph should be sufficient. Taking this out.

ryanartecona:
@kevincox Thanks for the thoughtful docs review! I've incorporated your suggestions.

kevincox:
Thanks, LGTM.

@domenkozar What do you think?

spinus:
+1 for the work, cheers.

Would be nice to add a test for it as well.

domenkozar:
Sorry it took so long, but I'll try to review this shortly. Who's using this feature?

domenkozar:
This also fixed https://github.com/NixOS/nixops/issues/461

kevincox:
I'm using this feature. This fixes problems whenever someone is using a key assigned to a named group. When nixops is run, and the group does not yet exist you will get an error uploading keys because it tries to chown it to the group (which does not yet exist).

spinus:
I'm not using yet but I would like to :-) for the same reasons Kevin mentioned.

PR #401, opened 2016-02-27
Create /etc/hosts with none backend.

### LINKED ISSUE #373: IPs of machines with the backend none aren't tracked
They're not listed in /etc/hosts, and their IPs are listed as blank in "nixops info".

kevincox:
I'm not sure whether the ip to use should default to `deployHost` or not but this seems safer plus `deployHost` might be a hostname. This uses the value of `networking.publicIPv4`.

Fixes #373


domenkozar:
Does the `none` backend populate networking.publicIPv4 in some way?

kevincox:
Not by default. You have to set it yourself at this point.

domenkozar:
@rbvermaa

domenkozar:
@rbvermaa @edolstra  any reason not to merge this?

domenkozar:
I see no reason why to not track this state, merging.

Issue #260, opened 2015-01-27
Support building fully remotely


wmertens:
If you're running nixops on your OS X laptop and deploy, it will first download a bunch of things to your laptop and fail at some point because you're on Darwin. Next, you log in on the deployed remote host and add your own ssh key so that nix-build can use the host. Then, you deploy again and it starts uploading what it just downloaded. This is not a nice user story.

It would be great if there was support for a buildhost, which runs the deployment build and then pushes to the other nodes in the network. One or more of the machines in the network are marked as workers and one is marked as nixops master. If the nixops master is defined it will perform all downloads and coordinate builds, and all other machines are populated from it.

Something like that? Or at the very least documenting a workflow like "deploy a VM from this image, do this to make it into a regular nixos machine with nixops and copy your network definitions on there" would be nice.

copumpkin:
:+1: ran into the exact same issue

domenkozar:
I have a patch somewhere...

wmertens:
@domenkozar about that patch...  send it to me and I'll clean it up? Or is it ready to go?

edolstra:
In theory this is already supposed to work. From the code:
```
        # If we're not running on Linux, then perform the build on the
        # target machines.  FIXME: Also enable this if we're on 32-bit
        # and want to deploy to 64-bit.
        if platform.system() != 'Linux' and os.environ.get('NIX_REMOTE') != 'daemon':
            if os.environ.get('NIX_REMOTE_SYSTEMS') == None:
                remote_machines = []
                for m in sorted(selected, key=lambda m: m.index):
                    key_file = m.get_ssh_private_key_file()
                    if not key_file: raise Exception("do not know private SSH key for machine ‘{0}’".format(m.name))
                    # FIXME: Figure out the correct machine type of ‘m’ (it might not be x86_64-linux).
                    remote_machines.append("root@{0} {1} {2} 2 1\n".format(m.get_ssh_name(), 'i686-linux,x86_64-linux', key_file))
                    # Use only a single machine for now (issue #103).
                    break
                remote_machines_file = "{0}/nix.machines".format(self.tempdir)
                with open(remote_machines_file, "w") as f:
                    f.write("".join(remote_machines))
                os.environ['NIX_REMOTE_SYSTEMS'] = remote_machines_file
            else:
                self.logger.log("using predefined remote systems file: {0}".format(os.environ['NIX_REMOTE_SYSTEMS']))

[...]

PR #412, opened 2016-03-12
Create containers remotely when currentSystem != Linux

### LINKED ISSUE #398: Failure deploying containers from OS X
For example, with 

nixops create -d vbox-contain nixops/examples/trivial-container-in-vbox.nix
nixops deploy -d vbox-contain
# ..Build vbox machine..
machine> building initial configuration...
error: assertion failed at /nix/store/4han2qc1f181nw7f1bvf1an721xdb796-nixpkgs-16.03pre75806.77f8f35/nixpkgs/pkgs/os-specific/linux/kernel/generic.nix:31:1
error: Command '['nix-build', '<nixpkgs/nixos>', '-A', 'system', '-I', 'nixos-config=/var/folders/lv/g1444dx54dz8q2tc2lk3vy8h0000gn/T/nixops-tmpgddZKk/machine-initial.nix', '-I', 'nixops=/nix/store/fiapphz7gkhiirm01f2bl8nzn6rd9350-nixops-1.3.1/share/nix/nixops']' returned non-zero exit status 1

I see that at an [backends/containers.py](https://github.com/NixOS/nixops/blob/e5fb3c50c8a259204b675337791edd67be32d9cb/nixops/backends/container.py#L140) an initial container config is built against builtins.currentSystem. I assume that's the cause of the error, or am I missing a trick somewhere? 

I've created a fork and branch that could avoid this but I'd like to check I'm not doing something dumb before I submit a PR. 

Thanks!

brocking:
Resolves #398

### DIFF
--- a/nix/container.nix
+++ b/nix/container.nix
@@ -33,6 +33,7 @@ in
 
     boot.isContainer = true;
 
+    nixpkgs.system = mkIf (! any (platform: platform == builtins.currentSystem) platforms.linux) (mkOverride 999 "x86_64-linux");
   };
 
 }
--- a/nixops/backends/container.py
+++ b/nixops/backends/container.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from nixops.backends import MachineDefinition, MachineState
+from nixops.nix_expr import py2nix
 import nixops.util
 import nixops.ssh_util
 import subprocess
@@ -126,10 +127,14 @@ class ContainerState(MachineState):
         if self.vm_id is None:
             self.log("building initial configuration...")
 
+            eval_args = self.depl._eval_args(self.depl.nix_exprs)
+            eval_args['checkConfigurationOptions'] = False
             expr = " ".join([
                 '{ imports = [ <nixops/container-base.nix> ];',
                 '  boot.isContainer = true;',
                 '  networking.hostName = "{0}";'.format(self.name),
+                '  nixpkgs.system = let final = import <nixops/eval-machine-info.nix> {0};'.format(py2nix(eval_args, inline=True)),
+                '  in final.resources.machines.{0}.nixpkgs.system;'.format(self.name),
                 '  users.extraUsers.root.openssh.authorizedKeys.keys = [ "{0}" ];'.format(self.client_public_key),
                 '}'])
 
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -250,16 +250,26 @@ class Deployment(object):
         flags.extend(["-I", "nixops=" + self.expr_path])
         return flags
 
+    def _eval_args(self, exprs):
+        args = {key: RawValue(val) for key, val in self.args.iteritems()}
+        exprs_ = [RawValue(x) if x[0] == '<' else x for x in exprs]
+        eval_args = {
+            'networkExprs' : exprs_,
+            'args' : args,
+            'uuid' : self.uuid,
+            'deploymentName' : self.name if self.name else ""
+        }
+        return eval_args
+
 
     def _eval_flags(self, exprs):
         flags = self._nix_path_flags()
-        args = {key: RawValue(val) for key, val in self.args.iteritems()}
-        exprs_ = [RawValue(x) if x[0] == '<' else x for x in exprs]
+        args = self._eval_args(exprs)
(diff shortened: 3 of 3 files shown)


gilligan:
I just ran into the same problem - @domenkozar could this be merged ?

domenkozar:
Someone has to review the code and test it. I currently don't have time for it, sorry :/

3noch:
Any progress on this?

johbo:
Just came across this issue. Just started to try this PR on my OSX box, so far it's working fine.

yurrriq:
This didn't work for me when trying to deploy Digital Ocean or VirtualBox.

johbo:
@yurrriq mind to share a few more details about your setup? 

On my end I think that things just worked because I did already have remote building set up, so the code which tries to auto-configure the remote machines was not hit in my case.

yurrriq:
I've got remote building set up too via [nix-docker](https://github.com/lnl7/nix-docker) and I use [nix-darwin](https://github.com/lnl7/nix-darwin) to manage [my config](https://github.com/yurrriq/dotfiles/blob/374ec28f72bf5e6368afc4ddbd9d7932f435c869/darwin-configuration.nix#L270-L281).

nick-b-h:
@yurrriq If the NIX_REMOTE_SYSTEMS variable is already set nixops won't amend it with the machines that it knows about. [1](https://github.com/NixOS/nixops/blob/21dc59f2294494024fd472d552033e372634ab55/nixops/deployment.py#L612) You will need to yourself add the details of the machines you want used for remote building to the file which that variable points to. 

On a more general note, I think the discussion in #260 has better solutions than what I attempted to resolve in this PR, because it relates to a slightly more general issue around nixops remote deployment. But when I opened this PR I didn't know about any of that discussion, and since it didn't attract much attention I thought I may as well leave it open. So if #260 is resolved this could be closed, but until then it may be helpful for it stay open for others to patch onto their nixops builds.

betaboon:
I just stumbled upon this problem when deploying containers into a virtualbox from macos.

I just tested this PR:
- With a already working remote-build setup (configured using the nix-daemon plist) this solves the problem of not being able to deploy containers into a virtualbox from macos.
- With the remote-build setup disabled (so no remote-build settings in nix-daemon plist) this still fails.


As #260 hasn't progressed in almost a year, maybe we can progress on this PR?

johbo:
Maybe we can split out a minimal change, in my case the change to `container.nix` does seem to do the trick already.

johbo:
The changes in https://github.com/johbo/nixops/commit/64d1636344f352b32a98597711c16776cd4d941c did work fairly well in my case from darwin.

This would allow us to merge the good parts and leave the question regarding the moved chunk of python code open for the future. On my end I am not even so sure if we really should aim to automagically remote-build on the target machine. My bet is that we can live with a requirement that people on non-linux will have to set up remote building for nix in order to be able to use nixops on top of it.

As far as I understand the change, it does the following two bits:

1. Apply a default for `nixpkgs.system` also for the container backend.

2. In `container.py` there is a special twist: In `create` we first create a blank container and this one also needs to inherit the `nixpkgs.system` value. The whole magic around `eval_args` seems to address this problem.

Hope that's an alternative which makes it easier to merge in the needed parts. :)

dhess:
Unless I've missed something, it appears that @johbo's nice patch (https://github.com/johbo/nixops/commit/64d1636344f352b32a98597711c16776cd4d941c) is no longer sufficient to build containers on macOS when using a recent `nixpkgs`. I've applied that patch to my own NixOps fork and when I try to build a container, I get a failure here:

https://github.com/NixOS/nixpkgs/blob/6ea4c3d335c9c7cf121125f41820bb00aa3043b2/pkgs/top-level/stage.nix#L154

So for some reason, NixOps is trying to evaluate `pkgsi686Linux` and I assume that, because the host platform is not Linux, this assertion is failing. I'm guessing this has something to do with the cross-compiling support not working properly with containers and remote builds from a Mac. (It's strange that a container built for an `x86_64-linux` would try to evaluate `pkgsi686Linux`, though, no?)

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #419, opened 2016-04-04
Use preferLocalBuild for decompressing the VirtualBox image

nathan7:
Redirecting this to a build server is both pointless and *incredibly*
slow.


domenkozar:
we should also set `allowSubstitutes = false`

nathan7:
@domenkozar Done!

domenkozar:
@rbvermaa any objections?

PR #423, opened 2016-04-12
deployment: Fix typo.

MostAwesomeDude:
Found while destroying some GCE resources. Guess this code hasn't been run in a while.


rbvermaa:
Thanks, will check why our tests didn't catch this, as we do have a test for when rollback is enabled.

Issue #429, opened 2016-05-11
Hetzner improvements wishlist


danbst:
I've been deploying to hetzner root server last day (with multiple failures) and would like to leave some feedback.

1 When redeploying kickstart partitioning, message `Connection to MY_IP closed by remote host.` injects into output log. This was frightening at first, because initial deploy had no such 

```
hetznerIron> installing machine...
hetznerIron> rebooting machine в??hetznerIronв?T (MY_IP) into rescue system
hetznerIron> sending reboot command... done.
hetznerIron> waiting for rescue system...Connection to MY_IP closed by remote host.
[down]..............................................................[up]
```

2: Default kickstart script doesn't work for new NVMe servers, default disks are not `sda,sdb`, but `nvme0n1,nvme1n1`
3: No example for LVM over RAID, and currently I couldn't find out how to do this
4: nixpart exceptions are frightening too, though not critical. See also https://github.com/aszlig/nixpart/issues/8
5: When nixpart exception occurs, I'd like a prompt "Would you like to enter rescue shell?".
5.1 NixOps does smth to my server, so I cannot access it. I go to Robot interface, reboot (or reboot in rescue) and can login. But after that nixops cannot login. To fix NixOps, I do this:
`nixops destroy -d hetzner`. Now nixops will create new robot account, with new root password.
5.2 NixOps does smth to my server, so I cannot access it. I figured out, I can `nixops export -d hetzner` and grep for password. I think password should be shown after nixpart error (ssh connection string is shown, this is good). But I am constantly running into security violations with `.ssh/known_hosts`, so manual logging into shell is kind of pain every time.
6: Imagine nixpart failed. You fix kickstart config and rerun deploy. Why NixOps reboots instance into rescue shell again? This takes time. There should be an option (or default) to reuse existing session for partitioning operations in rescue shell. I've done redeploy 20 times (always failed), and each time waited for instance reboot to be notified that partitioning failed.
7: BTW, I couldn't get LVM partitions with nixpkgs-15.09. I'm investigating into it currently (trying to build from nixops and nixpkgs master)
8: If instance reboot is nessecary, maybe build system BEFORE reboot? When building system fails (because of upstream problem, for example), it's annoying to wait system boot and see `error: build of "/nix/store/akrjra7z0gi4b42avr52hnzq25jbg6xz-hetzner-nixops-installer.drv" failed`.
9: (Not nixops complain, but maybe somebody knows) I didn't get bootable instance with this partitions and nixpkgs-15.09

[...]

PR #433, opened 2016-05-12
hetzner: do not reboot instance after partitioning failed

danbst:
There may be errors during partitioning phase (incorrect configuration, problems with devices, etc), but they are rarely critical to reboot rescue shell.
With this change time to redeploy after partition error is reduced by 1 minute.

I still do bootstrapping phase, because nixpart version could be changed

Also I'd like to ask nixops devs:
1. Is it correct place to define new state?
2. Bootstrapping phase doesn't remove old bootstrapping tools, just overwrites them. I see, there is check on whether space is enough. So with this change redeploy after error can fail on very-very small instances (where total space in rescue tmpfs is less them 700Mb). Though it affects only "paritioning exception" scenario, so no tests would be affected

part of https://github.com/NixOS/nixops/issues/429

### DIFF
--- a/nixops/backends/hetzner.py
+++ b/nixops/backends/hetzner.py
@@ -67,6 +67,8 @@ class HetznerState(MachineState):
     def get_type(cls):
         return "hetzner"
 
+    PARTITIONING_FAILED = 1000
+
     state = attr_property("state", MachineState.UNKNOWN, int)
 
     main_ipv4 = attr_property("hetzner.mainIPv4", None)
@@ -282,7 +284,11 @@ class HetznerState(MachineState):
                     self.reboot_rescue(install, partitions)
                     return
                 else:
+                    self.state = self.PARTITIONING_FAILED
                     raise
+            except:
+                self.state = self.PARTITIONING_FAILED
+                raise
 
             # This is the *only* place to set self.partitions unless we have
             # implemented a way to repartition the system!
@@ -338,12 +344,12 @@ class HetznerState(MachineState):
         server = self._get_server_by_ip(self.main_ipv4)
         server.rescue.activate()
         rescue_passwd = server.rescue.password
-        if hard or (install and self.state not in (self.UP, self.RESCUE)):
+        if hard or (install and self.state not in (self.UP, self.RESCUE, self.PARTITIONING_FAILED)):
             self.log_start("sending hard reset to robot... ")
             server.reboot('hard')
         else:
             self.log_start("sending reboot command... ")
-            if self.state == self.RESCUE:
+            if self.state == self.RESCUE or self.state == self.PARTITIONING_FAILED:
                 self.run_command("(sleep 2; reboot) &", check=False)
             else:
                 self.run_command("systemctl reboot", check=False)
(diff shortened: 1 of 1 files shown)


rbvermaa:
@aszlig Do you have any opinions on this?

nixops/backends/hetzner.py
```diff
@@ -338,12 +344,12 @@ def reboot_rescue(self, install=False, partitions=None, bootstrap=True,
...
         server.rescue.activate()
         rescue_passwd = server.rescue.password
-        if hard or (install and self.state not in (self.UP, self.RESCUE)):
+        if hard or (install and self.state not in (self.UP, self.RESCUE, self.PARTITIONING_FAILED)):
             self.log_start("sending hard reset to robot... ")
             server.reboot('hard')
         else:
             self.log_start("sending reboot command... ")
-            if self.state == self.RESCUE:
+            if self.state == self.RESCUE or self.state == self.PARTITIONING_FAILED:
```
aszlig: Typo: `,`

nixops/backends/hetzner.py
```diff
@@ -591,7 +597,11 @@ def create(self, defn, check, allow_reboot, allow_recreate):
 
         if not self.vm_id:
             self.log("installing machine...")
-            self.reboot_rescue(install=True, partitions=defn.partitions)
```
aszlig: The same can be achieved by just checking for the RESCUE state here.

aszlig:
I originally did this because sometimes when `blivet` has failed it could end up in an unrecoverable state, so I'm not sure whether it makes sense to reuse the rescue from that state. Maybe it would make sense to add a flag or even just emit a warning telling the user what to do if the he/she wants to throw away the current state.

danbst:
@aszlig Thanks for review!

1 > blivet has failed it could end up in an unrecoverable state
What kind of unrecoverable state? Did rescue reboot fix it? I ask, because I've caught the "unrecoverable state" that even rescue reboot didn't fix (caused by manually configured LVM over RAID). Even `nixops destroy` didn't fix it

2 > or even just emit a warning telling the user what to do if the he/she wants to throw away the current state
I do like this approach!
```
hetznerIron> installing machine...
hetznerIron> rescue system is present and accessible
hetznerIron> partitioning...
hetznerIron> Exception: partitioning failed! Try one of these before redeploy:
  1) Check your kickstart script.
  2) Run "destroy" command to restart rescue system.
  3) Login to rescue with password XXXX and try to fix problems manually.
```

3 This isn't related to my changes, but there is an easy way to brick Nixops rescue mode. Just press Ctrl-C during rescue reboot
```
hetznerIron> installing machine...
hetznerIron> rebooting machine в??hetznerIronв?T (MY_IP) into rescue system
hetznerIron> sending reboot command... done.
hetznerIron> waiting for rescue system...Connection to MY_IP closed by remote host.
[down]....^C
$
```
Now NixOps cannot reboot instance (wrong password). I think this is because of lines https://github.com/danbst/nixops/blob/7c60f056f2438c58796c9dc5dd4bc1030bfddcdf/nixops/backends/hetzner.py#L357-L358
```
        self._wait_for_rescue(self.main_ipv4)
        self.rescue_passwd = rescue_passwd
```
I can fill issue for this if reboot on bootstrap problems will be left.

nixops/backends/hetzner.py
danbst: I was too cautious here =)

domenkozar:
Maybe it just could be a flag to enable this?

danbst:
@domenkozar unfortunately, I have no free Hetzner machines to further maintain this PR.

PR #440, opened 2016-05-29
compute-images-packages: hash, fetchFromGitHub

anderspapitto:
the existing hash was broken, so I fixed it and upgraded to fetchFromGitHub along the way


anderspapitto:
incidentally, the hash being broken was preventing me from using nixops 1.3.1, as found in nixos-unstable. It might be worth cutting a new release.

domenkozar:
Note: this hash differs only on master, 16.03 has the old hash. But since you're now using `fetchFromGitHub` that shouldn't be relevant anymore.

PR #446, opened 2016-06-04
Virtualbox: boot.loader.grub.timeout was renamed to boot.loader.timeout

teozkr:


domenkozar:
This will break 16.03

teozkr:
Oh, sorry, didn't think about that.

domenkozar:
We can merge once 16.09 is released.

ckauhaus:
It think 16.03 has been released, so it should probably be merged now.

danbst:
Actually, there are more of this:
```
nixops/nix/libvirtd-image.nix:      boot.loader.grub.timeout = 0;
nixops/nix/hetzner.nix:    boot.loader.grub.timeout = 1;
nixops/nix/virtualbox-image-nixops.nix:  boot.loader.grub.timeout = 1;
nixops/nix/libvirtd.nix:    boot.loader.grub.timeout = 0;
```

edolstra:
Changing these is problematic because it means we won't be able to deploy older NixOS versions anymore. I guess we need some way to make option definitions conditional on the NixOS version. However, the module system AFAIK doesn't allow that directly. E.g. you can't say
```
config = mkIf (... version >= 16.09 ...) {
  boot.loader.timeout = 1;
};
```
because the options under a `mkIf` still need to exist even if their value is not used.

teozkr:
As a compromise, would it be viable/possible to add `boot.loader.timeout` as an alias option for older branches of Nixpkgs?

That way, you'd still be unable to deploy older builds of those branches, but if they're pegging Nixpkgs to a specific commit then it shouldn't be too big of a deal to peg Nixops as well.

edolstra:
Yeah, that would be fine.

domenkozar:
This exact occurance was already fixed by @rbvermaa and the code already does the correct thing in 16.03, so I'll close this and fix the remaining occurances.

PR #447, opened 2016-06-06
libvirt: use /root/.Xauthority

domenkozar:
Currently libvirt backend doesn't work on NixOS because of X permissions.

I found a solution to get it working, but it would be NixOS specific.

It works together with `$ xauth extract /root/.Xauthority ":0.0"` before libvirtd startup.


domenkozar:
cc @edolstra @rbvermaa

domenkozar:
I think I'll just enable headless by default and issue an warning if it's false.

Issue #434, opened 2016-05-17
Problem using custom packages in custom defined modules.


spinus:
I have few custom packages and custom modules.

In nixops I have custom packages defined as:
```
pkgs = (import <nixpkgs> {}) // { my-package = /*...*/ };
```
When I use want to install my custom package defined in nixops config file (for example in a service or systemPackages), that works.
When I want to use custom package in module, that doesn't work (attribute with custom package is not found).

I import modules using ``imports = [ ./path/to/module.nix ];`` syntax.
I think this does not work, because ``pkgs`` is not passed correctly to ``eval-config``, but I'm not sure how to pass it or make it working the other way.

ip1981:
`nixpkgs.config.packageOverrides` should do the trick. Use it in a separate module.

spinus:
@ip1981 I think it's not that. I don't know how to pass modified "pkgs" to imported modules.
So even if I tried what you write, modified version of pkgs is not passed correctly. 
Is there any nixops way to maybe use modified pkgs from my-packages as default "pkgs"?

More or less my network.nix
```
{system}:{

host1 = 
let pkgs = import ./my-packages.nix {inherit system;}
in {
  imports = [
    ./module1.nix # when package is used inside module it does not work
  ];
  environment.systemPackages = [pkgs.my-package]; # <- that works
  services.my-service.enable = true; # <- does not work, because my-package is not available inside my-module.nix
}
};
```

spinus:
Or maybe there is another way of providing custom packages/services in nixops (without forking nixpkgs)?

spinus:
@ip1981 Finally I understand what did you mean and where to put it (at the beginning I've put that on network level rather then host level). Works perfectly. Thank you.

working example would be:
```
{system}: {
  host1 = {
    imports = [ ./module1.nix ];  # now module can use packages overriden/declared below
    nixpkgs.config.overridePackages ....
  }
}
```
Also, nixpkgs can be defined inside a module, so you can group packages in separate module.

I think would be nice to add this information to manual.

domenkozar:
This has nothing to do with NixOps, but rather NixOS which is already documented. I don't think we should duplicate documentation from NixOS since 99% of it applies for NixOps.

Do you have some suggestions for the manual for this to be more clear @spinus ?

k0001:
[...]

PR #450, opened 2016-06-07
#434 add documentation about custom modules

spinus:


domenkozar:
This is all pure NixOS. I'd rather have a well documented section in NixOS manual and link from NixOps there to say what's all possible in NixOS is also possible in NixOps.

rbvermaa:
I agree with Domen that this is really NixOS related. Closing the PR. Please submit documentation for NixOS in the nixpkgs repository.

spinus:
I think there is a section in nixpkgs about that stuff. My idea here was to give the nixops user answers earlier on his road without hours/days of digging into related projects. Kind of a FAQ way.

But that's ok, I think probably somebody can find the answers anyway.

PR #482, opened 2016-08-02
Fallback to EC2/AWS env variable if access key id is not defined

lo1tuma:
According to the documentation the environment variable `EC2_ACCESS_KEY` or `AWS_ACCESS_KEY_ID` should be used if the `deployment.route53.accessKeyId` has not been defined.


copumpkin:
Why manually handle access keys like this when boto automatically handles the standard environment variables/config files/instance metadata fallthrough for you? If you do it yourself you need to manually pass around session tokens if you want to use temporary credentials, and various ugliness like that.

Issue #475, opened 2016-07-19
None backend - ssh key problems


spinus:
I can't deploy the box. Nix expression is almost empty. NixOps versions I tried are 1.3.1 and 1.4. Nixpkgs on host machine is unstable, on the virtual machine nixos-version = 16.03.1083.d745044 (Emu).

I can ssh to the box using ssh key.

```
nixops deploy
jenkins-slave1> generating new SSH keypair... done
Enter passphrase for key '/home/spinus/.ssh/id_rsa': 
Host key fingerprint is SHA256:aS+mlf1IQUQRygUL7il89qISapZc1IP4yaDbgFh+JwE
+--[ED25519 256]--+
|      . .o*o     |
|  E  . o =       |
|  ..o . + .      |
| o.+.+ . o       |
|+o= +.* S .      |
|= o=o+.o + .     |
|.=oo o. * +      |
|o=o  . = o o     |
|o  .. .   . .    |
+----[SHA256]-----+
jenkins-slave1> setting state version to 16.03
jenkins-slave1> waiting for SSH................................................
```

So, looks like in the first step nixops could connect to the machine and put ssh key there but in the second step stuff it's not working :[

I can ssh to the box using ssh and nixops ssh

domenkozar:
I saw something similar in the past. I think this is happening due to `_ssh_public_key_deployed` modified in https://github.com/NixOS/nixops/pull/364.

I long knew keeping such states would just cause trouble on the long run. You'll have to find out what really happened there and report it so we can fix it.

spinus:
@domenkozar any idea what's the best solution to debug it? I tried to add --debug but it's not that useful, no debug logging. Best way is to add logging statements to the code or run the code through the debugger or something else?

domenkozar:
@spinus I think it's best to place`import pdb;pdb.set_trace()` inside that code and see what's going on. Debugging isn't verbose enough.

spinus:
Thanks. I'll try.

dezgeg:
Are you by any chance trying to deploy using a hostname that is just an alias in `~/.ssh/config`? I found out that won't work as nixops tries to explicitly that `targetPort` on `targetHost` is open.

spinus:
@dezgeg yes, I'm doing that. I have to do that as I'm connecting to this vm through ssh-proxy/ssh-hop.
It's interesting as ssh command is working, so that would mean that each "command" is implementing ssh connection differently.

dezgeg:
[...]

PR #486, opened 2016-08-05
#475 wait for ssh using ssh

spinus:


spinus:
Should I run some tests before we can progress with it? Or the test are running automatically?

domenkozar:
@spinus would you explain why waiting on TCP port is not enough? Why do we need ssh?

spinus:
General answer: becasue SSH is "subset" of TCP and checking TCP is not enough.

Detailed answer: as I explained in https://github.com/NixOS/nixops/issues/475#issuecomment-234357671, I'm using SSH hop server to access target server. There is no way that tcp check can detect that.

rbvermaa:
@edolstra What do you think?

nixops/backends/__init__.py
```diff
@@ -259,12 +260,26 @@ def wait_for_ssh(self, check=False):
...
-        self.ssh_pinged = True
-        self._ssh_pinged_this_time = True
+
+        n = 0
+        timeout = 60
+
+        while True:
+            try:
+                self.ssh.run_command("exit", self.get_ssh_flags(),
+                    logged=False, timeout=timeout)
```
domenkozar: This timeout value changes behaviour, it used to be 1 second, now it's 60.

spinus: yes. Also this changed introduced "global" timeout, after 60 tries it stops, when previous method was checking forever I think. But those are details to this change and I can adjust it as you prefer, for me the core idea here is to use SSH.

spinus:
@edolstra would you take a look, please?

domenkozar:
Looks good, someone just need to test this and confirm it works:

- by deploying any backend and machine comes up
- by deploying any backend and breaking networking so it times out

spinus:
@domenkozar I tested but assume it does not count :-)

Do you think if I try to write a test using virtualbox backend, or even better standard nixos test with qemu and "none" nixops backend (to mirror live scenario using none backend) and include here. Do we have any ability to run those tests as part of nixops merge/release?

domenkozar:
Tests needing IO are ran manually before each release, so very welcome to add them :)

wizeman:
I'm not familiar with the nixops code, but I think the fix in this PR is not quite right.

I think it should be changing the `wait_for_tcp_port()` and `ping_tcp_port()` functions in `nixops/util.py`, instead of `wait_for_ssh()`. As far as I can see, all uses of the `*_tcp_port()` functions are for checking whether SSH is working.

For me, `ping_tcp_port()` is not working because my target host is IPv6, so a few nixops commands don't work for me. This PR could fix the IPv6 problem, and `nixops deploy` indeed works but `nixops check` doesn't correctly detect the hosts as being up, and I bet `nixops deploy --force-reboot` doesn't either, because `reboot_sync()` calls `wait_for_tcp_port()` instead of `wait_for_ssh()`.

There are other uses of `wait_for_tcp_port()` in some backends which I think will also not be fixed correctly with this PR.

@spinus: Does `nixops check` detect your target server as being up (the one that requires the SSH hop)? Does `nixops deploy --force-reboot` correctly detect your server as being down and then being back up?

spinus:
@wizeman I'll try to refactor and test as you suggested, not sure when

spinus:
@wizeman actually, I just realized that I used wait_for_ssh because the tunneling logic is on the ssh layer. TCP layer probably should not know anything about ssh tunneling, right?

wizeman:
@spinus: Sure, but I would not call a couple of *_tcp_port() functions a TCP _layer_...

Since all uses of the *_tcp_port() functions seem to be checking for SSH, I would just rename those functions to ping/wait_for_ssh_port(), and make sure your changes go in those functions instead. Using your terminology, this would effectively be eliminating the TCP layer, and replacing it with an SSH layer.

In fact, if in the future nixops ever needs to check for an open TCP port (other than the SSH port) it would probably have to go through SSH anyway, as in your use case you can only access the server via an SSH hop server. So even if you decided to keep these TCP functions, they would have to be reimplemented to support your use case.

But I am not actually a member or maintainer of this project so I'm not sure if the maintainers agree with me...

sorki:
I've tried this with recent nixops and it resolves my issue with ipv6 only targets and none backend.

nixops check reports Up: No though.

domenkozar:
Fixed in https://github.com/NixOS/nixops/pull/735

PR #493, opened 2016-08-24
virtualbox: use VBoxHeadless and allow dots in machine names

danbst:
VBoxHeadless --startvm gives much better error message in case kernel modules
are not installed (typical situation when you install virtualbox
with nix-env -i, and not via virtualization.virtualbox.host.enable)

Also, VBoxHeadless doesn't depend on qt, so this is related to
https://github.com/NixOS/nixpkgs/issues/8834


domenkozar:
Thanks!

PR #505, opened 2016-09-08
Fix issue with ssh flags used with scp.

jraygauthier:
As the '-p' flag attempt to preserve file attributes
with scp, it was making the scp command fail in
contexts where preserving attributes is not possible
(this is without mentionning the '22' argument that
followed and which seems to be accepted silently).

Here's the error it fixes:

```
afrontend> uploading key ‘my_secret.key’...
afrontend> scp: /run/keys/my_secret.key: No such file or directory
afrontend> error: command ‘['scp', '-P', '22', '-i', '/run/user/1000/nixops-tmp4MXxgW/id_nixops-afrontend', '-p', '22', '-oControlPath=/run/user/1000/nixops-ssh-tmpakFvsQ/master-socket', '/run/user/1000/nixops-tmp4MXxgW/key-afrontend', u'root@10.233.1.2:/run/my_secret.key']’ failed on machine ‘afrontend’ (exit code 1)
error: activation of 1 of 1 machines failed (namely on ‘afrontend’)
```

See the `['-P', '22']` part followed not long after by wrong `['-p', '22']`.

Note that this is a prerequisite for #300

### DIFF
--- a/nixops/backends/container.py
+++ b/nixops/backends/container.py
@@ -63,9 +63,9 @@ class ContainerState(MachineState):
         flags = super(ContainerState, self).get_ssh_flags(*args, **kwargs)
         flags += ["-i", self.get_ssh_private_key_file()]
         if self.host == "localhost":
-            flags.extend(MachineState.get_ssh_flags(self))
+            flags.extend(MachineState.get_ssh_flags(self, *args, **kwargs))
         else:
-            cmd = "ssh -x -a root@{0} {1} nc -c {2} {3}".format(self.get_host_ssh(), " ".join(self.get_host_ssh_flags()), self.private_ipv4, self.ssh_port)
+            cmd = "ssh -x -a root@{0} {1} nc -c {2} {3}".format(self.get_host_ssh(), " ".join(self.get_host_ssh_flags(*args, **kwargs)), self.private_ipv4, self.ssh_port)
             flags.extend(["-o", "ProxyCommand=" + cmd])
         return flags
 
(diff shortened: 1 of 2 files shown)


domenkozar:
This doesn't call `get_ssh_flags` anymore, which is needed sometimes to pass the private ssh key

jraygauthier:
Good point. It seems that default options were working fine for the localhost case. Bringing back
`get_ssh_flags` uncovered supplementary arguments propagation errors in the container backend.
This one should do the trick.

nixops/backends/container.py
```diff
@@ -63,9 +63,9 @@ def get_ssh_flags(self, *args, **kwargs):
         flags = super(ContainerState, self).get_ssh_flags(*args, **kwargs)
         flags += ["-i", self.get_ssh_private_key_file()]
         if self.host == "localhost":
-            flags.extend(MachineState.get_ssh_flags(self))
+            flags.extend(MachineState.get_ssh_flags(self, *args, **kwargs))
```
domenkozar: Why is this needed?

jraygauthier: so that [MachineState.get_ssh_flags](https://github.com/jraygauthier/nixops/blob/aa967926bfe5668a869fe0e714037c8de2e90548/nixops/backends/__init__.py#L233) gets its expected `scp` flag as specified by the caller. Otherwise it will default to `False`.

I honestly find these var args parameters confusing, I'm not sure why there was a need for it in the first place but I tried to make do with it.

domenkozar:
@jraygauthier could you rebase? I think we can merge this.

jraygauthier:
@domenkozar This PR is no longer relevant now that #804 fixes the problem.

PR #522, opened 2016-10-09
AWS SNS integration

AmineChikhaoui:
Adding an AWS sns resource.
Some limitations that exist currently with this implementation: we can't remove a subscriber if it's removed from the deployment definition, we can handle that later I think.
Also generally subscriptions and sns policies can be resources themselves in a way that someone could need to deploy subscribers/policies to existent topics but for now I'm only handling the deployment of a self contained topic definition.

**Update** I added the ability to remove subscriptions per @rbvermaa request.

### DIFF
--- a/examples/sns-topic.nix
+++ b/examples/sns-topic.nix
@@ -1,0 +1,20 @@
+{ account
+, region ? "us-east-1"
+, description ? "SNS topic example"
+, ...
+}:
+{
+  network.description = description;
+
+  resources.snsTopics.example-topic = {
+	 name="sns-topic";
+	 displayName="Nixops SNS topic";
+	 accessKeyId = account;
+	 subscriptions = [
+		{
+		 protocol = "email";
+		 endpoint = "amine@chikhaoui.tn";
+		}
+	];
+	inherit region; };
+}
--- a/examples/sns-topic.nix
+++ b/examples/sns-topic.nix
@@ -7,14 +7,15 @@
   network.description = description;
 
   resources.snsTopics.example-topic = {
-	 name="sns-topic";
-	 displayName="Nixops SNS topic";
-	 accessKeyId = account;
-	 subscriptions = [
-		{
-		 protocol = "email";
-		 endpoint = "amine@chikhaoui.tn";
-		}
-	];
-	inherit region; };
+    name = "sns-topic";
+    displayName = "Nixops SNS topic";
+    accessKeyId = account;
+    subscriptions = [
+    {
+      protocol = "email";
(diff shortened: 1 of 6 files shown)


edolstra:
This PR contains a bunch of commits/reverts that don't seem to belong here (e.g. https://github.com/NixOS/nixops/pull/522/commits/4e8101cf226d17da372cf7250895c3d816f1d23d). Can you squash the PR into a single commit to get rid of them? (`git rebase -i`)

AmineChikhaoui:
done @edolstra

examples/sns-topic.nix
```diff
@@ -0,0 +1,20 @@
+{ account
+, region ? "us-east-1"
+, description ? "SNS topic example"
+, ...
+}:
+{
+  network.description = description;
+
+  resources.snsTopics.example-topic = {
+	 name="sns-topic";
```
edolstra: No tabs please, 2 spaces per indentation level. Also, spaces around "=" please.

nixops/resources/sns_topic.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_type(cls):
+        return "sns-topic"
+
+    @classmethod
+    def get_resource_type(cls):
+        return "snsTopics"
+
+    def __init__(self, xml):
+        nixops.resources.ResourceDefinition.__init__(self, xml)
+        self.topic_name = xml.find("attrs/attr[@name='name']/string").get("value")
```
AmineChikhaoui: ah great. Yeah looks much cleaner, will do that.

nixops/resources/sns_topic.py
```diff
@@ -0,0 +1,181 @@
...
+            self._destroy()
+            self._conn = None
+
+        self.region = defn.region
+        self.connect()
+
+        if self.arn == None:
+            self.log("creating SNS topic ‘{0}’...".format(defn.topic_name))
+            topic = self._conn.create_topic(defn.topic_name)
+            arn = topic.get('CreateTopicResponse').get('CreateTopicResult')['TopicArn']
```
edolstra: What happens if we're interrupted after the `create_topic` AWS call, but before `self.depl._db` is updated? Will that leak a topic? If so, is there a way to make it idempotent (e.g. using a creation token like for EC2 instance creation)?

nixops/resources/sns_topic.py
AmineChikhaoui: @edolstra I removed the xml stuff and tested again. I also added one missing edge case handling which is when the sns topic is removed from the console so that it gets created again.

PR #526, opened 2016-10-19
Add support for ~/.aws/credentials

### LINKED ISSUE #396: Support ~/.aws/credentials
NixOps has an ad hoc configuration file `~/.ec2-keys` for AWS credentials. However, there is a standard mechanism for storing AWS credentials: `~/.aws/credentials` (http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html). So it would be nice to use that instead.

aske:
Fixes #396.

Now you can use a standard way to store credentials in `~/.aws/credentials`.
This default location can be changed by setting the `$AWS_SHARED_CREDENTIALS_FILE` variable.

Previously search for access keys was conducted first in `~/.ec2-keys`, then in the environment variables.
Now the priority order looks like this: `~/.ec2-keys` > `~/.aws/credentials` > env variables.

Also not directly related, but adding new option to decouple an actual `ec2.accessKeyId` from useful symbolic names could be a good idea (see discussion in the issue).

### DIFF
--- a/nixops/ec2_utils.py
+++ b/nixops/ec2_utils.py
@@ -11,32 +11,50 @@ from boto.exception import EC2ResponseError
 from boto.exception import SQSError
 from boto.exception import BotoServerError
 
-def fetch_aws_secret_key(access_key_id):
-    """Fetch the secret access key corresponding to the given access key ID from the environment or from ~/.ec2-keys"""
-    secret_access_key = os.environ.get('EC2_SECRET_KEY') or os.environ.get('AWS_SECRET_ACCESS_KEY')
-    path = os.path.expanduser("~/.ec2-keys")
-    if os.path.isfile(path):
-        f = open(path, 'r')
-        contents = f.read()
-        f.close()
-        for l in contents.splitlines():
-            l = l.split("#")[0] # drop comments
-            w = l.split()
-            if len(w) < 2 or len(w) > 3: continue
-            if len(w) == 3 and w[2] == access_key_id:
-                access_key_id = w[0]
-                secret_access_key = w[1]
-                break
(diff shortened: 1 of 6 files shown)


ip1981:
Yes!

nixops/ec2_utils.py
```diff
@@ -11,32 +11,51 @@
...
+    """
+        Fetch the secret access key corresponding to the given access key ID from ~/.ec2-keys,
+        or from ~/.aws/credentials, or from the environment (in that priority).
+    """
 
-    return (access_key_id, secret_access_key)
+    def parse_ec2_keys():
+        path = os.path.expanduser("~/.ec2-keys")
+        if os.path.isfile(path):
+            f = open(path, 'r')
```
domenkozar: Best to use `with open(path, 'r') as f:`

nixops/ec2_utils.py
```diff
@@ -11,32 +11,51 @@
...
+                if len(w) == 3 and w[2] == access_key_id: return (w[0], w[1])
+                if w[0] == access_key_id: return (access_key_id, w[1])
+        return None
+
+    def parse_aws_credentials():
+        path = os.getenv('AWS_SHARED_CREDENTIALS_FILE', "~/.aws/credentials")
+        conf = Config(os.path.expanduser(path))
+
+        if access_key_id == conf.get('default', 'aws_access_key_id'):
+          return (access_key_id, conf.get('default', 'aws_secret_access_key'))
```
domenkozar: identation is 4 spaces

aske:
@domenkozar Fixed, thanks!

PR #540, opened 2016-11-02
virtualbox.{nix,py}: add .vcpu option

Cireo:
We follow the naming convention used by libvirt, and convert this
option to a "--cpu" command line option for VBoxManage.

### DIFF
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -62,6 +62,7 @@ to the configuration. Otherwise, VirtualBox will fail when it tries to open a gr
     { config, pkgs, ... }:
     { deployment.targetEnv = "virtualbox";
       deployment.virtualbox.memorySize = 1024; # megabytes
+      deployment.virtualbox.vcpu = 2; # number of cpu
     };
 }
 </programlisting>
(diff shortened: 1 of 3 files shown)


domenkozar:
I wonder if we should instead just add something like `deployment.virtualbox.vmFlags` that would pass everything to `VBoxManager modifyvm ${vmFlags}`. Otherwise we'll end up with dozens of flags eventually.

Cireo:
To be honest, I think the total number of *common* flags could be relatively small (less than a dozen), and this offers a good way to be consistent across multiple virtualization options.  I suppose this is a tradeoff between easily switching deployment providers (so you only have to learn once) and having configuration options correspond to the vendor implementation (which makes it easier if you already know your vendor).

That being said, I think a `vmFlags` would be a nice addition to ensure there is always some way to pass things in, and to make whatever customization options you want.

Cireo:
Should I send up another commit that augments or replaces this commit with a `vmFlags` option?

domenkozar:
It's not ideal to override vcpu to be 1 by default. It would be better to not have a default and only pass the number if it's set.

domenkozar:
@Cireo if you make that change I'll merge

Cireo:
@domenkozar sorry for the delays.  I have rebased against the latest nixops master, and added the options you suggested.

domenkozar:
Thanks!

Issue #554, opened 2016-11-22
multiple libvirt/qemu machines deployed locally leads to ssh connection fail race condition


NotBrianZach:
I'm trying to deploy locally on nixos using nixops and getting a random failure trying to connect to the qemu vms.
I have tried running nixops deploy with dev-shell of a fork of the most recent nixops (as of this moment), same result (as the earlier nixops v 1.4).
```
Warning: Permanently added '192.168.122.38' (ED25519) to the list of known hosts.
postgresqlServer> setting state version to 17.03
postgresqlServer> waiting for SSH...
Warning: Permanently added '192.168.122.201' (ED25519) to the list of known hosts.
Permission denied (publickey,password,keyboard-interactive).
apiServer.......> could not connect to ‘root@192.168.122.201’, retrying in 1 seconds...
Warning: Permanently added '192.168.122.156' (ED25519) to the list of known hosts.
Permission denied (publickey,password,keyboard-interactive).
renderServer....> could not connect to ‘root@192.168.122.156’, retrying in 1 seconds...

```

nixops info looks like this (though sometimes it's only one/none/three machines unknown)
```
+------------------+----------------------+----------+-------------+-----------------+
| Name             |        Status        | Type     | Resource Id | IP address      |
+------------------+----------------------+----------+-------------+-----------------+
| apiServer        | Unknown / Up-to-date | libvirtd |             | 192.168.122.201 |
| postgresqlServer |   Up / Up-to-date    | libvirtd |             | 192.168.122.38  |
| renderServer     | Unknown / Up-to-date | libvirtd |             | 192.168.122.156 |
+------------------+----------------------+----------+-------------+-----------------+
```
edit: still looking into this, I think it might be similar problem to https://github.com/NixOS/nixops/pull/364/files

NotBrianZach:
looks like maybe
https://github.com/NixOS/nixops/blob/master/nixops/backends/libvirtd.py#L87
could be culpable if there's multithreading involved, could get overridden perhaps?
currently trying to write a patch that passes env directly, see if that changes anything

PR #557, opened 2016-11-25
libvirt/qemu race condition fix.

NotBrianZach:
addresses #554, works on my box, there are two issues with this fix I think, 
1. security, since we're invoking shell to pass environment, if someone hijacks args we pass in, could be worse than just using execve or whatever directly
 
Can't tell if there is another way to pass the env directly, there is an env attribute to the Popen call, but passing it overrides the environment as far as I can tell, which has quite a bit of state.

2. related to 1, invoking shell leads to slightly more spam at startup I think where it tells you what .nix files it's evaluating.

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -2,6 +2,7 @@
 
 from distutils import spawn
 import os
+import copy
 import random
 import string
 import subprocess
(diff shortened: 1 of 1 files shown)


domenkozar:
Why not just pass the variable via `env` parameter? https://docs.python.org/2/library/subprocess.html#subprocess.Popen

NotBrianZach:
just double checked this, I get the following error simply passing in env with public key to _logged_exec, I believe this is because (and correct me if I'm wrong) env overrides the environment, instead of simply appending any supplied variables/values.
[nix-shell:~/nixops]$ nixops deploy
trace: warning: The option `boot.loader.grub.timeout' defined in `<unknown-file>' has been renamed to `boot.loader.timeout'.
trace: warning: The option `boot.loader.grub.timeout' defined in `<unknown-file>' has been renamed to `boot.loader.timeout'.
trace: warning: The option `boot.loader.grub.timeout' defined in `<unknown-file>' has been renamed to `boot.loader.timeout'.
error: Multiple exceptions: [Errno 2] No such file or directory, [Errno 2] No such file or directory, [Errno 2] No such file or directory

Issue #515, opened 2016-10-04
Support Digital Ocean


bachp:
[Digital Ocean](https://www.digitalocean.com/) has nice servers for a good price.
It would be a good candidate for an additional backend.

The also have a pretty good [API](https://developers.digitalocean.com/documentation/v2/)

yotann:
They don't support NixOS or custom distros. I'm sure it's possible to install Debian and overwrite it with NixOS, like [digitalocean-debian-to-arch](https://github.com/gh2o/digitalocean-debian-to-arch) does for Arch Linux, but it's a bit of a hack.

[Vultr](https://www.vultr.com/) is a very similar service which *does* support custom distros, and they have an API.

bachp:
There is also [nixos-infect](https://github.com/elitak/nixos-infect).

domenkozar:
I'm going to close this as DigialOcean doesn't support "upload an image" API. We can have an issue for nixos-infect or similar if wanted/needed.

domenkozar:
https://github.com/NixOS/nixops/issues/562

PR #562, opened 2016-12-05
Draft Digital Ocean support

teh:
This introduces some basic digital ocean support using nixos-infect. I'm a bit unclear about coding conventions and the resource allocation model in nixops. E.g. I couldn't quite figure out how ssh keys are meant to be allocated, so I'm looking for a hard-coded `ssh-key` value.

Anyway: This works but I appreciate the code will need some work.

See also https://github.com/NixOS/nixops/issues/515


do.nix
```diff
@@ -0,0 +1,10 @@
+{ fetchurl, python2Packages }:
```
domenkozar: We'll have to move this to nixpkgs and backport to 16.09 and 16.03

examples/trivial-digital-ocean.nix
```diff
@@ -0,0 +1,13 @@
+{
+  resources.sshKeyPairs.ssh-key = {};
+
+  machine = { config, pkgs, ... }: {
+    services.nginx.enable = true;
+    networking.firewall.allowPing = true;
```
domenkozar: Is this absolutely necessary to deploy DO? iirc `allowPing` is true by default since 16.03

nixops/backends/digital_ocean.py
```diff
@@ -0,0 +1,169 @@
...
+
+class DigitalOceanState(MachineState):
+    @classmethod
+    def get_type(cls):
+        return "digital-ocean"
+
+    state = nixops.util.attr_property("state", MachineState.MISSING, int)  # override
+    public_ipv4 = nixops.util.attr_property("publicIpv4", None)
+    default_gateway = nixops.util.attr_property("defaultGateway", None)
+    public_dns_name = nixops.util.attr_property("publicDnsName", None)
```
domenkozar: This var is not used anywhere

nixops/backends/digital_ocean.py
```diff
@@ -0,0 +1,169 @@
...
+            self.log("droplet not found - assuming it's been destroyed already")
+        self.public_ipv4 = None
+        self.droplet_id = None
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
+        ssh_key = self.depl.active_resources.get('ssh-key')
+        if ssh_key is None:
+            raise Exception('Please specify a ssh-key resource (resources.sshKeyPairs.ssh-key = {}).')
+
+        if self.droplet_id is not None:
```
domenkozar: (nitpick) log that it's a noop due to already existing resource. Helpful for debugging.

nixops/backends/digital_ocean.py
```diff
@@ -0,0 +1,169 @@
...
+        # Not sure when I'd have more than one interface from the DO
+        # API but networks is an array nevertheless.
+        self.default_gateway = droplet.networks['v4'][0]['gateway']
+        self.netmask = droplet.networks['v4'][0]['netmask']
+
+        # run modified nixos-infect
+        # - no reboot
+        # - predictable network interface naming (enp0s3 etc)
+        self.wait_for_ssh()
+        self.run_command('bash </dev/stdin 2>&1', stdin=open(infect_path))
```
domenkozar: Log `Running nixos-infect`

do.nix
teh: Created https://github.com/NixOS/nixpkgs/pull/20942

examples/trivial-digital-ocean.nix
teh: No, removed, thanks!

nixops/backends/digital_ocean.py
teh: Removed

nixops/backends/digital_ocean.py
teh: I'm a bit confused about this one: `create` is called on every invocation. So e.g. in ec2 we have the following, analogous bit if code:

```
        # Create the instance.
        if not self.vm_id:
            self.log("creating EC2 instance (AMI ‘{0}’, type ‘{1}’, region ‘{2}’)...".format(
```

So I think not creating a new instance when `vm_id` exists is "normal" operation?

nixops/backends/digital_ocean.py
teh: done

nixops/backends/digital_ocean.py
```diff
@@ -165,5 +164,6 @@ def create(self, defn, check, allow_reboot, allow_recreate):
         # - no reboot
         # - predictable network interface naming (enp0s3 etc)
         self.wait_for_ssh()
+        self.log_start("running modified nixos-infect")
```
domenkozar: In order not to confuse users, I'd rather skip `modified` since it's already mentioned in comments. It's really a developer note rather than nixops user's.

nixops/backends/digital_ocean.py
domenkozar: Good point, let's not spam users.

domenkozar:
This looks like a good start. cc @bachp

cc @edolstra @rbvermaa 

If there are no other objections I'll merge this in a few days. It would be good idea if someone else tested it, but I really don't have time atm and I trust @teh he'll eventually fix issues if they rise up.

teh:
Thanks! I too would be keen on someone else testing before merge.

Note that branch requires nixpkgs-master because of the digitial-ocean dependency.

domenkozar:
I've backported digital-ocean to 16.09, it didn't reach channels yet.

examples/trivial-digital-ocean.nix
```diff
@@ -0,0 +1,12 @@
+{
+  resources.sshKeyPairs.ssh-key = {};
+
+  machine = { config, pkgs, ... }: {
+    services.nginx.enable = true;
+    services.openssh.enable = true;
+
+    deployment.targetEnv = "digital-ocean";
+    deployment.digital-ocean.region = "ams2";
```
edolstra: Maybe this should be `digitalOcean`, since we generally use camelCase in option names.

nix/digital-ocean.nix
```diff
@@ -0,0 +1,47 @@
...
+      type = types.str;
+      description = ''
+        The region. See https://status.digitalocean.com/ for a list
+        of regions.
+      '';
+    };
+
+    deployment.digital-ocean.size = mkOption {
+      example = "512mb";
+      type = types.str;
```
edolstra: It might be nicer to use an integer here (e.g. size in megabytes) rather than an opaque string.

nix/digital-ocean.nix
```diff
@@ -0,0 +1,47 @@
...
+      description = ''
+        The size slug between <literal>512mb</literal> and <literal>64gb</literal>;
```
edolstra: What is a size slug? Also ";" should be ".".

edolstra:
How does `nixos-infect` compare to `nixos-assimilate` and `nixos-in-place`? It would be suboptimal if we have to maintain multiple scripts that do essentially the same thing.

nix/digital-ocean.nix
teh: Updated - does this make more sense?

examples/trivial-digital-ocean.nix
teh: Ah totally right. Updated & tested, apologies.

teh:
AFAICT `nixos-assimilate` is unmaintained. `nixos-in-place` works fine but does more work than we need, e.g. it keeps the old OS around. `nixos-infect` uses the built-in NIXOS_LUSTRATE [1] so it's the least amount of extra code to ship to assimilate a foreign distribution. 

I have no strong feelings about the mechanism we use to take over the original system, happy to replace if there are advantages to another approach.

[1]
https://github.com/NixOS/nixpkgs/pull/17784

teh:
One thing I'm also not sure about is this: Should we enable openssh by default?

domenkozar:
Looks good to me. @edolstra ?

teh:
@domenkozar - I think we should enable openssh by default, otherwise remote nix updates break if the users don't manually enable openssh. What do you think? (That's the last change I'm still sitting on, other than that I've tested a fair bit)

domenkozar:
Yeah, one can still use `mkForce` to disable it if needed.

domenkozar:
I'll merge this in following days if noone else responds.

grahamc:
👍 Super nice and clear example to follow on implementing a new backend. Thank you for this!

mhsjlw:
@teh was this tested with multi-server deployments

teh:
@mhsjlw - no. Did you find any issues? Happy to help and fix if you have a reproducible case.

mhsjlw:
Yep, if you try to deploy multiple servers you send the same SSH key to DigitalOcean each server creation, but you don't make a new SSH key per server, so DigitalOcean rejects the keys and kills server provisioning. Do you want to use 1 SSH key for each server, or do you want to have a separate one for each, either way, you need to make sure that the same key isn't uploaded again.

mhsjlw:
A reproducible case is simply taking the trivial DigitalOcean and just copying it as another server in the same infrastructure file

peterhoeg:
They now support custom ISOs although not via the API (will be ready end this year)

PR #548, opened 2016-11-15
modules: optionSet -> submodule

ericsagnes:
This replace `optionSet` usages by `submodule`. #545 

I am not very familiar with nixops codebase, so please review carefully the changes before merging. 

The tarball is building fine.

cc @domenkozar

PR #576, opened 2016-12-23
Attempt to fix an issue introduced by #548

AmineChikhaoui:
The EC2 deployments evaluations are failing after the optionSet -> submodule PR merge.
```
[nix-shell:~/work/nixops]$ nixops info -d minimal-lb 
error: attribute ‘cipher’ missing, at /home/aminechikhaoui/work/nixops/nix/ec2.nix:461:10
warning: evaluation of the deployment specification failed; status info may be incorrect
```
Removing the types.uniq looks like a possible fix but I wasn't able to understand the issue or why removing that is fixing it.

@domenkozar @rbvermaa any ideas ?

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -412,7 +412,7 @@ in
       options = {
         ec2 = mkOption {
           default = null;
-          type = with types; uniq (nullOr (submodule ec2DiskOptions));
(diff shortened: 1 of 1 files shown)


edolstra:
I don't understand where the error message comes from, but the `uniq` seems unnecessary in any case. So let's merge this.

PR #582, opened 2017-01-10
libvirtd : Allow instanciating vms on a remote host

regnat:
Nixops has a libvirtd backend, but it is only able to instanciate vms
on the local machine.

This pr adds a `deployment.libvirtd.host` option (and also a
`deployment.libvirtd.remote_user` option) to define a vm that will be
instanciated on another machine.


flokli:
This seems to be similar and less flexible than https://github.com/NixOS/nixops/pull/824 (which allows to specify a URI).

regnat:
Agreed, Closing this in favour of #824

PR #585, opened 2017-01-16
libvirt: Allow more complex network setups

Nadrieril:
This PR adds support for more complex network setups in libvirt, notably bridging and static IPs.
It removes the hacky way to get static MACs. People who need a static MAC address can set it in the interface config in deployment.libvirtd.interfaces.
It also adds different methods for detecting the VM's IP address, in case DHCP is not used for example.

### DIFF
--- a/nix/libvirtd.nix
+++ b/nix/libvirtd.nix
@@ -32,6 +32,45 @@ let
         umount /mnt
       ''
   );
+
+  interfaceOpts = {
+    options = {
+      type = mkOption {
+        type = types.str;
+        default = "network";
+        example = "bridge";
+        description = ''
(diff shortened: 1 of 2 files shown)


Nadrieril:
Actually, since bridges can be defined at the network level (https://libvirt.org/formatnetwork.html#examplesBridge) (thanks nekroze@freenode), the whole of deployment.libvirtd.interfaces might not be needed.
It is still useful to add additional options to a given interface, like a mac address though.

Nadrieril:
I'm splitting this PR since I'm not sure about the interfaces part. See https://github.com/NixOS/nixops/pull/586 for the IP detection thing

Issue #500, opened 2016-09-05
libvirtd backend: dhcpcd inside guest not getting ipv4 address via dhcp from host


qknight:
i'm using **nixos 16.03.1271.546618c (Emu) with nixops 1.4** and i wanted to
**deploy to libvirtd**. according to
https://nixos.org/nixops/manual/#idm140737319108016 this should be working.

my problem is, i can't manage to get a IPv4 via dhcpcd from a
kvm/libvirtd-client (guest). 
**NOTE:** i have a manually configued qemu guest as well. dhcp does not work there either but using static ipv4 addresses i can access the internet. nixops wants to use the NAT adapter in contrast to my manually configured qemu guest where i've added a 'bridge name: virbr0' with 'device model: rtl8139'.

i see this:

```
nixops deploy ...
...
mailserver> waiting for IP address to appear in DHCP
leases.........................................................................
.....................................
```

this dottet line never finishes. i've also used virt-manager to deploy a
nixos based guest and using static ip-addresses i can access the
internet from the guest. using dhcpcd from that guest never terminates
successfully either. i see the requests using tcpdump -i virbr0 (on the
host).

this is my host configuration:

```
  ...
  virtualisation.libvirtd.enable = true;
  networking = {
    firewall = {
      allowedTCPPorts = [ 22 80 443 ];
      allowPing = true;
      checkReversePath = false;
    };
  };
```

i'm quite lost now, what could i try to fix this?

best wishes,
joachim

# host networking

```
ip a

3: virbr0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue
state UP group default qlen 1000
    link/ether 52:54:00:50:21:e8 brd ff:ff:ff:ff:ff:ff
    inet 192.168.122.1/24 scope global virbr0
       valid_lft forever preferred_lft forever
    inet6 fe80::5054:ff:fe50:21e8/64 scope link
       valid_lft forever preferred_lft forever
4: virbr0-nic: <BROADCAST,MULTICAST> mtu 1500 qdisc pfifo_fast master
virbr0 state DOWN group default qlen 1000
    link/ether 52:54:00:50:21:e8 brd ff:ff:ff:ff:ff:ff
21: vnet0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast
master virbr0 state UNKNOWN group default qlen 1000
[...]

PR #586, opened 2017-01-18
libvirt: Add different options for VM IP detection

Nadrieril:
This PR adds different methods for detecting the VM's IP address, in case DHCP is not used for example.
It also removes the hacky way to get static MACs since we can get the devices' MAC from libvirt. I understand this could cause libvirt's DHCP to more often assign different IPs to a given machine across deployments. I however believe that people needing a static IP should not have to rely on a DHCP anyways (I'm working on enabling that too).

### DIFF
--- a/nix/libvirtd.nix
+++ b/nix/libvirtd.nix
@@ -95,6 +95,17 @@ in
       description = "Names of libvirt networks to attach the VM to.";
     };
 
+    deployment.libvirtd.privateIPv4 = mkOption {
+      default = "dhcp";
+      example = "10.1.2.3";
+      type = types.str;
+      description = ''
+        IP to use to ssh into the machine.
+        Put 'dhcp' to let nixops get the IP from libvirt's dhcp (works with default libvirt network);
+        put 'arp' to let nixops detect IP in the host's ARP table (works with most setups).
+      '';
+    };
+
     deployment.libvirtd.extraDevicesXML = mkOption {
       default = "";
       type = types.str;
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -7,6 +7,7 @@ import random
 import string
 import subprocess
 import time
+from xml.etree import ElementTree
 
 from nixops.backends import MachineDefinition, MachineState
 import nixops.util
@@ -42,8 +43,6 @@ class LibvirtdState(MachineState):
     private_ipv4 = nixops.util.attr_property("privateIpv4", None)
     client_public_key = nixops.util.attr_property("libvirtd.clientPublicKey", None)
     client_private_key = nixops.util.attr_property("libvirtd.clientPrivateKey", None)
-    primary_net = nixops.util.attr_property("libvirtd.primaryNet", None)
-    primary_mac = nixops.util.attr_property("libvirtd.primaryMAC", None)
     domain_xml = nixops.util.attr_property("libvirtd.domainXML", None)
     disk_path = nixops.util.attr_property("libvirtd.diskPath", None)
     vcpu = nixops.util.attr_property("libvirtd.vcpu", None)
@@ -66,19 +65,9 @@ class LibvirtdState(MachineState):
     def _vm_id(self):
         return "nixops-{0}-{1}".format(self.depl.uuid, self.name)
 
-    def _generate_primary_mac(self):
-        mac = [0x52, 0x54, 0x00,
-               random.randint(0x00, 0x7f),
-               random.randint(0x00, 0xff),
-               random.randint(0x00, 0xff)]
-        self.primary_mac = ':'.join(map(lambda x: "%02x" % x, mac))
-
     def create(self, defn, check, allow_reboot, allow_recreate):
         assert isinstance(defn, LibvirtdDefinition)
         self.set_common_state(defn)
-        self.primary_net = defn.networks[0]
-        if not self.primary_mac:
-            self._generate_primary_mac()
         self.domain_xml = self._make_domain_xml(defn)
 
         if not self.client_public_key:
@@ -114,16 +103,9 @@ class LibvirtdState(MachineState):
         qemu = spawn.find_executable(qemu_executable)
         assert qemu is not None, "{} executable not found. Please install QEMU first.".format(qemu_executable)
 
-        def maybe_mac(n):
-            if n == self.primary_net:
-                return '<mac address="' + self.primary_mac + '" />'
-            else:
-                return ""
-
         def iface(n):
             return "\n".join([
                 '    <interface type="network">',
-                maybe_mac(n),
                 '      <source network="{0}"/>',
                 '    </interface>',
             ]).format(n)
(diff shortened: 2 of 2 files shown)


nix/libvirtd.nix
```diff
@@ -95,6 +95,12 @@ in
       description = "Names of libvirt networks to attach the VM to.";
     };
 
+    deployment.libvirtd.privateIPv4 = mkOption {
+      default = "dhcp";
+      type = types.str;
+      description = "IP to use to ssh into the machine. Put 'dhcp' to let nixops get the IP from libvirt's dhcp (works with default libvirt network); put 'arp' to let nixops detect IP in the host's ARP table (works with most setups)";
```
domenkozar: Split into multiple lines using

```
description = ''
'';
```

Nadrieril:
@domenkozar I have fixed what you suggested.
Are you waiting for some one else's approval or is this mergeable ?

rlupton20:
Suggest this might help with issue #500 (or at least similar issues).

For me, I'm not using libvirtd for dhcp, so queries with virsh fail, but the arp possibilities here would work.

Of course, from a completely self-interested point of view, I'd like to see this merged 😄

rlupton20:
Using libvirt, openvswitch and dhcp (dnsmasq) on the router, with the arp setting, trying to create a new vm still gets stuck looking for an arp. However, I can clear it by pinging it's IP from the host.

I'm guessing that's because openvswitch is dealing with all the networking to the VM, so the host never gets the chance to add the IP to it's ARP tables.

 Ugh, there should be a nicer way to deal with all of this...

rlupton20:
Rough not-totally-thought-through-idea - manually send out an arp (if you have the MAC address) from the host.
If we parse out the arp tables, and send out an arp request if the mac address isn't in the table, might that subsume all these case by case considerations?

Nadrieril:
Yeah, for arp to work, the VM and the host need to have exchanged at least one packet, which does not always happen. That's why I tend to use only static IPs

rlupton20:
I'm just looking into inverse arps and similar schemes (but it usually needs special permissions, presumably because we're grabbing layer 2 / raw sockets).

Ping sweeps would be another possibility:
- `fping` seems quite good but needs root permissions.
- `nmap -sP subnet` also would do, and doesn't need root permissions. Old 'n' gold. Running as root with `-n` even displays the MAC address for us.

libvirtd and various vswitches must be a fairly common use case. ping sweeping a subnet seems like a sledgehammer approach to discovering an IP but at the moment I haven't found a better approach to doing this.

My idea at the moment would be, on merge of the arp functionality you've built, to add another option to `libvirtd` to aid IP discovery, by periodically ping sweeping a specified subnet. This should make using arps usable in other cases, but requires specifying a subnet.

All this seems a bit ugly though, and it would be nice to manage this all through libvirtd. Better suggestions?

Nadrieril:
I kind of like this idea ^^ Wouldn't it be enough do ping sweep only just before trying to get the IP from the arp table ?
I'm not sure we can do much better with libvirt unfortunately. See e.g. https://libvirt.org/formatnwfilter.html#nwfelemsRulesAdvIPAddrDetection , where libvirt has to listen to passing packets to determine the IP assigned to the VM's interface.

rlupton20:
Yeah, that would be more or less how I'd implement it. The only issue I can see is if we don't get an IP in time for the ping sweep to find it. I guess if its an option, we can do it on every test of the arp table, because you'd likely only enable this if you know your virtual network isn't going to be witnessed directly by the host.

rlupton20:
Perhaps that's what you meant, actually.

Nadrieril:
Pretty much, yes :)

rlupton20:
I guess we wait for the merge. @domenkozar ?

nixops/backends/libvirtd.py
```diff
@@ -161,34 +147,64 @@ def iface(n):
...
+                    # TODO: parse command output with a proper regexp, and/or use "net-dhcp-leases --network <network> [--mac <mac>]"
+                    cmd = [ "virsh", "-c", "qemu:///system", "net-dhcp-leases", "--network", net ]
+                    lines = subprocess.check_output(cmd).split()
+                    try:
+                        i = lines.index(mac)
+                    except ValueError:
+                        continue
+                    else:
+                        ip_with_subnet = lines[i + 2]
+                        ip = ip_with_subnet.split('/')[0]
```
NotBrianZach: not really a review;
but I wanted to be able to quickly start and restart vms; and since the vms were using dhcp lookups to query services running on the other vms, I ended up adding code like this here in my fork (to avoid having to wait for leases to expire (I think this was (at least part of) the problem (also occurred with your arp method, and was fixed by similar code))
```
                        self._logged_exec(
                            ["virsh", "-c", "qemu:///system",
                             "net-update", net, "add",
                             "ip-dhcp-host",
                             "<host mac='{0}' name='{1}' ip='{2}' />".format(
                                 mac, self.name, ip),
                             "--live", "--config"
                             ])
```

along with something like this in nixops destroy (which is stop in libvirtd.py)
```
           map(lambda (net, mac):
                self._logged_exec(
                    ["virsh", "-c", "qemu:///system",
                     "net-update", net, "delete",
                     "ip-dhcp-host",
                     "<host mac='{0}' name='{1}' ip='{2}' />".format(
                         mac, self.name, self.private_ipv4),
                     "--live", "--config"
                     ]),
                nets
                )
```
all this to say, is a similar issue why you are interested in static ips/ would this change make sense or is there a hidden drawback I don't see?

NotBrianZach: accidentally opened pull request with referenced changes here https://github.com/NixOS/nixops/pull/610/files :P

Nadrieril: My PR adds different methods for nixops to get the VM IP, but does nothing to give the VM a particular IP.
Your change would totally make sense in another PR that would aim at giving the VM a given static IP. It is rather complementary to what I'm doing here, and would be a logical next step.

Nadrieril: Wait no, I had not looked at the context of your code. What you are doing is rather weird since it tells libvirt to give to the VM the IP it already has. Moreover, it has unreasonable assumptions like the fact that the given network has a dhcp server, which is exactly what I'm trying to get rid of

NotBrianZach: I don't disagree it's weird (leads to pollution of network xml if you don't shut them down properly), problem I had that this fixes is quickly restarting setups like this :
infrastructure-libvirtd.nix
```
let
  defaultConfig =
    {
      deployment.targetEnv = "libvirtd";
      deployment.libvirtd.privateIPv4 = "dhcp";
    };
in
{
  apiServer = defaultConfig;
  postgresqlServer = defaultConfig;
}
```
network.nix
```
  apiServer =
    { config, pkgs, ... }:
      systemd.services.postgrest = {
        description = "Wraps postgresql database with REST api";
        # Start the service after the network is available
        after = [ "network.target" ];
        wantedBy = [ "multi-user.target" ];
        serviceConfig = {
          # The actual command to run
          ExecStart = "${postgrest}/bin/postgrest postgres://root@postgresqlServer:5432/postgres \\
               --port 3030 \\
               --schema public \\
               --anonymous root \\
               --pool 200";
            }; 
  postgresqlServer =
    { config, pkgs ? import <nixpkgs> {}, ... }:
    { ...etc. etc. }
```

Nadrieril: A more proper way to do what you want would be to add a "static_ip" option to the libvirt conf, and add/remove the `ip-dhcp-host` config when the VM is created/destroyed. But doing it from the detected IP is wrong.

NotBrianZach: ...okay I see what you're saying, thanks, I apologize for newbness.

Nadrieril: Don't, your idea was reasonable. Sorry if I sounded rough. I would definitely be interested in a PR implementing that once this PR is merged.

rlupton20:
I'd be keen to see this merged before doing this (if it's actually a good idea), but might we pull out IP detection as a separate abstract class, which has instances for each (broad) method of obtaining an IP? I can see these nested ifs getting a little out of hand the more configuration options are added here.

Is there an argument for doing this more broadly? I'm guessing other hosting systems provide a more reliable way to obtain your IP.

As mentioned, would be keen to see this merged first, so I can do a little work on it, and use it without having to pull from your fork!

nixops/backends/libvirtd.py
NotBrianZach: hmm looked into this some more and it doesn't seem like there's an easy way to do this through libvirt api without upstream change to said api. 
https://www.redhat.com/archives/libvir-list/2016-June/msg01564.html
setting static ip address for guests only works on type=ethernet interfaces.

NotBrianZach: also just getting rid of --config (and keeping --live flag) removes the need to to the update/delete in nixops destroy (stop).

NotBrianZach: https://github.com/NixOS/nixops/pull/613

Nadrieril:
I'm closing this because I'm not using nixops anymore. If anyone still wants this, you can fetch the branch from my fork and make a new PR yourself.

PR #587, opened 2017-01-19
libvirt: Allow control over boot image configuration

### LINKED ISSUE #499: Libvirtd: set root key without overriding root's key
If the NixOS configuration overrides `users.users.root.openssh.authorizedKeys.keys`, NixOps won't be able to ssh into Libvirtd machine anymore since the key will get replaced.

Nadrieril:
This PR adds a deployment.libvirtd.boot_config option that is used to specify the NixOS configuration to use for the first boot.
This is useful in non-standard configurations, for example in networks without DHCP, in case of needing external filesystems, or if special devices are enabled.
It ensures sharing of the base image to make build times reasonable.

### DIFF
--- a/nix/libvirtd-image.nix
+++ b/nix/libvirtd-image.nix
@@ -1,82 +1,132 @@
-{ system ? builtins.currentSystem, size ? "10", config }:
-let
-  pkgs = import <nixpkgs> {};
-  cfg = (import <nixpkgs/nixos/lib/eval-config.nix> {
-    inherit system;
-    modules = [ config ];
-  }).config;
+rec {
+  base_image_config = {
+    fileSystems."/".device = "/dev/disk/by-label/nixos";
 
-in pkgs.vmTools.runInLinuxVM (
-  # TODO: Use <nixpkgs/nixos/lib/make-disk-image.nix> when
-  # https://github.com/NixOS/nixpkgs/issues/20471 is fixed
-  pkgs.runCommand "libvirtd-image"
-    { memSize = 768;
-      preVM =
+    boot.loader.grub.version = 2;
+    boot.loader.grub.device = "/dev/sda";
+    boot.loader.timeout = 0;
+  };
+
+
+  create_nixos_image = {
+    system ? builtins.currentSystem,
+    pkgs ? import <nixpkgs> {},
+    size ? "10",
+    config ? base_image_config
+  }:
+  let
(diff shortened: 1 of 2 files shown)


rbvermaa:
Could you remove the 499 fix from this PR?

Nadrieril:
It doesn't work without it, so I would rather wait for #596 to be merged before removing it (by then rebasing on top of master).

domenkozar:
I don't think we should inline make-disk-image here. The last thing we want is to fragment that code again, after investing so much time to generalize it.

Nadrieril:
Well, it was already pretty much inlined already. Should I replace it with make-disk-image directly ? I thought there were some issues with it like https://github.com/NixOS/nixpkgs/issues/20471.
On the other hand I do need to inline bits of it for the edit_image function, because I don't think there exists any equivalent function in nixpkgs yet.

copumpkin:
Also somewhat relevant (especially if you duplicate the image building code): https://github.com/NixOS/nixpkgs/pull/21943

Nadrieril:
Yes, I'm following your PRs closely :)

3noch:
#596 was merged!

Nadrieril:
I'm closing this because I'm not using nixops anymore. If anyone still wants this, you can fetch the branch from my fork and make a new PR yourself.

3noch:
@Nadrieril Would you mind leaving it open? Regardless of whether or not it is useful to you, it is likely still useful to the rest of us.

domenkozar:
+1, also if you can let us know why you're not using nixops, it might be useful feedback :)

Nadrieril:
Ok.
I'm not using nixops anymore for two reasons: I needed something like `nixos-rebuild test` and after trying to implement it in nixops I gave up; also I wanted to have a build host separate from the targethost.
In the end it turned out that `nixos-rebuild` did almost everything I needed. I'm only missing libvirt management but I'm working on a nixos libvirt module that would do that.

Nadrieril:
Oh, and also I wanted remote deployment but the PR to add remote libvirt deployment seems stalled. And statelessness as much as possible, to make it easier to deploy from multiple machines.
Sorry if I sound like I'm complaining, I think the general idea is that nixops was not the tool I needed.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #596, opened 2017-01-28
Fix #499

### LINKED ISSUE #499: Libvirtd: set root key without overriding root's key

Nadrieril:
This PR adds the auto-generated ssh public key to the machine configuration via MachineState.get_physical_spec().
This fixes #499.

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -63,6 +63,9 @@ class LibvirtdState(MachineState):
         return super_flags + ["-o", "StrictHostKeyChecking=no",
                               "-i", self.get_ssh_private_key_file()]
 
+    def get_physical_spec(self):
+        return {('users', 'extraUsers', 'root', 'openssh', 'authorizedKeys', 'keys'): [self.client_public_key]}
+
     def _vm_id(self):
         return "nixops-{0}-{1}".format(self.depl.uuid, self.name)
 


rbvermaa:
@Nadrieril How does this fix the issue actually? The issue mentions 'if the NixOS configuration overrides the option'. If it is overridden, how would adding this fix the issue? Wouldn't it be still overriden?

Nadrieril:
No because nixos combines both lists of keys.
Given how nixos options combine, by default if your configuration adds a key, both your key and the nixops key are added to the final combination, so it works.

rbvermaa:
NixOS only combines values with the same priority. If e.g. the option is overriden with a lower prio, this would still not work. I've asked Domen for clarification in the issue.

Nadrieril:
Oh right. Still better than the previous behaviour though, which was that setting root.ssh.keys even at default priority would erase nixops' key.
Would there be a better way to do it through nixos configuration ? I'm not super familiar with the intricacies of config merging.

Nadrieril:
This PR may not fix all possible problems, but it still fixes the most current case which is people setting `users.extraUsers.root.openssh.authorizedKeys.keys = [ my_key ]`.
I therefore think we could merge it and think about a better solution only if people actually need it.

danbst:
@rbvermaa please, merge this. A very nasty issue, because extra keys for root currently break `libvirtd` deployments.

I'm now using this patch locally

PR #598, opened 2017-01-31
libvirt: Fix #584

### LINKED ISSUE #584: Libvirt images are world-writable
See https://github.com/NixOS/nixops/blob/master/nixops/backends/libvirtd.py#L104
This is a really problematic security issue, the libvirt backend should not be used as it is.

Nadrieril:
This PR makes the libvirt images readable only by their owner/group (i.e. the user/group under which is run qemu, by default root/root on NixOS afaik). It fixes a critical vulnerability in the libvirt backend.
The user does not actually need read/write permissions on that file for nixops to run correctly. If they want access to that file for other reasons, they should configure their libvirt installation to manage permissions accordingly (see e.g. https://libvirt.org/drvqemu.html#securitydac).

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -100,8 +100,8 @@ class LibvirtdState(MachineState):
             self.disk_path = self._disk_path(defn)
             self._logged_exec(["qemu-img", "create", "-f", "qcow2", "-b",
                                base_image + "/disk.qcow2", self.disk_path])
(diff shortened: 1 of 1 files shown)


bjornfor:
I'd say, please write a commit message that explains what this does (the PR cover message is good -- but that's not what enters the git repo) and someone with commit access please merge. (I too wondered why the image had perms 666 today...)

Nadrieril:
Done, I think this can be merged now.

rbvermaa:
Commited in 1266cdcba

Issue #575, opened 2016-12-20
Support Vultr


3noch:
[Vultr](https://www.vultr.com/) is similar to DigitalOcean but it also supports image uploads.

ravloony:
I was able to deploy to Vultr with no issues.

3noch:
@ravloony With nixops? How did you do it?

ravloony:
This would be great. I was able to install NixOS on a Vultr instance.

3noch:
@ravloony Ah great! Yes Vultr would be an improvement even over DigitalOcean from NixOps POV.

ravloony:
@3noch You're fast1 :-D I brain farted there. I was not able to deploy from NixOps, but only install from ISO.

3noch:
@ravloony Heh...just happened to be on GH at the right time. 

I don't know enough about NixOS internals but I'll bet Vultr would be one of the easier backends to implement.

wmertens:
FYI they just added the NixOS 18.03 minimal ISO to the public library. So not quite NixOps support, but at least it's now feasible to make script that creates a NixOS VM from scratch without an existing snapshot.

lihop:
Neat. However I can't find a way to create a server using an ISO from the ISO Library (I started [this discussion](https://discuss.vultr.com/discussion/1741/create-server-via-api-using-iso-from-iso-library#latest) about it on the Vultr forums).

While looking through the API docs I did notice the [/v1/iso/create_from_url](https://www.vultr.com/api/#iso_create_from_url) endpoint which allows you to upload a custom ISO to your account from a URL.

Unfortunately the public library ISO doesn't let you add an SSH key or run startup script, and because it's based on the official NixOS Minimal Installation CD, ssh is disabled by default, so the only way to access the server is through the web based KVM console.

wmertens:
https://www.vultr.com/api/#server_create says you can pass an ISOID. You can also set the userdata with https://www.vultr.com/api/#server_set_user_data, which I would expect can be parsed by cloud-init

And then if https://github.com/NixOS/nixpkgs/issues/39076 gets traction, fully automated NixOS deployments on Vultr and others would be easy.

wmertens:
ah, but the problem is finding the isoid for the public library :-/

wmertens:
Looking at the deploy html, it turns out that the NixOS 18.03 ISO has the id `410303` :)

wmertens:
[...]

PR #634, opened 2017-03-23
[WIP] Add vultr backend

disassembler:
addresses issue #575

### DIFF
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -1494,6 +1494,45 @@ xlink:href="https://github.com/elitak/nixos-infect">nixos-infect</link>
 <literal>NIXOS_LUSTRATE</literal> under the hood.
 </para>
 </section>
+<section xml:id="sec-deploying-to-vultr"><title>Deploying to Vultr</title>
+
+<para><xref linkend="ex-trivial-vultr.nix" /> shows how to run
+a vultr instance. We only support instance creation and
+destruction at the moment.
+</para>
+
+<para>Note that an image needs created in vultr using an iso with your default ssh
+  key for deployment. Follow these
+<link xlink:href="https://www.vultr.com/docs/install-nixos-on-vultr">directions</link>
+to setup an image using nixos. Make sure to assign a public <literal>ssh-key</literal>
+to root in configuration.nix and allow root login with key. This key is only used
+for initial connection and a new key stored in the nixops state will be used for future connections. After booting the instance, create a snapshot in the vultr UI.
+</para>
+
+<example xml:id="ex-trivial-vultr.nix">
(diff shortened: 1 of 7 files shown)


domenkozar:
Can you describe steps how to test the backend?

disassembler:
I'll try to add documentation and tests this weekend. So far it's working great! Nixops is so much better than ansible/terraform for this use case.

disassembler:
documentation and trivial example added. Still needs tests.

3noch:
So glad to see this getting some attention! Yay!

dhess:
Hi, I pulled these 2 commits into my local tree, which is caught up with `master`. Installation works fine, but as soon as I reboot the instance, it fails to mount the root filesystem.

```
...
running udev...
starting device mapper and LVM...
Scanning for Btrfs filesystems
mounting /dev/vda1 on /...
mount: mounting /dev/vda1 on /mnt-root/ failed: Invalid argument

An error occurred in stage 1 of the boot process, which must mount the
root filesystem on `/mnt-root` and then start stage 2.  Press one
of the following keys:

   r) to reboot immediately
   *) to ignore the error and continue
```

Are you seeing this at all? I followed the steps outlined in https://www.vultr.com/docs/install-nixos-on-vultr, the only difference being that I used `ext4` rather than `btrfs` on the root filesystem.

disassembler:
Yeah, btrfs is currently hard-coded in the backend python file (similar to ext4 being hard-coded in the digital ocean one). I need to mess around with it and see what would need done to move that into the config object instead. I'm looking at the libvirtd implementation of nixops to see how I could do that, but not making much progress.

dhess:
Ahh, ok. Thanks for the clarification, I’ll try to work around this.

wmertens:
Instead of the custom snapshot, wouldn't it be easier to use the smallest image they have and run https://github.com/jeaye/nixos-in-place on it? That would make using Vultr as simple as providing your credentials…

pierrebeaucamp:
I agree that the best UX would be to only provide the credentials and let nixops worry about the rest. I wonder how hard it would be for nixops to create a snapchat on its own. But I would be more than happy to have a first version of vultr support in nixops, even if it would still require the manual steps.

wmertens:
@pierrebeaucamp you're right, it shouldn't block inclusion. Once you have a snapshot you can basically use it forever.

I would like to see https://www.vultr.com/api/ mentioned on where to get the `dcid` and `vpsplanid` values. (funny, I just noticed you can't get the $2.5 sandbox plan via the API). Since you can only really get those values via the API, perhaps there could be a command to retrieve those values? I don't know if nixops allows custom commands like that… Perhaps get those values if they are missing from the description?

grahamc:
Please do this in a plugin.

PR #638, opened 2017-03-31
Always print the stacktrace if unforeseen exception in deploy

Profpatsch:
“This thread shouldn’t throw an exception”, but it sometimes does, and when it
does it should print the whole stacktrace for debugging, not just a cute
message, since `debug` is an alias of `False` (line 38).


domenkozar:
I think `debug` is set globally from the cli options. But I agree, it's important to see what happened.

@edolstra @rbvermaa thoughts?

Issue #465, opened 2016-07-08
GCE machines don't recover after instance type change/reboot


MostAwesomeDude:
I changed the instance type of a GCE machine. nixops stops the machine, obstensibly to change the instance type, but then the machine is not started up again. Also the instance type wasn't changed. I can manually change the instance type, but now nixops thinks that my machine is down and won't bring it back up. (I can, and have, manually started the machine from the GCE console.)

```
$ nixops deploy --allow-reboot --include zulu
zulu......> warning: change of the instance type requires a reboot
No handlers could be found for logger "libcloud.common.google"
zulu......> stopping GCE machine... stopped
building all machine configurations...
these derivations will be built:
  /nix/store/4axpnynh2dcyq10nikil49cwsi29cqq6-system-units.drv
  /nix/store/4837mz5p6ybzk3nh4axdar1g268597kc-etc.drv
  /nix/store/vwz668iqw2shwggdpk49lw0xpvxl5c26-nixos-system-zulu-16.09pre85639.453086a.drv
  /nix/store/jkh2yyyjigzsldbjd51lh02jgp34k4c9-nixops-machines.drv
building path(s) ‘/nix/store/1cyla5cnw3kj6g0qz8gxjgani5i9hzgv-system-units’
building path(s) ‘/nix/store/1d9kq12i74h7fn6p37h6fxx0x8ml0d0x-etc’
building path(s) ‘/nix/store/hqhrh536k45iz7cz3yvr86vf1zr8wl7x-nixos-system-zulu-16.09pre85639.453086a’
building path(s) ‘/nix/store/ilqpjmk5gpvhaqq83nnx810r8avk8qyl-nixops-machines’
zulu......> copying closure...
ssh: connect to host 104.196.121.223 port 22: Connection timed out
zulu......> could not connect to ‘root@104.196.121.223’, retrying in 1 seconds...
```

Here's what nixops thinks of my deployment:

```
$ nixops check
No handlers could be found for logger "libcloud.common.google"
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+-------+
| Name    | Exists | Up  | Reachable | Disks OK | Load avg.      | Units                                  | Notes |
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+-------+
| alpha   | Yes    | Yes | Yes       | Yes      | 0.00 0.01 0.05 | proc-sys-fs-binfmt_misc.mount [failed] |       |
[...]

PR #642, opened 2017-04-07
Adding the ability to pass/modify a service account in a GCE instance

AmineChikhaoui:
This allows setting a service account option in a GCE machine definition and changing it as well if the definition is modified by either changing the service account or the scopes (require a reboot).

Example:
```
gce = { pkgs, resources, lib, ...}:  {
      environment.systemPackages = [ pkgs.google-cloud-sdk ];
      deployment.targetEnv = "gce";
      deployment.gce = credentials // {
        region = "europe-west1-b";
        instanceType = "n1-standard-4";
        network = resources.gceNetworks.lb-net;
        instanceServiceAccount = { 
          email = "gce-test-1@...iam.gserviceaccount.com";
          scopes = [ "https://www.googleapis.com/auth/monitoring.write" "https://www.googleapis.com/auth/devstorage.read_only" ];
        };  
      };  
    };
```

This also fix the weird behavior of #465 by avoiding the deployment failure but doesn't fix the original issue which is the update of the instance type (will take care of it in a different PR).

### DIFF
--- a/nix/gce.nix
+++ b/nix/gce.nix
@@ -264,6 +264,32 @@ in
         '';
       };
 
+      instanceServiceAccount = mkOption {
+        default  = {};
+        type = (types.submodule {
+          options = {
+            email = mkOption {
+              default = "default";
+              type = types.str;
+              description = ''
+                Email address of the service account.
+                If not given, Google Compute Engine default service account is used.
+              '';
+            };
+            scopes = mkOption {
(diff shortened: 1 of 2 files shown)


AmineChikhaoui:
@Phreedom Can you have a look at this ? Thanks in advance.
cc @rbvermaa

PR #644, opened 2017-04-08
GCE backend: Add preemptible option for instances

ljli:
I'm using this since a few days ago, so far no problems.


domenkozar:
@Phreedom @rbvermaa

Issue #646, opened 2017-04-10
Add .path property to keys


nh2:
Keys are put into `/run/keys` by convention; it would be great if the final path of a key would be exposed in a `.path` property that I can reference in my configuration in order to protect me from typos.

domenkozar:
Implemented in https://github.com/NixOS/nixops/pull/664

PR #650, opened 2017-04-16
Stop storing keys in the nix store for storeKeysOnMachine.

kevincox:
This patch changes storeKeysOnMachine to work roughly the same way when
true as when false. The only difference is that the keys are stored in
/var/keys (which is usually a real disk) as opposed to /run/keys (which
is RAM).

This is an improvement on the previous version of storeKeysOnMachine
where the keys were stored in the Nix store and linked from /run/keys.
The previous version doesn't allow setting unix permissions on the keys,
meaning any process on the server can read all keys.

This solution has the downside that rolling back doesn't roll back the
keys, however this is consistent with how storeKeysOnMachine=false works
so shouldn't be a major concern. Furthermore if someone wants to have
the keys rollback with the system they can use builtins.toFile instead
of the keys mechanism which works the same way that the previous
storeKeysOnMachine=true worked.

Additionally a symlink /etc/nixos-keys has been added that will be set
to the appropriate key location to make it easier to switch between the
options.


kevincox:
cc: @aszlig @ryanartecona @edolstra as you guys have touched this code last.

domenkozar:
So really `storeKeysOnMachine` should be called `persistKeys`?

kevincox:
Yeah, it probably could have a better name. But that is orthogonal to this PR.

kevincox:
On Apr 25, 2017 19:28, "aszlig" <notifications@github.com> wrote:

First of all, having /etc/nixops-keys sounds a bit dangerous to me, given
that files in /etc are usually generated and might be overwritten if there
is an environment.etc."nixops-keys" option definition. So I'd opt to use
something in /var/lib instead.

How about setting up the symlink using Nix. That sounds like the best of
both worlds.

But that aside, having different key directories based on whether
storeKeysOnMachine is true or false is also not necessary here, because we
can simply bind-mount the keys from let's say /var/lib/nixops-keys to
/run/keys - even on a one by one basis.

Call me whatever you will but I like the simplicity of a different
directory over bind mounting a dink directory over a ramfs. However if
people prefer a bind mount I'll galdly change it.

aszlig:
@kevincox: Instead of bind mounts, I guess symlinks are okay as well, since that's what's already done in the existing implementation, although a bind mount would have the advantage that it will work even if programs don't properly dereference symlinks or doing permission checks on the symlink.

Having it in a different directory now means that existing deployments need to check `storeKeysOnMachine` in order to figure out the right directory, so even if we go that road it might be a good idea to make this available via a module option or even argument.

Another problem I see here (which is problematic for `storeKeysOnMachine` as well, but it's a `ramfs` so keys don't persist) is that old keys are not removed.

kevincox:
> Having it in a different directory now means that existing deployments need to check storeKeysOnMachine in order to figure out the right directory, so even if we go that road it might be a good idea to make this available via a module option or even argument.
I didn't really want to worry about linking from a ramfs but since I'm chmoding them anyways I may as well.

Updated.

> Another problem I see here (which is problematic for storeKeysOnMachine as well, but it's a ramfs so keys don't persist) is that old keys are not removed.
I didn't really want to remove the keys on deploy because the new keys are uploaded before the new config is activated. I could probably add it to the chmod activation script but since removing keys is pretty rare I propose pushing this to the next PR.

ip1981:
Guys, how about adding an option for keys path? Defaulting to `/run/keys`? So instead of guessing on `storeKeysOnMachine`, one can choose path where to put keys?

I used to have services for persisting keys: https://gist.github.com/ip1981/755f7bf71c157ebb1bd8f6d404cae5d3

And I'm really ready for configurable keys path: https://github.com/ip1981/nixsap/blob/49659d55878d09827566902d07d01f051926aff2/modules/deployment/keyrings.nix#L38 :)

domenkozar:
We clearly need to document motivations for these options. We are mixing different needs here.

The way I see it:

1) Storing keys in nix store is useful for rollbacks. Since that exposes the keys, this is not the default.

2) Storing keys in `/run/keys/*` location could be configurable together with fixing https://github.com/NixOS/nixops/issues/646

domenkozar:
A different take on this: https://github.com/NixOS/nixops/pull/664

kevincox:
> Guys, how about adding an option for keys path? Defaulting to /run/keys. So instead of guessing on storeKeysOnMachine, one can choose path where to put keys.

This is what the first version did basically. It was binary but changing the target would be trivial. It also added a `/etc/nixops-keys` symlink to abstract where it was stored from users.

> Storing keys in /nix/store/ is useful for rollbacks. Since that exposes the keys as world readable, this is not the default.

If you want to do this simply use builtins.toFile. I don't see any advantage to using the keys system.

> Storing keys in /run/keys/* location could be configurable together with fixing #646

I'm okay with that.

ip1981:
Honestly, I think rollbacks are overestimated. First of all, it means discrepancy between the machine specification and actual state. Second, deploy with previous configuration is in fact a rollback. If you haven't done GC, it will be very fast. Quick rollback in case of f*ck up... should not happen :) Thank to Nix, I can test configurations and migrations to death before pushing to production.

ip1981:
I only see some value in "keys in the store" that if key content changes you will get activation/restart. There is a hopefully nice workaround: use new key name for new content, e .g. by appending a number :)

kevincox:
I think rollbacks aren't commonly useful to have for the key level. I would argue that they are very useful however when you need to get an ssh server running so that you can deploy the new version :)

domenkozar:
https://github.com/NixOS/nixops/pull/664 is ready for review as an alternative implementation. It doesn't change behavior of `storeKeysOnMachine = true`, but it allows to specify `destDir` when `storeKeysOnMachine = false`.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #656, opened 2017-04-30
libvirtd: add address_to method

sargon:
The mechanism generating the extraHosts declaration is calling this
method. So without it, the extraHosts stays basic and
inter machine communication can't defined in a sane way.
Furthermore without this the libvirtd backend dosnt' behave like
manual is stating.


rbvermaa:
Committed in 78c732c

rbvermaa:
Sorry for the delay, libvirtd was broken on NixOS 17.03 for a while, so I wasn't able to test until now.

PR #658, opened 2017-05-06
Add `--no-build-output` option

ip1981:
Useful in unattended systems for continuous integration / delivery.

Workaround is to create wrappers for `nix-build`. 

P. S. In conjunction with https://github.com/NixOS/nix/issues/443#issuecomment-296752535 :)


domenkozar:
I'd prefer to call `--no-build-output` not to make too much of a mess.

PR #661, opened 2017-05-09
deployment.keys:  'keyFile' option to provide secret contents from local files

deepfire:
This is supposed to address https://github.com/NixOS/nix/issues/1307 by adding an option `filename` (as an alternative to `text`) to `deployments.keys` to avoid ever storing a secret in Nix string.

Note, that this doesn't add the corresponding part to `system.activationScripts.nixops-keys`.

### DIFF
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -6,13 +6,37 @@ let
   keyOptionsType = types.submodule {
     options.text = mkOption {
       example = "super secret stuff";
-      type = types.str;
+      default = null;
+      type = types.nullOr types.str;
       description = ''
-        The text the key should contain. So if the key name is
-        <replaceable>password</replaceable> and <literal>foobar</literal>
-        is set here, the contents of the file
+        When non-null, this designates the text that the key should
+        contain. So if the key name is <replaceable>password</replaceable>
+        and <literal>foobar</literal> is set here, the contents of the file
         <filename>/run/keys/<replaceable>password</replaceable></filename>
         will be <literal>foobar</literal>.
+
+        NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
+        to be set.
+      '';
+    };
+
+    options.keyFile = mkOption {
+      default = null;
+      type = types.nullOr types.path;
+      description = ''
+        When non-null, contents of the specified file will be deployed to
+        the specified key on the target machine.  If the key name is
+        <replaceable>password</replaceable> and <literal>/foo/bar</literal> is set
+        here, the contents of the file
+        <filename>/run/keys/<replaceable>password</replaceable></filename>
+        deployed will be the same as local file <literal>/foo/bar</literal>.
+
+        Since no serialization/deserialization of key contents is involved, there
+        are no limits on that content: null bytes, invalid Unicode,
+        <literal>/dev/random</literal> output -- anything goes.
+
+        NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
+        to be set.
       '';
     };
 
@@ -123,32 +147,41 @@ in
       "things to break."
     )];
 
+    assertions = flip mapAttrsToList config.deployment.keys (key: opts: {
+      assertion = (opts.text == null && opts.keyFile != null) ||
+                  (opts.text != null && opts.keyFile == null);
+      message = "Deployment key '${key}' must have either a 'text' or a 'keyFile' specified.";
+    });
+
     system.activationScripts.nixops-keys = stringAfter [ "users" "groups" ]
       ''
         mkdir -p /run/keys -m 0750
         chown root:keys /run/keys
 
         ${optionalString config.deployment.storeKeysOnMachine
-            (concatStrings (mapAttrsToList (name: value:
-              let
-                # FIXME: The key file should be marked as private once
-                # https://github.com/NixOS/nix/issues/8 is fixed.
-                keyFile = pkgs.writeText name value.text;
-              in "ln -sfn ${keyFile} /run/keys/${name}\n")
-              config.deployment.keys)
+            (concatStrings (mapAttrsToList
+                            (name: value: let
+                                            # FIXME: The key file should be marked as private once
+                                            # https://github.com/NixOS/nix/issues/8 is fixed.
+                                            keyFile = if !isNull value.keyFile
+                                                      then value.keyFile
+                                                      else pkgs.writeText name value.text;
+                                          in "ln -sfn ${keyFile} /run/keys/${name}\n")
+                           config.deployment.keys)
             + ''
               # FIXME: delete obsolete keys?
               touch /run/keys/done
             '')
         }
 
-        ${concatStringsSep "\n" (flip mapAttrsToList config.deployment.keys (name: value:
-          # Make sure each key has correct ownership, since the configured owning
-          # user or group may not have existed when first uploaded.
-          ''
-            [[ -f "/run/keys/${name}" ]] && chown '${value.user}:${value.group}' "/run/keys/${name}"
-          ''
-        ))}
+        ${optionalString (!config.deployment.storeKeysOnMachine)
+          (concatStringsSep "\n" (flip mapAttrsToList config.deployment.keys (name: value:
+            # Make sure each key has correct ownership, since the configured owning
+            # user or group may not have existed when first uploaded.
+            ''
+              [[ -f "/run/keys/${name}" ]] && chown '${value.user}:${value.group}' "/run/keys/${name}"
+            ''
+        )))}
       '';
 
     systemd.services = (
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -22,8 +22,12 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
 
         def _extract_key_options(x):
             opts = {}
-            for key in ('text', 'user', 'group', 'permissions'):
-                elem = x.find("attrs/attr[@name='{0}']/string".format(key))
+            for (key, xmlType) in (('text',        'string'),
+                                   ('keyFile',     'path'),
+                                   ('user',        'string'),
+                                   ('group',       'string'),
+                                   ('permissions', 'string')):
+                elem = x.find("attrs/attr[@name='{0}']/{1}".format(key, xmlType))
                 if elem is not None:
                     opts[key] = elem.get("value")
             return opts
@@ -212,7 +216,15 @@ class MachineState(nixops.resources.ResourceState):
         for k, opts in self.get_keys().items():
             self.log("uploading key ‘{0}’...".format(k))
             tmp = self.depl.tempdir + "/key-" + self.name
-            f = open(tmp, "w+"); f.write(opts['text']); f.close()
+
+            if 'text' in opts: 
+                with open(tmp, "w+") as f:
+                    f.write(opts['text'])
+            elif 'keyFile' in opts:
+                self._logged_exec(["cp", opts['keyFile'], tmp])
+            else:
+                raise Exception("Neither 'text' or 'keyFile' options were set for key '{0}'.".format(k))
+
             outfile = "/run/keys/" + k
             outfile_esc = "'" + outfile.replace("'", r"'\''") + "'"
             self.run_command("rm -f " + outfile_esc)


nixops/backends/__init__.py
```diff
@@ -212,7 +212,15 @@ def send_keys(self):
         for k, opts in self.get_keys().items():
             self.log("uploading key ‘{0}’...".format(k))
             tmp = self.depl.tempdir + "/key-" + self.name
-            f = open(tmp, "w+"); f.write(opts['text']); f.close()
+
+            # Ensure that keys.<key>.file XOR keys.<key>.file
+            self.log ("'text' in opts: %s, 'filename' in opts: %s" % ('text' in opts, 'filename' in opts))
+            assert     ('text' in opts  or 'filename' in opts), "Specification error for key '%s': neither `keys.%s.text` nor `keys.%s.file` was specified."  % (k, k, k)
+            assert not ('text' in opts and 'filename' in opts), "Specification error for key '%s': `keys.%s.text` and `keys.%s.file` are mutually exclusive." % (k, k, k)
```
ip1981: Isn't assert in Nix more explicit?

ip1981:
Almost the same can be achieved with `builtins.readFile`. I mean not in NixOps, but in machines specifications.

domenkozar:
@ip1981 except that with `builtins.readFile` you hit https://github.com/NixOS/nix/issues/1307 as random data can contain null bytes.

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
+      '';
+    };
+
+    options.filename = mkOption {
+      example = "./super-secret.key";
+      default = null;
+      type = types.nullOr types.str;
+      description = ''
+
+        When this is non-null, contents of the specified file will be deployed to
```
domenkozar: s/file/filename/

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
+        specified.
+      '';
+    };
+
+    options.filename = mkOption {
+      example = "./super-secret.key";
+      default = null;
+      type = types.nullOr types.str;
+      description = ''
+
```
domenkozar: (nitpick) Extra unneeded empty line.

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
-        The text the key should contain. So if the key name is
-        <replaceable>password</replaceable> and <literal>foobar</literal>
-        is set here, the contents of the file
+        When this is non-null, this designates the text that the key should
+        contain. So if the key name is <replaceable>password</replaceable>
+        and <literal>foobar</literal> is set here, the contents of the file
         <filename>/run/keys/<replaceable>password</replaceable></filename>
         will be <literal>foobar</literal>.
+
+        NOTE: either this XOR the <literal>filename</literal> option ought to be
```
domenkozar: I'd rephrase this to: `NOTE: Either <literal>text</literal> or <literal>filename</literal> have to be set.`

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
+      description = ''
+
+        When this is non-null, contents of the specified file will be deployed to
+        the specified key on the target machine.  If the key name is
+        <replaceable>password</replaceable> and <literal>/foo/bar</literal> is set
+        here, the contents of the file
+        <filename>/run/keys/<replaceable>password</replaceable></filename> will be
+        the same as that of <literal>/foo/bar</literal> on the deploying host.
+
+        NOTE: either this XOR the <literal>text</literal> option ought to be
```
domenkozar: I'd rephrase this to: NOTE: Either <literal>text</literal> or <literal>filename</literal> have to be set.

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
+      example = "./super-secret.key";
+      default = null;
+      type = types.nullOr types.str;
+      description = ''
+
+        When this is non-null, contents of the specified file will be deployed to
+        the specified key on the target machine.  If the key name is
+        <replaceable>password</replaceable> and <literal>/foo/bar</literal> is set
+        here, the contents of the file
+        <filename>/run/keys/<replaceable>password</replaceable></filename> will be
```
domenkozar: the contents of the file deployed ... will be the same as local file <literal>/foo/bar</literal>

nix/keys.nix
```diff
@@ -133,6 +155,8 @@ in
               let
                 # FIXME: The key file should be marked as private once
                 # https://github.com/NixOS/nix/issues/8 is fixed.
+                #
+                # FIXME: implement value.filename for this as well
```
domenkozar: I suggest we fail here if `text` is not set.

Given the discussion in https://github.com/NixOS/nixops/pull/650, this will change soon so no need to implement `filename` handling sanely.

nixops/backends/__init__.py
```diff
@@ -212,7 +212,14 @@ def send_keys(self):
         for k, opts in self.get_keys().items():
             self.log("uploading key ‘{0}’...".format(k))
             tmp = self.depl.tempdir + "/key-" + self.name
-            f = open(tmp, "w+"); f.write(opts['text']); f.close()
+
+            # Ensure that keys.<key>.text XOR keys.<key>.filename
+            assert     ('text' in opts  or 'filename' in opts), "Specification error for key '%s': neither `keys.%s.text` nor `keys.%s.filename` was specified."  % (k, k, k)
+            assert not ('text' in opts and 'filename' in opts), "Specification error for key '%s': `keys.%s.text` and `keys.%s.filename` are mutually exclusive." % (k, k, k)
```
domenkozar: Let's move this to Nix. For example like in https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/config/users-groups.nix#L539-L564

nixops/backends/__init__.py
```diff
@@ -212,7 +212,14 @@ def send_keys(self):
...
+
+            if   'text'     in opts: f = open(tmp, "w+"); f.write(opts['text']); f.close()
+            elif 'filename' in opts: self._logged_exec(["cp", opts['filename'], tmp])
```
domenkozar: A bit more Pythonic code would be:

```python
    if 'text' in opts: 
        with open(tmp, "w+") as f:
            f.write(opts['text'])
    elif 'filename' in opts:
         self._logged_exec(["cp", opts['filename'], tmp])
    else:
         raise Exception("Neither text of filename were set ...")
```

nix/keys.nix
```diff
@@ -6,13 +6,35 @@ let
...
     options.text = mkOption {
       example = "super secret stuff";
-      type = types.str;
+      default = null;
+      type = types.nullOr types.str;
       description = ''
-        The text the key should contain. So if the key name is
-        <replaceable>password</replaceable> and <literal>foobar</literal>
-        is set here, the contents of the file
+        When this is non-null, this designates the text that the key should
```
domenkozar: `When non-null this designates ...`

ip1981:
On the other side, file-based keys are more feasible (to avoid secrets in nix expressions) and should be implemented this way from the start :)

nix/keys.nix
deepfire: A file name specifies a file, I'd say..

nixops/backends/__init__.py
deepfire: Done, moved to Nix option-specific assertions, as per @domenkozar  suggestion.

nix/keys.nix
deepfire: This will fail (by the virtue of the assertions now moved to Nix) if _neither_ `text` nor `filename` is set is that good enough?

nixops/backends/__init__.py
deepfire: Done.

nix/keys.nix
domenkozar: If only `filename` is set and `deployment.storeKeysOnMachine` is set to true, this will fail since `value.text` one line below will be `null`.

We have to add another assertion to say that `deployment.storeKeysOnMachine && filename` is not supported.

nix/keys.nix
```diff
@@ -6,13 +6,34 @@ let
...
+
+        NOTE: Either <literal>text</literal> or <literal>filename</literal> have
+        to be set.
+      '';
+    };
+
+    options.filename = mkOption {
+      example = "./super-secret.key";
+      default = null;
+      type = types.nullOr types.str;
```
ip1981: I'd make it a path, to be sure nix can find the file (by absolute path) :)
It does not need to be a derivation,

nixops/backends/__init__.py
```diff
@@ -212,7 +212,10 @@ def send_keys(self):
         for k, opts in self.get_keys().items():
             self.log("uploading key ‘{0}’...".format(k))
             tmp = self.depl.tempdir + "/key-" + self.name
-            f = open(tmp, "w+"); f.write(opts['text']); f.close()
+
+            if   'text'     in opts: f = open(tmp, "w+"); f.write(opts['text']); f.close()
+            elif 'filename' in opts: self._logged_exec(["cp", opts['filename'], tmp])
```
ip1981: Absolute path would help for files named like `--help` ;)

nixops/backends/__init__.py
```diff
@@ -212,7 +212,15 @@ def send_keys(self):
...
             tmp = self.depl.tempdir + "/key-" + self.name
-            f = open(tmp, "w+"); f.write(opts['text']); f.close()
+
+            if 'text' in opts: 
+                with open(tmp, "w+") as f:
+                    f.write(opts['text'])
+            elif 'filename' in opts:
+                self._logged_exec(["cp", opts['filename'], tmp])
+            else:
+                raise Exception("Neither 'text' or 'filename' options were set for key '%s'." % k)
```
domenkozar: To support Python 3 use "".format()

nix/keys.nix
deepfire: I have tried to make it a path, but couldn't make the XML query work.

nix/keys.nix
```diff
@@ -6,13 +6,34 @@ let
...
+      description = ''
+        When non-null, contents of the specified file will be deployed to
+        the specified key on the target machine.  If the key name is
+        <replaceable>password</replaceable> and <literal>/foo/bar</literal> is set
+        here, the contents of the file
+        <filename>/run/keys/<replaceable>password</replaceable></filename>
+        deployed will be the same as local file <literal>/foo/bar</literal>.
```
domenkozar: (nitpick): explain how this option mitigates https://github.com/NixOS/nix/issues/1307

nixops/backends/__init__.py
ip1981: Or use Python's [shutils](https://docs.python.org/3/library/shutil.html) or something like that. Absolute paths are still good :)

nix/keys.nix
ip1981: If the problem is with slashes, it should be fixed anyway, because someone might wan to use subdirectories: `./secrets/my.key`

nixops/backends/__init__.py
domenkozar: I don't understand your comment, what absolute path?

ip1981: From here https://github.com/NixOS/nixops/pull/661#discussion_r115726566. Absolute paths (starting with `/`) could help avoid problems with `cp`.

nix/keys.nix
deepfire: Done.

nix/keys.nix
deepfire: No, it's with XML queries for non-`str` types being harder, and unprecedented (according to my, admittedly limited, grep skill) in the nixops repository.  I could get down to the bottom of matter, but it would take more effort.

nix/keys.nix
```diff
@@ -123,6 +144,16 @@ in
       "things to break."
     )];
 
+    assertions = flip mapAttrsToList config.deployment.keys (key: opts: {
+      assertion = (opts.text == null && opts.filename != null) ||
+                  (opts.text != null && opts.filename == null);
+      message = "Deployment key '${key}' must have either a 'text' or a 'filename' specified.";
```
ip1981: Add
```
 || (opts.text == null && opts.filename == null)
```

and corresponding python code can be removed.

nix/keys.nix
deepfire: Done.

nix/keys.nix
```diff
@@ -123,6 +148,16 @@ in
...
+    }) ++        flip mapAttrsToList config.deployment.keys (key: opts: {
+      assertion = !config.deployment.storeKeysOnMachine ||
+                  !opts.filename;
+      message = "On-machine storage of 'filename'-specified deployment keys is not supported. Offending key is: '${key}'.";
```
domenkozar: I'd change the wording to: `deployment.storeKeysOnMachine` and `filename`-specified deployment keys is not ...

ip1981: But it could work if you capture it into Nix store (`"${filename}"`).

ip1981: Not supporting `filename` in some cases is surprising.

nix/keys.nix
deepfire: Alright, done the switch to `path`-typing.

nixops/backends/__init__.py
deepfire: Switched to `pathname` type, so this should be covered, right @ip1981?

nixops/backends/__init__.py
deepfire: Done.

deepfire:
@ip1981, do you think we can move this forward without having the `filename` support the `storeKeysOnMachine` case?

Do you have any other concerns?

domenkozar:
cc @rbvermaa @edolstra

ip1981:
> do you think we can move this forward without having the filename support the `storeKeysOnMachine` case?

I'd take a break, sleep over, and fix that :) Or at least mark that as "to be done".

deepfire:
@domenkozar, I'm preparing a PR to facilitate specifying target location for the keys on top of this PR.

domenkozar:
For supporting `storeKeysOnMachine` we would just do `keyFile = value.filename` and add an if/else to pick between text or filename.

@deepfire if that works, let's get it done :)

deepfire:
I'm thinking of renaming this option to `keys.<key>.keyfile`, because of terminologic ambiguities arising -- what do you guys think, @domenkozar, @ip1981?

domenkozar:
Maybe `source`? Then deployment location would be `target` and we can deprecate `text`.

deepfire:
This would create an ambiguity -- `destination` ought to be a directory, right?  And `source` will be a file.

Type mismatch : -)

deepfire:
Or, perhaps we can make `source` to be a directory too -- and fetch the key from there by its implied name, @domenkozar?

domenkozar:
Good point. Let's then go with `keyFile` and `destination`.

deepfire:
@domenkozar, just to make sure -- should we add the `destination` in this same PR (that's my current understanding)?

domenkozar:
No, let's keep that separate.

ip1981:
> `keys.<key>.keyfile`

I will love `keys.<key>.file`.

deepfire:
@domenkozar, @ip1981, what do we decide about the option's name?

ip1981:
I like shorter `file`, then `source`.

nix/keys.nix
deepfire: Supported, in the latest iteration.

deepfire:
`storeKeysOnMachine = true` with `keyFile` supported in the latest iteration.

nix/keys.nix
```diff
@@ -6,13 +6,38 @@ let
...
         <filename>/run/keys/<replaceable>password</replaceable></filename>
         will be <literal>foobar</literal>.
+
+        NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
+        to be set.
+      '';
+    };
+
+    options.keyFile = mkOption {
+      example = ./super-secret.key;
```
ip1981: Ouch... I'm not sure this evaluates. Maybe drop example? Everyone knows what is a file :)

deepfire: Hmm, oddly it never had given me a problem.  What was the error message?

But ok, removed it.

ip1981: Nothing, really. Just random path in docs, maybe. Never mind.

domenkozar:
I'm going to let this one here until tomorrow and merge.

nix/keys.nix
```diff
@@ -123,32 +147,41 @@ in
...
         }
 
-        ${concatStringsSep "\n" (flip mapAttrsToList config.deployment.keys (name: value:
-          # Make sure each key has correct ownership, since the configured owning
-          # user or group may not have existed when first uploaded.
-          ''
-            [[ -f "/run/keys/${name}" ]] && chown '${value.user}:${value.group}' "/run/keys/${name}"
-          ''
-        ))}
+        ${optionalString (!config.deployment.storeKeysOnMachine)
```
domenkozar: Why did you add this condition? Now  the code doesn't respect `user` and `group` settings anymore if I'm not mistaken.

deepfire: @domenkozar, because calling `chown` on a symlink pointing to read-only Nix store failed, when I was testing it -- I was actually surprised how it worked before.

domenkozar: Opened https://github.com/NixOS/nixops/issues/666

domenkozar:
Besides #666 which was the case before this PR, I consider this done and working.

PR #664, opened 2017-05-11
Add deployment.keys.<K>.destDir and deployment.keys.<K>.path

deepfire:
This allows customisation of the default `/run/keys` location for the deployment keys.

### DIFF
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -42,7 +42,7 @@ let
 
     options.destDir = mkOption {
       default = /run/keys;
-      type = types.nullOr types.path;
+      type = types.path;
       description = ''
         When specified, this allows changing the destDir directory of the key
         file from its default value of <filename>/run/keys</filename>.
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -12,7 +12,7 @@ let
         When non-null, this designates the text that the key should contain. So if
         the key name is <replaceable>password</replaceable> and
         <literal>foobar</literal> is set here, the contents of the file
-        <filename><replaceable>destinationFolder</replaceable>/<replaceable>password</replaceable></filename>
+        <filename><replaceable>destDir</replaceable>/<replaceable>password</replaceable></filename>
         will be <literal>foobar</literal>.
 
         NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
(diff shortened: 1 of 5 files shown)


deepfire:
cc @domenkozar

nix/keys.nix
```diff
@@ -24,11 +24,11 @@ let
...
       type = types.nullOr types.path;
       description = ''
-        When non-null, contents of the specified file will be deployed to
-        the specified key on the target machine.  If the key name is
+        When non-null, contents of the specified file will be deployed to the
+        specified key on the target machine.  If the key name is
         <replaceable>password</replaceable> and <literal>/foo/bar</literal> is set
         here, the contents of the file
-        <filename>/run/keys/<replaceable>password</replaceable></filename>
+        <filename><replaceable>destinationFolder</replaceable>/<replaceable>password</replaceable></filename>
```
ip1981: I'd love `destDir` more :)  The shorter, the better. No "folders" at least. Child trauma.

nixops/backends/__init__.py
```diff
@@ -237,7 +257,9 @@ def send_keys(self):
               )
             )
             os.remove(tmp)
-        self.run_command("touch /run/keys/done")
+        self.run_command("mkdir -m 0750 -p /run/keys && "
+                         "chown root:keys  /run/keys && "
```
ip1981: Why is it necessary?
mkdir won't change mode if the  path already exists.
I would expect everything to be idempotent.
/run/keys is actually created ny NixOS' initrd.

nix/keys.nix
deepfire: @domenkozar, I need you two to come to a consensus : -)

ip1981: I can stand "folders", no problem.

nixops/backends/__init__.py
ip1981: Given `destinationFolder` exists, it's strange to see raw `/run/keys` anyway.

deepfire: @ip1981, that's because all key might have an overridden `destinationFolder`, but we need a well-known location for the `done` flag -- and that`s `/run/keys`..

Just think about it as if the snippet at lines 214-215 was moved down, here.  Makes it simpler to understand the change, really.

ip1981: Ok, makes sense. Although this known location is _only_ used in https://github.com/NixOS/nixops/blob/master/nix/keys.nix (needs update as well). And for me the most wanted consequence of this is reboot-able servers, which do not lose keys after reboot. If any service is bound to "old" `/run/keys/done` it is a bug (or a DRY sin).

nix/keys.nix
domenkozar: I'm fine with `destDir`

domenkozar:
This needs a test case in https://github.com/NixOS/nixops/blob/master/tests/functional/test_send_keys_sends_keys.py since it's critical to work

deepfire:
@domenkozar, done, using https://github.com/NixOS/nixops/pull/668

domenkozar:
@rbvermaa @edolstra any comments? Looks good to me, I still need to test.

domenkozar:
I've also fixed #646, will test this tomorrow.

domenkozar:
I've tested this on ec2 backend and it works.

Issue #649, opened 2017-04-15
Support multiple versions of nixpkgs in one network


grahamc:
I run a network of several servers. Some of them are standard and work fine with the current stable NixOS, however some of them are fairly exotic and require features on a custom nixpkgs, and some require features from NixOS unstable.

For this reason I would like to be able to express a desired nixpkgs to use on a per-server basis. Example:

```
{
  server1 = { }: { ...} # uses the default nixpkgs

  server2 = { }: {
    deployment.nixpkgs = "./custom-nixpkgs";
  };

  server3 = {}: {
    deployment.nixpkgs = "https://d3g5gsiof5omrk.cloudfront.net/nixos/unstable/nixos-17.09pre105617.26f5fa8f97";
  }
}
```

This may also be helpful in a 'canary' deployment story.

I believe the changes would be here: https://github.com/NixOS/nixops/blob/master/nix/eval-machine-info.nix#L48

grahamc:
I think:
 - nix.nixPath only takes effect next build _on_ the server
 - nixops doesn't use the target server's path anyway
 - I need my path to be set up prior to evaluating the config                                                                      

I think `value = import <nixpkgs/nixos/lib/eval-config.nix>` would be changed to `value = import "${server.nixpkgs}/nixos/lib/eval-config.nix>` ... or something.

grahamc:
I tried:

``
        importSources =
          concatMap (module:
              if (builtins.hasAttr "deployment" module)
                  && (builtins.hasAttr "nixpkgs" module.deployment)
              then ["${module.deployment.nixpkgs}"]
              else [])
            ([{deployment = { nixpkgs = <nixpkgs>;};}] ++ networks);
        importSource = last importSources;
      in
      { name = machineName;
        value = import "${importSource}/nixos/lib/eval-config.nix" {
```

but found numerous that I couldn't work through. I'm thinking at a minimum, we can't accept URLs but instead use `deployment.nixpkgs = pkgs.fetchUrl`. Still experimenting.

**Comment by grahamc** [comment]
Looking again, it isn't clear to me if pkgs in `${importSource}/nixos/lib/eval-config.nix` would refer to `${importSource}/pkgs` or the outer `NIX_PATH`'s nixpkgs.

**Comment by grahamc** [comment]
Based on `eval-config.nix`'s:

```
with import <nixpkgs/nixos/lib/testing.nix> { inherit system; };
with pkgs;
```

and `testing.nix`'s:

```
[...]

PR #665, opened 2017-05-13
Support multiple versions of nixpkgs in one network

grahamc:
Having a machine named foo, described as:

    foo = { # ...snipped...
      deployment.nix_path.nixpkgs = (builtins.filterSource
        (path: type: type != "directory" || baseNameOf path != ".git")
        ./../nixpkgs);
    });

will have the custom nixpkgs set in the `NIX_PATH` as
`nixpkgs=path-to-custom-nixpkgs`.

Note this does not work with foo = { config, ... }: {... machines, but
having a second nix file in the network would work, and also:

    let
      canary = machine: {
        deployment.nix_path.nixpkgs = (builtins.filterSource
          (path: type: type != "directory" || baseNameOf path != ".git")
          ./../nixpkgs);
        imports = [machine];
      };

      machine = { ... }: {
        # your machine config
      };
    in {
      machineA = machine;
      machineB = canary machine;
    }

Note that because this uses scopedImport, the nixops network and
machines may use `import <nixpkgs>` and have a consistent view of
nixpkgs.

Closes https://github.com/NixOS/nixops/issues/649#issuecomment-294330624

### DIFF
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -39,13 +39,42 @@ rec {
         # Get the configuration of this machine from each network
         # expression, attaching _file attributes so the NixOS module
         # system can give sensible error messages.
+
         modules =
           concatMap (n: optional (hasAttr machineName n)
             { imports = [(getAttr machineName n)]; inherit (n) _file; })
           networks;
-      in
-      { name = machineName;
-        value = import <nixpkgs/nixos/lib/eval-config.nix> {
+
+        machineConfs =
+          concatMap (n: optional (hasAttr machineName n)
+            (getAttr machineName n))
+          networks;
+
+        nameToPath = attrs: name: {
+          prefix = name;
+          path = attrs."${name}";
+        };
+
+        attrsetToPaths = attrset: map (nameToPath attrset)
+          (builtins.attrNames attrset);
+
+        importSources =
+          (concatMap (module:
+              if (!builtins.isFunction module
+                  && builtins.hasAttr "deployment" module)
+                  && (builtins.hasAttr "nix_path" module.deployment)
+              then attrsetToPaths module.deployment.nix_path
+              else [])
+            machineConfs) ++ builtins.nixPath;
+
+        __nixPath = importSources;
+
+        machineImport = builtins.scopedImport {
+          inherit __nixPath;
+        };
+      in {
+        name = machineName;
+        value = machineImport <nixpkgs/nixos/lib/eval-config.nix> {
           modules =
             modules ++
             defaults ++
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -103,6 +103,13 @@ in
       '';
     };
 
+    deployment.nix_path = mkOption {
+      default = {};
+      type = types.attrsOf types.str;
+      description = ''
+      '';
+    };
+
     deployment.hasFastConnection = mkOption {
       default = false;
       type = types.bool;


grahamc:
Thank you, @cleverca22, for the excellent suggestions & help with scopedImport.

grahamc:
I'd love a review by @edolstra and @rbvermaa.

ip1981:
Multiple nixpkgs look like an overkill to me. Usually overriding individual packages and using custom modules is enough.

grahamc:
Depends on the size and variety of the network. Three critical use cases of mine:

1. Half x86-64 nodes (stable), half aarch64 nodes (a patched master). This requires completely separate nixpkgs versions.

2. Canary and beta deployments, allowing rolling out a nixpkgs update in a staged / predictable pattern.

3. Modules which carry patches and can't be upstreamed

nix/options.nix
```diff
@@ -103,6 +103,13 @@ in
       '';
     };
 
+    deployment.nix_path = mkOption {
+      default = {};
+      type = types.attrsOf types.str;
```
ip1981: Maybe `path` is more adequate?

ip1981:
Does it work well with modules, overrides, etc.?

ip1981:
And, of course, there is a plain workaround - different deployments with different nix path :)

grahamc:
> modules
It works well with `export NIXOS_EXTRA_MODULE_PATH=...`

> overrides
like `pkgs.my-package.overrideAttrs` ? Can you provide an example of this use so I can be sure?

> different deployments with different nix path :)
Yes, different deployments or nixops deploy with different `NIX_PATH` in the environment variables / `--exclude` / `--include` parameters, but I think I can interpret your `:)` as not needing an explanation on why that is painful :P / not very nice.

ip1981:
> like `pkgs.my-package.overrideAttrs` ? Can you provide an example of this use so I can be sure?

I mean `nixpkgs.config.packageOverrides` for a while:

```nix
{
  # ...
  machineFoo = {config, ...} : {
    imports = [ <my-modules> ]; # with `nixpkgs.config.packageOverrides`, probably recursive.
    # ...
  };
  # ...
}
```

My point is to be able to use hierarchy of modules (via `imports`) in which topmost modules (who imports) can override packages defined in underlying modules (who is imported). At all levels. So no one is locked to nixpkgs.

ip1981:
As far as I understand, you have to evaluate machine's spec in order to get it's specific nix path. And once evaluated, new nix path will override all packages overridden during evaluation. Correct?

nix/eval-machine-info.nix
```diff
@@ -39,13 +39,42 @@ rec {
...
+          prefix = name;
+          path = attrs."${name}";
+        };
+
+        attrsetToPaths = attrset: map (nameToPath attrset)
+          (builtins.attrNames attrset);
+
+        importSources =
+          (concatMap (module:
+              if (!builtins.isFunction module
```
ip1981: Why can't it be a function? Is `module` a "machine" definition? If yes, it usually a function.

grahamc:
> As far as I understand, you have to evaluate machine's spec in order to get it's specific nix path.

No, we don't evaluate any functions to find the nix path. This also answers your question about why it can't be a function: We can't call a machine's function without passing in a `pkgs`, and we don't have that when we're looking for which nixpkgs to include.

For this reason, the imports and overrides all work just fine.

Examples, this works:

```nix
{
  machineB = {
    deployment.nix_path.nixpkgs = "https://github.com/nixos/nixpkgs/archives/master.tar.gz";
  }
}
```

This won't work, because it is a function:
```nix
{
  machineB = { ... }: {
    deployment.nix_path.nixpkgs = "https://github.com/nixos/nixpkgs/archives/master.tar.gz";
  }
}
```

See also my canary example in the issue body for how to use both functions and attrsets together.

nix/eval-machine-info.nix
grahamc: See https://github.com/NixOS/nixops/pull/665#issuecomment-301757938: we can't introspect a function in this way.

ip1981:
Ok, I think I understood.

What you are doing is exactly `nixops -I ... --include ...`, but you do it in Nix, not in Bash :)
But since you do it in Nix with modules and options it might suggest, that one can use all the power of modules. But I believe that is not true. You have to evaluate modules to extract nix path, then evaluated again with new path. At least this is what I'd expect from modules. If you don't evaluate modules to extract  nix path it's dirty cheating :)

I remember a try to make nixops modular with a mean to define nix path. This didn't end well. See https://github.com/zalora/nixops.

I think wrappers of nixops is simpler, safer and cleaner solution (`nixops -I ... --include ...`)

ip1981:
NixOS modules is extremely cool and powerful feature, maybe underestimated yet. So I'd say no to features that do not integrate flawlessly into module system (like not working with functions) `¯\_(ツ)_/¯`

grahamc:
> But I believe that is not true.

Yes, that is true :( a `mkForce` inside a module utilized from a function will not work. That is not very nice.

What if we took advantage of the `network` nixops attrset key:

```nix
{
  network = {
    description = "Load balancer test";
    default_nix_path = {
      nixpkgs = "./stable-nixpkgs";
    };
    nix_path_overrides = {
      machineB = {
        nixpkgs = "./unstable-nixpkgs";
      };
    }
  };

  machineB = { ... }: {}
}
```

Now it is out of band, yet declared in the file?

domenkozar:
Maybe `network.nixpath` and `deployment.nixpath`?

grahamc:
The problem Igor is pointing out is that if you inside a module set the Nix
path it may or may not take effect, depending on if it was set inside a
function or not.

ip1981:
Well, yes. To me, the only feasible way is to have nix path defined somewhere out of machine's "mega-module". I don't really  care if it would be bash or nix :).

You can define you `path.nix` mapping machines to nix path, and then ... build a wrapper or whatever :) Even make use of nix-shell.

nix/options.nix
edolstra: It should be `nixPath`. (NixOS already has a `nix.nixPath` option.)

Arguably the type should be `types.listOf types.str`, like `nix.nixPath`. Nix search path entries are not required to have a prefix.

nix/eval-machine-info.nix
```diff
@@ -39,13 +39,42 @@ rec {
...
+                  && builtins.hasAttr "deployment" module)
+                  && (builtins.hasAttr "nix_path" module.deployment)
+              then attrsetToPaths module.deployment.nix_path
+              else [])
+            machineConfs) ++ builtins.nixPath;
+
+        __nixPath = importSources;
+
+        machineImport = builtins.scopedImport {
```
edolstra: IIRC, `scopedImport` slows down evaluation because it disables the file cache in the evaluator. If that's still the case, `machineImport` should be something like `machineImport = if __nixPath == [] then import else builtins.scopedImport ...`.

edolstra:
Semantically, it would be cleanest to move the NixOS configuration of a machine to an option below the machine, so we can add options like `nixPath` to the machine that are not part of the NixOS configuration. E.g.
```
machines.foo =
  { nixPath = ...;
    nixosConfig = { config, pkgs, ... }:
      { environment.systemPackages = [ pkgs.foo ];
      };
  };
```
This allows `machines` (and the entire NixOps network specification) to be defined using the module system, e.g.
```
machines = mkOption {
  type = types.attrsOf (submodule {
    options = {
      nixPath = mkOption { ... };
      nixosConfig = mkOption { ... };
      ...
    };
  };
};
```
This is more elegant than having a separate `network.nixPathOverrides` option.

However it's not clear how to do this in a backwards-compatible way...

grahamc:
If I manage to make a PR implementing @edolstra's suggestion, can that possibly merge? (cc @rbvermaa)

moretea:
We could deprecate the current way, and add a compatibility layer.

This compatibility layer will just map over the attr/value pairs, and convert them to a format that is compatible with the module based version.

Would that be acceptable @edolstra @rbvermaa @grahamc?

mickours:
I'm really interested in this feature. Any chance this would be merged?

timclassic:
I am also interested in this feature.  @grahamc, did you ever move forward with your PR attempt?

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #670, opened 2017-05-20
Fix #669: Use security group id in the api call in case it's deployed in a VPC

### LINKED ISSUE #669: Deploy with --check of an AWS security group in vpc fails due to wrong api call
In case a security group is deployed in VPC, `nixops deploy --check` will fail as the boto api call should use the sg id instead of the name.
```
$ nixops deploy  --check
error: EC2ResponseError: 400 Bad Request
<Response><Errors><Error><Code>InvalidParameterValue</Code><Message>Invalid value 'charon-d6c4461c-373c-11e7-be3e-0ae168952ac0-sg' for groupName. You may not reference Amazon VPC security groups by name. Please use the corresponding id for this operation.</Message></Error></Errors><RequestID>ea21316b-0308-4313-85f2-9ba1cadf8960</RequestID></Response>
```

AmineChikhaoui:


AmineChikhaoui:
oops, will fix it.

domenkozar:
Is this causing `node10......................> got (possibly transient) EC2 error code 'InvalidParameterValue': Value () for parameter groupId is invalid. The value cannot be empty. retrying...`

domenkozar:
Hmm, no that seems to be the problem of instance not having any SG provisioned.

domenkozar:
Or maybe: `The security group 'charon-669b6a14-46c1-11e7-bef2-06258a1e40fd-allow-all-eu-west-1' does not exist in default VPC 'vpc-7fbdc01a'`

domenkozar:
Nevermind, created https://github.com/NixOS/nixops/issues/674

AmineChikhaoui:
@rbvermaa it should be fixed now

AmineChikhaoui:
@rbvermaa I think you accidentally pushed some stuff for debugging in master https://github.com/NixOS/nixops/commit/2a8d9b99d0054bdae285f1e4d1ca712c4d91e2a7.

OTOH, I'm changing this PR to WIP until I check what are the remaining issues with deploying security groups with --check.

PR #682, opened 2017-06-16
automatically resolve channel uris

AmineChikhaoui:
This is needed to get the same behavior in other nix-* commands by specifying which channel to use by just doing `-I nixpkgs=channel:nixos-17.03`


edolstra:
Thanks, I just ran into this yesterday!

PR #688, opened 2017-06-20
Allow using Hetzner with existing sub-accounts

### LINKED ISSUE #592: Using Hetzner with an Admin user (per-server sub-account)
When using the Hetzner backend, nixops always tries to do `creating an exclusive robot admin account for ...`.

I already have an exclusive sub-account for my machine, and I'd like nixops to use that straight away, and not create one itself, so that I don't have to give nixops access to my entire Hetzner Webservice API credentials (which in fact also didn't work for me, as the `hetzner` Python package that it uses then complains that [`The user {0} is a dedicated web service user`](https://github.com/RedMoonStudios/hetzner/blob/280f8d3105937d46da21d534f82315f83db3ed84/hetzner/robot.py#L86), but that would be a separate issue, let me know if I should file it).

The trouble is that I can't set the user/password for the sub-account with environment variables; the current code runs:

```python
        if not self.robot_admin_user or not self.robot_admin_pass:
            self.log_start("creating an exclusive robot admin account for "
                           "‘{0}’... ".format(self.name))
```

and `self.robot_admin_pass`/`self.robot_admin_user` don't come from the `HETZNER_ROBOT_USER`/`HETZNER_ROBOT_PASS` environment variables (only `robot_user` and `robot_pass` do, and setting only those doesn't work).

It would be great if I could tell nixops via environment variables the user/password to use an already created per-machine sub-account.

nh2:
Fixes #592.

See commit messages and added documentaiton for details.

### DIFF
--- a/doc/manual/overview.xml
+++ b/doc/manual/overview.xml
@@ -1361,12 +1361,21 @@ machine3.deployment.azure = {
 to their server management interface, called the <link
 xlink:href="https://robot.your-server.de/">Robot</link>. This account is
 <emphasis>only</emphasis> used for the initial deployment and the destruction of
-a machine. In particular the initial deployment creates a separate Robot account
+a machine. In particular the initial deployment creates a separate Robot sub-account
+(Hetzner calls this the "Admin login" because you'd give it to your server's sysadmin)
 just for the machine that's going to be created, so a person who has access to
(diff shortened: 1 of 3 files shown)


domenkozar:
cc @aszlig

nh2:
CC @basvandijk

Issue #689, opened 2017-06-20
Make nixops derivation overridable


nh2:
I'd like to run `nixops` from nixpkgs, but `override` the derivation to give it a custom nixops commit.

It seems that currently I can't do that, because the [nixpkgs derivation builds from a tarball](https://github.com/NixOS/nixpkgs/blob/1c366b4e0647abea0385395dcb2981a5b3161cf8/pkgs/tools/package-management/nixops/default.nix#L6) instead from nixops's git directory.

Is it possible to write a `mkDerivation` that uses nixops straight from git, so that I can point it at my custom branch?

domenkozar:
You can build it like so:

```
$ cat nixops-unstable.nix
let
  nixopsUnstable = builtins.fetchTarball "https://github.com/NixOS/nixops/archive/4a204e28daef51418f9116170dcead44849546ad.tar.gz";
in (import "${nixopsUnstable}/release.nix" {}).build.x86_64-linux
```

and then `nix-build nixops-unstable.nix`

nh2:
OK I'll give that a shot and will report.

domenkozar:
ping :)

teto:
Similarly I've been trying to override nixops with a local fork where I have stacked several PRs.

I am interested in how people override nixops as well, I tried from an overlay
```
  nixops-dev = super.nixops.overrideAttrs ( src: {

    name = "nixops-dev";
    src = builtins.fetchGit {
      url = /home/teto/nixops;
      rev = "7c71333a3ff6dc636d0b2547f07b105571a3027b";
    };
  });
```
but I consistently end up with
```
adding 'nixops/resources/vpc_route_table.py'
adding 'nixops/resources/vpc_route_table_association.py'
adding 'nixops/resources/vpc_subnet.py'
adding 'nixops-_version_.data/scripts/nixops'
adding 'nixops-_version_.dist-info/DESCRIPTION.rst'
adding 'nixops-_version_.dist-info/metadata.json'
adding 'nixops-_version_.dist-info/top_level.txt'
adding 'nixops-_version_.dist-info/WHEEL'
adding 'nixops-_version_.dist-info/METADATA'
adding 'nixops-_version_.dist-info/RECORD'
installing
/build/source/dist /build/source
nixops-_version_-py2-none-any.whl is not a valid wheel filename.
```
In a repl, nixops-dev version looks good. The `@version@` in setup.py doesn't seem to be substituted

noonien:
[...]

PR #691, opened 2017-06-20
release.nix: Make `shellHook` work even when started from a different dir

nh2:
This provides at least a workaround for #689, and seems to have no drawbacks.

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -96,9 +96,9 @@ rec {
         ];
 
       # For "nix-build --run-env".
-      shellHook = ''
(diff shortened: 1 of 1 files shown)


rbvermaa:
This makes the PYTHONPATH point to a copy of the source code.

nh2:
> This makes the PYTHONPATH point to a copy of the source code.

@rbvermaa I don't get it, isn't that the goal? Why is that a problem?

Or maybe I'm not understanding how that `shellHook` is supposed to be used?

Issue #694, opened 2017-06-28
DataDog monitor destruction response isn't properly analysed


deepfire:
As can be seen at https://github.com/NixOS/nixops/blob/master/nixops/resources/datadog-monitor.py#L127, the response to a monitor removal request isn't analysed, and I quote:

```
    self.connect(self.app_key,self.api_key)
    if self.monitor_exist(self.monitor_id) == False:
        self.warn("datadog monitor with id {0} already deleted".format(self.monitor_id))
    else:
        self.log("deleting datadog monitor ‘{0}’...".format(self.monitor_name))
        self._dd_api.Monitor.delete(self.monitor_id)
```

This can sometimes lead to a combination of:
```
cardano_node_process.......> deleting datadog monitor ‘cardano-node process is down’...
```
..and the fact DataDog monitor being still alive and well, as can be seen through the Web UI.

AmineChikhaoui:
hm thanks @deepfire , will look into it

deepfire:
According to https://datadogpy.readthedocs.io/en/latest/#datadog.api.Monitor:

```
 delete(id, **params)

    Delete an API resource object
    Parameters:	id (id) – resource object to delete
    Returns:	Dictionary representing the API’s JSON response
```

AmineChikhaoui:
@deepfire yeah I originally though that datadog api will raise an exception in case of errors but should be easy to handle that but I'm curious what can go wrong in delete monitor api call, was there changes to the monitor from outside nixops and if so what kind of changes ?

PR #695, opened 2017-06-28
catch exceptions while deleting datadog resources

AmineChikhaoui:
there was some issues reported in #694, this will make the `nixops destroy` raise exceptions at least instead of failing silently.

### DIFF
--- a/nixops/resources/datadog-monitor.py
+++ b/nixops/resources/datadog-monitor.py
@@ -54,12 +54,12 @@ class DatadogMonitorState(nixops.resources.ResourceState):
     @property
     def resource_id(self):
         return self._monitor_url + self.monitor_id if self.monitor_id else None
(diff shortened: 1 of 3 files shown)


domenkozar:
@AmineChikhaoui btw, should datadog monitor names be unqiue across deployment networks?

AmineChikhaoui:
@domenkozar no you can have the same name across many monitors, the name is basically the subject of the alert email and what's showing in the datadog ui. For uniqueness, what matters is the monitor id which is generated.

domenkozar:
@AmineChikhaoui interesting :) We had a case where deployment was `nixops destroy`-ed, but the machines part of that deployment were still triggering alerts as CRITICAL.

Most probably datadog keeps the last known state if the node is down.

Issue #648, opened 2017-04-13
Restructure the ResourceState class to make it dictionary based (store the state as a key value dict)


AmineChikhaoui:
Currently the Resource state attributes are handled by setting them as python **property** using `nixops.util.attr_property` then their _set_attr/_get_attr methods are overridden to make them able to read from the state file or write to it automatically.
On the other hand the resource definition was migrated to a dictionary based structure in [this commit](https://github.com/NixOS/nixops/commit/b0770ee2d946ebce8624778ef630718629f63494) to avoid the extra xml parsing.
As I'm thinking about implementing something to store the diff between the state/defn (end goal is to have something that gives the plan like what terraform does), I thought we can add an attribute to the ResourceState class e.g _state which will be using for example StateDict class that either inherit from **dict** or implement the dictionary interface. In the implementation of such StateDict class we can make \__getitem\__/\__setitem\__  our access point to the state file i.e statements like the following will handle the state file like what we had previously with minimal changes
```
with  self.depl._db:
  self._state['region'] = defn.config['region'] # the self._state['key']=val statement will write to the state file
```
and `self._state['region']` will read from the state file and return the value.
Of course we need to handle stuff like default values and make the class behave exactly like a regular dictionary otherwise it can cause confusion and unexpected behavior. But first, I wanted to have some opinions about such changes before starting any actual implementation.
I think if we have such structure, doing a diff will be much cleaner by comparing the keys of 2 similar dictionaries. So any thoughts ?
attn @edolstra @aszlig

edolstra:
What do you mean with "storing the diff"? I assume you just *compute* the diff between the state and the definition at runtime and then execute some actions depending on the diff.

How should we handle state elements that do not have a corresponding definition element? For example, EC2 instances have an instance ID in their state but not their definition.

Do you have some ideas on how the diff output will be used? I'm thinking that diff engine, if it detects that some attribute `foo` has changed, could call a method `realize_foo_change` on the resource (e.g. `realize_instance_type_change` would stop the instance, change the instance type and then start it again). If the required method doesn't exist, it would print a warning or error (e.g. "resource X does not support changing the 'region' attribute"). Currently this is all done in an ad hoc / incomplete way in the `create` method, so this would a big improvement.

AmineChikhaoui:
@edolstra yeah sorry it wasn't clear by "storing" just during the runtime and not something persistent.
[...]

PR #697, opened 2017-07-06
AWS VPC resources + Diff engine

AmineChikhaoui:
This PR adds integration for AWS VPC resources and an implementation of the ideas discussed in #648 .
The suggested way of implementing resources is by using a diff engine that can be instantiated using `nixops.resources.ec2_common.setup_diff_engine()` this diff engine will handle the comparison between the state and the target configuration and then generate a plan which is composed of the various handlers that needs to be called to realize the target state.
A resource can basically define the handlers and setup the right dependencies between them e.g `self.handle_entries = Handler(['entries'], after=[self.handle_create_network_acl], handle=self.realize_entries_change)`
This means that this handler will handle the entries option changes and will run after create_network_acl if they're invoked in the same deploy operation and will use the method realize_entries_change to realize the required changes.
There is an experimental option **--plan-only** (supported only for the VPC resources which can be added to the `nixops deploy` cmd to show the changes that will be made and exit.

**Note that** this doesn't implement a global DAG for the resources in a network but rather for the attributes handlers per resource.

For the VPC resources, what was already implemented is:
- [x] VPC 
- [x] VPC subnet
- [x] VPC internet gateway 
- [x] VPC egress only internet gateway
- [x] VPC NAT gateway 
- [x] VPC DHCP options
- [x] VPC network ACL
- [x] Had to also rewrite the elastic IP resource to handle the vpc domain.
- [x] VPC customer gateway
- [x] VPC network interface
- [x] VPC network interface attachement
- [x] VPC route table
- [x] VPC route
- [x] VPC route table association
- [x] VPC endpoint
- [x] VPN connection/VPN gateway
- [x] Update ec2 instances + old resources to be able to accept a vpc resource in the vpcId options

The remaining work:
- [ ] VPC peering connection (will be done in a separate PR due to some complexity in the resource state transitions)
- [ ] Add examples/documentation.
  
I would appreciate some initial feedback on the suggested implementation of #648 though to be able to make appropriate changes if needed.


domenkozar:
I'm happy to test this once it's ready :)

edolstra:
What's the best way to play with this? Deploy `examples/vpc.nix`?

AmineChikhaoui:
@edolstra yep, examples/vpc.nix should be good to play with the changes.

Also, @domenkozar if you have some use cases that you can start testing with this PR that would be awesome ;)

domenkozar:
Not really (yet!), but I'd like to find some time to review - hopefully in a few days :)

rehno-lindeque:
It looks like setting the environment variable `EC2_ACCESS_KEY` instead of explicitly using `accessKeyId` doesn't work with any of the new resources.

```
$ nixops deploy
error: The option `accessKeyId' is used but not defined.
```

(An easy workaround exists though):

```nix
  resources = {
    vpc.my-vpc = {
      region = awsRegion;
      accessKeyId = builtins.getEnv "EC2_ACCESS_KEY"; # Set accessKeyId from environment
      instanceTenancy = "default";
      enableDnsSupport = true;
      enableDnsHostnames = true;
      cidrBlock = "10.0.0.0/16";
      tags = {
        Source = "NixOps";
      };
    };
  };
```

AmineChikhaoui:
@rehno-lindeque Oh right, yeah that should be easy to fix by adding a default value to accessKeyId option. I will add that, but don't think it's really specific to the new resources as I see few old ones having the same option declaration. Thanks !

rehno-lindeque:
I'm finding that subnet resources seem to be recreated on every deploy (even when nothing changes in the expression at all).

E.g.

```
$ nixops create deploy.nix
$ nixops deploy

   # Everything goes fine...

$ nixops deploy
error: subnet subnet-051be83a definition changed and it needs to be recreated use --allow-recreate if you want to create a new one

$ nixops deploy --allow-recreate
my-network-subnet.....> warning: subnet definition changed, recreating...
my-network-subnet.....> deleting subnet subnet-051be83a
my-network-subnet.....> got (possibly transient) EC2 error '{'Message': "The subnet 'subnet-051be83a' has dependencies and cannot be deleted.", 'Code': 'DependencyViolation'}'. retrying...
my-network-subnet.....> got (possibly transient) EC2 error '{'Message': "The subnet 'subnet-051be83a' has dependencies and cannot be deleted.", 'Code': 'DependencyViolation'}'. retrying...
my-network-subnet.....> got (possibly transient) EC2 error '{'Message': "The subnet 'subnet-051be83a' has dependencies and cannot be deleted.", 'Code': 'DependencyViolation'}'. retrying...
my-network-subnet.....> got (possibly transient) EC2 error '{'Message': "The subnet 'subnet-051be83a' has dependencies and cannot be deleted.", 'Code': 'DependencyViolation'}'. retrying...
```

My resources look like this:

```nix
let
  networkName = "my-network";
  awsRegion = "us-east-1";
in
{
  resources = {
    ec2KeyPairs.keypair = { region = awsRegion; };
    vpc."${networkName}-vpc" = {
      region = awsRegion;
      instanceTenancy = "default";
      enableDnsSupport = true;
      enableDnsHostnames = true;
      cidrBlock = "10.0.0.0/16";
      tags = {
        Source = "NixOps";
      };
    };
    ec2SecurityGroups = {
      sg =
        { resources, lib, ... }:
        {
          region = awsRegion;
          vpcId = resources.vpc."${networkName}-vpc";
          rules = [
            { toPort = 22; fromPort = 22; sourceIp = "0.0.0.0/0"; }
          ];
        };
    };
    vpcSubnets = {
        "${networkName}-subnet" =
          { resources, ... }:
          {
            region = awsRegion;
            vpcId = resources.vpc."${networkName}-vpc";
            cidrBlock = "10.0.0.0/16";
            zone = "";
            mapPublicIpOnLaunch = true;
            tags = { Source = "NixOps"; };
          };
      };
    vpcRouteTables = {
      route-table =
        { resources, ... }:
        {
          region = awsRegion;
          vpcId = resources.vpc."${networkName}-vpc";
        };
    };
    vpcRouteTableAssociations = {
      "${networkName}-association" =
        { resources, ... }:
        {
          region = awsRegion;
          subnetId = resources.vpcSubnets."${networkName}-subnet";
          routeTableId = resources.vpcRouteTables.route-table;
        };
    };
    vpcRoutes = {
      igw-route =
        { resources, ... }:
        {
          region = awsRegion;
          routeTableId = resources.vpcRouteTables.route-table;
          destinationCidrBlock = "0.0.0.0/0";
          gatewayId = resources.vpcInternetGateways.igw;
        };
    };
    vpcInternetGateways.igw = 
      { resources, ... }:
      {
        region = awsRegion;
        vpcId = resources.vpc."${networkName}-vpc";
      };
  };
}
```

AmineChikhaoui:
@rehno-lindeque yeah well that's a limitation that I had to do. So you're setting `zone = ""` which makes the api choose a zone for you then store it in the state, the issue is in the next deploy the definition ("" in your case) and the state will be different which triggers the handler of the zone (same handler that creates the subnet as obviously we can't change the subnet zone after creation).
I'll maybe make sure in the implementation to throw an exception to enforce users to set a zone. But anyway, with the current implementation the zone needs to be set eventhough the AWS api can handle it.

rehno-lindeque:
Thanks a lot @AmineChikhaoui, I was just about to update my comment :+1:

nixops/resources/aws_vpn_connection.py
```diff
@@ -0,0 +1,146 @@
...
+
+       customer_gtw_id = config['customerGatewayId']
+       if customer_gtw_id.startswith("res-"):
+           res = self.depl.get_typed_resource(customer_gtw_id[4:].split(".")[0], "vpc-customer-gateway")
+           customer_gtw_id = res._state['customerGatewayId']
+
+       vpn_gateway_id = config['vpnGatewayId']
+       if vpn_gateway_id.startswith("res-"):
+           res = self.depl.get_typed_resource(vpn_gateway_id[4:].split(".")[0], "aws-vpn-gateway")
+           vpn_gateway_id = res._state['vpnGatewayId']
```
edolstra: Since this idiom appears so frequently, we should create a helper function for it, e.g.
```
vpn_gateway_id = get_resource('vpnGatewayId', "aws_vpn-gateway")
```

Also, we should check whether `res._state['vpnGatewayId']` is initialized. I'm not sure if that's guaranteed to be the case when using `nixops deploy --include ...` (since some dependencies might be excluded).

nixops/resources/aws_vpn_connection.py
```diff
@@ -0,0 +1,146 @@
...
+    state = nixops.util.attr_property("state", nixops.resources.ResourceState.MISSING, int)
+    access_key_id = nixops.util.attr_property("accessKeyId", None)
+    _reserved_keys = EC2CommonState.COMMON_EC2_RESERVED + ["vpnConnectionId"]
+
+    def __init__(self, depl, name, id):
+        nixops.resources.ResourceState.__init__(self, depl, name, id)
+        self._state = StateDict(depl, id)
+        self._client = None
+        self.handle_create_vpn_conn = Handler(['region', 'vpnGatewayId', 'customerGatewayId', 'staticRoutesOnly'], handle=self.realize_create_vpn_conn)
+        self.handle_tag_update = Handler(['tags'], after=[self.handle_create_vpn_conn], handle=self.realize_update_tag)
```
edolstra: So I noticed that tags don't get updated, but it's not clear to me why. Shouldn't handle_tag_update trigger whenever the tags attribute changes?

nixops/resources/aws_vpn_connection.py
```diff
@@ -0,0 +1,146 @@
...
+                    self.warn("vpn connection {} was already deleted".format(self._state['vpnConnectionId']))
+                else:
+                    raise e
+
+        with self.depl._db:
+            self.state = self.MISSING
+            self._state['vpnConnectionId'] = None
+            self._state['vpnGatewayId'] = None
+            self._state['customerGatewayId'] = None
+            self._state['staticRoutesOnly'] = None
```
edolstra: It would be nice if we could eliminate this kind of boilerplate code:

* The check for `self.UP` could be eliminated if we only call `destroy` in a created state. (If `destroy` were to be implemented as a diff handler implementing an UP -> UNKNOWN transition, this would be automatic since `UP` would be a precondition for calling the handler.)

* Unsetting all _state fields could be done automatically after calling `destroy`.

nixops/resources/aws_vpn_connection_route.py
```diff
@@ -0,0 +1,126 @@
...
+    @classmethod
+    def get_type(cls):
+        return "aws-vpn-connection-route"
+
+    @classmethod
+    def get_resource_type(cls):
+        return "awsVPNConnectionRoutes"
+
+    def show_type(self):
+        return "{0}".format(self.get_type())
```
edolstra: I wonder if we can eliminate having a separate `<Resource>Definition` class. It's pretty much all boilerplate code at this point...

nixops/resources/aws_vpn_connection_route.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @property
+    def resource_id(self):
+        return self._state.get('destinationCidrBlock', None)
+
+    def prefix_definition(self, attr):
+        return {('resources', 'awsVPNConnectionRoutes'): attr}
+
+    def get_definition_prefix(self):
+        return "resources.awsVPNConnectionRoutes."
```
edolstra: Can't `get_definition_prefix` be derived from `prefix_definition`, or vice versa?

edolstra: In fact, given that `AWSVPNConnectionRouteDefinition.get_resource_type` already returns `awsVPNConnectionRoutes`, maybe these methods can be avoided entirely?

nixops/resources/aws_vpn_gateway.py
```diff
@@ -0,0 +1,155 @@
...
+                raise Exception("VPN gateway {} defintion changed"
+                                " use --allow-recreate if you want to create a new one".format(
+                                    self._state['vpnGatewayId']))
+            self.warn("VPN gateway changed, recreating...")
+            self._destroy()
+            self._client = None
+
+        self._state['region'] = config['region']
+
+        self.connect()
```
edolstra: It would be nice to get rid of these `self.connect()` calls by replacing `self._client` with something like `self.get_client()` (which would cache the client object).

nixops/resources/ec2_common.py
```diff
@@ -1,10 +1,14 @@
 import socket
 import getpass
+
 import nixops.util
 import nixops.resources
+from nixops.diff import Diff, Handler
 
 class EC2CommonState():
 
+    COMMON_EC2_RESERVED = ['accessKeyId', 'ec2.tags']
```
edolstra: Why is `ec2.tags` reserved?

nixops/resources/vpc.py
```diff
@@ -0,0 +1,275 @@
...
+            self._state['region'] = None
+            self._state['cidrBlock'] = None
+            self._state['instanceTenancy'] = None
+            self._state['enableDnsSupport'] = None
+            self._state['enableDnsHostname'] = None
+            self._state['enableVpcClassicLink'] = None
+
+    def create_after(self, resources, defn):
+        return {r for r in resources if
+                isinstance(r, nixops.resources.elastic_ip.ElasticIPState)}
```
edolstra: Random thought: we could get rid of the whole `create_after` ordering by constructing the *actual* dependency graph of resources (rather than an ordering between resource types). I.e., if a resource X has an attribute that refers to `res-Y`, then it has a dependency on Y.

This would eliminate more boring code (`create_after`) and improve parallelism since more resources could be created in parallel.

nixops/resources/vpc_network_interface_attachement.py
```diff
@@ -0,0 +1,202 @@
...
+            if response.get('Attachment', None):
+                if response['Attachment']['Status'] == 'attached':
+                    break
+                elif response['Attachment']['Status'] != "attaching":
+                    raise Exception("eni attachment {0} in an unexpected state {1}".format(
+                        eni_id, response['Attachment']['Status']))
+                self.log_continue(".")
+                time.sleep(1)
+            else:
+                raise Exception("eni {} doesn't have any attachement {}".format(eni_id))
```
edolstra: There are a lot of these `wait_for_X` methods and they all look very similar. Could the commonality be factored out? (We already have `nixops.util.check_wait` for waiting in a loop, see e.g. `wait_for_volume_available`.)

Also `attachement` -> `attachment`.

nixops/resources/vpc.py
```diff
@@ -0,0 +1,275 @@
...
+
+        self.ensure_state_up(check)
+
+    def ensure_state_up(self, check):
+        config = self.get_defn()
+        self._state["region"] = config['region']
+        self.connect()
+        # handle vpcs that are deleted from outside nixops e.g console
+        if self._state.get('vpcId', None):
+            if check:
```
edolstra: I wonder if should get rid of the `check` parameter, given that nowadays we have a separate `MachineState._check()` method. That is, `nixops deploy --check` could be implemented by first calling `_check()` on each resource to verify each resource and update its state to match its actual condition, and then calling `create` as usual (but without a `check` flag).

There is a FIXME in `MachineState._check()` that it should be moved to `ResourceState` BTW.

nixops/diff.py
```diff
@@ -0,0 +1,154 @@
...
+                          'name', 'creationTime']
+
+    def set_reserved_keys(self, keys):
+        self._reserved.extend(keys)
+
+    def get_keys(self):
+        diff = [k for k in self._diff.keys() if k not in self._reserved]
+        return diff
+
+    def plan(self,show=False):
```
edolstra: Maybe add some comments describing what this does?

examples/vpc.nix
```diff
@@ -0,0 +1,126 @@
...
+{
+  
+  machine =
+    {config, resources, pkgs, lib, ...}:
+    {
+      deployment.targetEnv = "ec2";
+      deployment.ec2.accessKeyId = accessKeyId;
+      deployment.ec2.associatePublicIpAddress = true;
+      deployment.ec2.region = region;
+      deployment.ec2.instanceType = "r3.xlarge";
```
edolstra: Maybe use a less beefy instance type, for those of us who start this example and then forgot they were running it :-)

nix/aws-vpn-connection.nix
```diff
@@ -0,0 +1,53 @@
...
+    
+    accessKeyId = mkOption {
+      type = types.str;
+      description = "The AWS Access Key ID.";
+    };
+
+    region = mkOption {
+      type = types.str;
+      description = "AWS region.";
+    };
```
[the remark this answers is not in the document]
AmineChikhaoui: @edolstra yep I agree

nixops/resources/aws_vpn_connection.py
AmineChikhaoui: @edolstra hm weird, thought It was fixed. I'll double check.

nixops/resources/aws_vpn_connection.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
+        diff_engine = self.setup_diff_engine(config=defn.config)
+
+        self.access_key_id = defn.config['accessKeyId'] or nixops.ec2_utils.get_access_key_id()
+        if not self.access_key_id:
+            raise Exception("please set 'accessKeyId', $EC2_ACCESS_KEY or $AWS_ACCESS_KEY_ID")
+
+        for handler in diff_engine.plan():
+            handler.handle(allow_recreate)
```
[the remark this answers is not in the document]
AmineChikhaoui: :+1: makes sense

nixops/resources/vpc.py
AmineChikhaoui: @edolstra Couldn't agree more. I was even thinking at some point that we should add `nixops refresh` as a separate command which takes care of updating the state. Didn't know about _check method though, so will definitely use it.

nixops/resources/vpc.py
AmineChikhaoui: Yeah that would be ideal. Still thinking what's the best way to do that but having a global DAG will make things much better.

nixops/resources/ec2_common.py
AmineChikhaoui: @edolstra oops I think I can remove that.

nixops/resources/aws_vpn_connection.py
edolstra: Well, that was last week, so maybe it works now...

nixops/resources/ec2_common.py
AmineChikhaoui: @edolstra actually since we have `tags = nixops.util.attr_property("ec2.tags", {}, 'json')` in `nixops.ec2_common.EC2CommonState` I need to keep this for now as I'm inheriting from that class.

examples/vpc.nix
```diff
@@ -0,0 +1,126 @@
...
+  resources.ec2KeyPairs.keypair = { inherit region accessKeyId; };
+
+  resources.ec2SecurityGroups = {
+    sg =
+      { resources, lib, ... }:
+      {
+        inherit region accessKeyId;
+        vpcId = resources.vpc.vpc-nixops;
+        rules = [
+          { toPort = 22; fromPort = 22; sourceIp = "41.231.120.171/32"; }
```
Mic92: Please add a check if `fromPort` is set otherwise the following error message is thrown: 

```
AttributeError: 'NoneType' object has no attribute 'get'
```

from


```
from_port = int(rule_xml.find("attr[@name='fromPort']/int").get("value"))
```

Mic92:
Can you add an example how to add an elastic ip to a machine in a VPC subnet?

Mic92:
Removing ip addresses also does not work yet.

AmineChikhaoui:
@Mic92 can you share more details on the issue you had ? is this during a machine destroy or elastic ip destroy ?

AmineChikhaoui:
@domenkozar @rbvermaa I'm ok to merge this if there are no objections. I tried to address some of Eelco's comments and I'll create few issues for the remaining improvements to work on them later.
But I think globally it would be nice if we can merge this so that we can use VPC resources and use the diff engine for any new resource as it make it much easier to handle the state.
Let me know if you have any comments :)

Mic92:
I have a machine, that I only start for benchmarks. As I have not found a way to remove an elastic ip, when the machine was stopped, I have removed it manually in the web interface. To assign a new ip I created a new resource in nixops each time. Now I have a number of obsolete `nix-ip$i` resources, I cannot get rid off:

```
$ nixops deploy -d cntr --kill-obsolete --show-trace --debug
resource ‘cntr-ip3’ is obsolete
resource ‘cntr-ip2’ is obsolete
resource ‘cntr-ip4’ is obsolete
resource ‘cntr-ip’ is obsolete
resource ‘cntr-ip3’ is obsolete
resource ‘cntr-ip2’ is obsolete
resource ‘cntr-ip4’ is obsolete
resource ‘cntr-ip’ is obsolete
error: Multiple exceptions: list index out of range, list index out of range, list index out of range, list index out of range
------------------------------
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/parallel.py", line 41, in thread_fun
    result_queue.put((worker_fun(t), None))
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/deployment.py", line 1066, in worker
    if m.destroy(wipe=wipe): self.delete_resource(m)
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 109, in destroy
    eip = self.describe_eip()
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 103, in describe_eip
    return response['Addresses'][0]
IndexError: list index out of range
------------------------------
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/parallel.py", line 41, in thread_fun
    result_queue.put((worker_fun(t), None))
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/deployment.py", line 1066, in worker
    if m.destroy(wipe=wipe): self.delete_resource(m)
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 109, in destroy
    eip = self.describe_eip()
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 103, in describe_eip
    return response['Addresses'][0]
IndexError: list index out of range
------------------------------
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/parallel.py", line 41, in thread_fun
    result_queue.put((worker_fun(t), None))
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/deployment.py", line 1066, in worker
    if m.destroy(wipe=wipe): self.delete_resource(m)
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 109, in destroy
    eip = self.describe_eip()
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 103, in describe_eip
    return response['Addresses'][0]
IndexError: list index out of range
------------------------------
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/parallel.py", line 41, in thread_fun
    result_queue.put((worker_fun(t), None))
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/deployment.py", line 1066, in worker
    if m.destroy(wipe=wipe): self.delete_resource(m)
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 109, in destroy
    eip = self.describe_eip()
  File "/home/joerg/.homesick/repos/dotfiles/nixos/vms/nixops/nixops/resources/elastic_ip.py", line 103, in describe_eip
    return response['Addresses'][0]
IndexError: list index out of range
```

AmineChikhaoui:
@Mic92 thanks should be fixed in https://github.com/NixOS/nixops/pull/697/commits/f6de2a9daede81af807ddf014ad3a0a0bd08dd70.

Mic92:
@AmineChikhaoui thanks. that solved the issue.

AmineChikhaoui:
@domenkozar @rbvermaa Would you mind looking at this again. I've added few basic tests to `tests/functional/vpc.py`. Let me know if there are any specific tests you think should be added.

AmineChikhaoui:
yay 🎆 , thanks @edolstra

PR #699, opened 2017-07-12
doc: libvirtd: use group write bits on /var/lib/libvirt/images

bjornfor:
The current example chowns the images directory to "myuser:libvirtd" and
then uses the ability of the _user_ to write to the image directory. I
think it's better to have the directory owned by "root:libvirtd" and
rely on on the _group_ to be able to write to the images directory. That
way, more than one user is allowed to manage the deployment.

Moreover, perhaps NixOS can configure these permission bits by itself
and this manual step can be removed in the future.


bjornfor:
Thanks!

PR #702, opened 2017-07-14
libvirtd: create "persistent" instead of "transient" domains

bjornfor:
This ensures that nixops deployments/VMs survive a reboot(!).

Persistent domains/VMs are created by "virsh define domain.xml" whereas
transient domains are created (and started) with "virsh create domain.xml".

See
https://wiki.libvirt.org/page/VM_lifecycle#Transient_guest_domains_vs_Persistent_guest_domains

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -106,6 +106,11 @@ class LibvirtdState(MachineState):
             # TODO: use libvirtd.extraConfig to make the image accessible for your user
             os.chmod(self.disk_path, 0666)
             self.vm_id = self._vm_id()
+            dom_file = self.depl.tempdir + "/{0}-domain.xml".format(self.name)
+            nixops.util.write_file(dom_file, self.domain_xml)
+            # By using "virsh define" we ensure that the domain is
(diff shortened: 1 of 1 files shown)


Nadrieril:
:+1: for the general idea. However, I believe that `self.depl.tempdir` is meant to be a temporary directory (such as `/tmp/...`) which therefore may not survive a reboot.
To make images properly persistent you would need a persistent directory to store image definition files too.

bjornfor:
@Nadrieril: ```virsh define /tmp/domain.xml``` copies the xml file to ```/var/lib/libvirt/qemu/domain.xml```. So it does survive a reboot. I tested it. Modifications to the nixops deployment work too. (I was afraid for a moment that it'd create a _new_ deployment everytime I make a change and run ```nixops deploy -d example-libvirtd``` (and reboot my host in between), but it doesn't.)

Nadrieril:
Oh, great :)

domenkozar:
This looks good to me - it's consistent with the rest of the backend.

bjornfor:
@domenkozar: Thanks!

PR #705, opened 2017-07-21
libvirt: undefine domain on "nixops destroy"

bjornfor:
This fixes a regression introduced in commit e4105635c18f ("libvirtd:
create "persistent" instead of "transient" domains"). Now that domains
are persistent, we must de-allocate resources with "virsh undefine".


bjornfor:
Thanks!

Issue #261, opened 2015-01-31
Cannot (re)deploy old deployment (missing "backing file")


bjornfor:
I created a nixops deployment some time ago (~2 months). Now I cannot start/deploy it anymore, because of what looks like garbage collection having removed the "backing file":

```
$ nixops deploy -d trivial
machine> error: Failed to create domain from /tmp/nixops-tmpBdXUKd/machine-domain.xml
machine> error: Cannot access backing file '/nix/store/nmg8c0b84b5mx1f24l9vqkj2fz9rmpyg-libvirtd-ssh-image/disk.qcow2' of storage file '/home/bfo/nixops-libvirt-images/nixops-60d1a626-8635-11e4-bbc6-00012e496794-machine.img' (as uid:0, gid:0): No such file or directory
machine> 
error: command ‘['virsh', '-c', 'qemu:///system', 'create', '/tmp/nixops-tmpBdXUKd/machine-domain.xml']’ failed on machine ‘machine’ (exit code 1)
```

How to rebuild the missing "backing file"? Shouldn't nixops do this automatically?

I looked at the available options for the "deploy" command, in case there was some option to rebuild the "backing file" or something. I tested "nixops deploy -d trivial --repair" (and --allow-recreate), but that didn't fix it.

"nixops start -d trivial" prints the same message (missing backing file).

I'm running nixops-1.3pre1362_2a616f8.

domenkozar:
So here's whats' going on:

- libvirt backend creates `base_image` with `nix-build` and points output link to `/tmp/`
- `"qemu-img", "create", "-f", "qcow2", "-b",  base_image + "/disk.qcow2", self.disk_path` is invoked

### current behaviour

Once you reboot, GC root is gone and base image can be garbage collected. Once you deploy again, `disk.qcow2` can't find it's base image anymore.

### proposed solution

We should really build a basic qemu image for each release and always use just that. If it's GC-ed, substitute it again. No need to build an image each time, best to just deploy to the image (same as VB does).

johan-boule:
Will the proposed solution also work with unstable channels ?
In the meantime, before a solution is found, we can transform the image into standalone one by copying all backing data into the final image file itself, with :
`qemu-img rebase -b '' /home/bfo/nixops-libvirt-images/nixops-60d1a626-8635-11e4-bbc6-00012e496794-machine.img`
After this command, the image file is self-contained and no longer linked to any backing file anymore.

Nadrieril:
[...]

PR #706, opened 2017-07-21
libvirt: stop using disk backing file

bjornfor:
Use "qemu-img rebase ..." to get rid of the backing file reference when
creating a VM. The result is a self-contained VM image in
/var/lib/libvirt/images/. This fixes nixops libvirt deployments from
breaking after the nix garbage collector removes the backing file.

Pros and cons about using disk backing file (as currently implemented):

Pros:
* Disk storage savings for the initial deployment, iff deploying 1+ VMs.

Cons:
* After a period of guest (VM) system updates and GC cycles, no (user)
  data from the backing file is needed anymore. Every bit of useful disk
  data is stored in the "overlay" image. So long term it is even worse
  for disk usage to have a backing file.
* Today, nothing prevents the backing file from being garbage collected.
  When that happens, the VMs stop working (incomplete disk).

And a data point. The nixops "virtualbox" backend uses a "clonehd"
command, which is similar to what this patch does.

Alternative fixes:
* Prevent the backing file from ever being garbage collected. That has a
  (longterm) negative effect on disk usage, as argued above.
* Have nixops prevent GC'ing the base image and also "rebase" images on
  top of updated backing files, presumably on every "deploy" command.
  That's more complicated (and expensive, according to "man qemu-img")
  but can potentially save significant amount of disk space, for large
  deployments.

Observations:
The libvirt base image is built in two steps, in two different
derivations. The 2nd derivation uses the disk image from the 1st
derivation as backing file. I kept this functionality as it saves
copying the whole disk image around. The "qemu-img rebase" operation at
VM creation is not expensive, AFAICT. I didn't notice any difference in
deployment speed (as a human without a stopwatch).

Fixes https://github.com/NixOS/nixops/issues/261

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -4,6 +4,7 @@ from distutils import spawn
 import os
 import copy
 import random
+import shutil
 import string
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


bjornfor:
Keep in mind what the baseline is: a libvirtd deployment is currently broken by nix garbage collection. Which means (IMHO) that the libvirtd backend is broken. Can there be _any_ libvirtd backend users at this point? In addition to this (blocker), not long ago (and still in the latest nixops release) is the fact that libvirtd deployments don't even survive a (host) reboot.

I'm not in a rush, so please take time to review. I just felt like getting that information out :-)

bjornfor:
Thanks!

Issue #538, opened 2016-11-01
Uploading to S3 buckets


chris-martin:
#173 discusses creating S3 buckets.

Can nixops also upload things into the buckets?

cessationoftime:
@edolstra I plan to try to build a solution for uploading to S3 buckets, but I am wondering if it is more appropriate to build it as part of the existing [S3BucketState](https://github.com/NixOS/nixops/blob/master/nixops/resources/s3_bucket.py#L34) class, or as a separate resource/class (S3BucketFile) that depends on an existing bucket (whether the bucket is NixOps managed or not).  I think the latter, but I'm not yet familiar enough with how NixOps works to be certain.

edolstra:
I'm not sure. On the one hand, fine-grained resources are nice because they can be managed (created/deleted/...) separately. On the other hand, there could be a lot of files which would potentially clutter up the UI.

cessationoftime:
I've made progress on this and I can upload files to S3 buckets. However I want to upload the output of a custom derivation.

So I have code like:
```
let
  mypkgs = import ./my-pkgs { };

  ec2Machine =
    { config, pkgs, resources, ... }:
    { deployment.targetEnv = "ec2";
...
...
    };

in
{
  ...
  ...
  resources.s3BucketUploads.frontend-upload =
    { inherit region accessKeyId;
      description = "web frontend";
      bucketName = "deploy.haste.us";
      source = "${mypkgs.frontend}/bin";
      #source = "/home/cessationoftime/workspace/frontend/dist/bin";
      bucketDestination = "frontendFolder";
    };

    web1 = ec2Machine;
}
```
When I use a source directory value of "/home/cessationoftime/workspace/frontend/dist/bin" the upload works fine, but if I use "${mypkgs.frontend}/bin" then the directory cannot be found.  I need a way to trigger the build for ${mypkgs.frontend} before performing the upload and I'm not sure how best to go about that.   I see the [build_configs function](https://github.com/NixOS/nixops/blob/7fc0377cbd7331cad8dcfb83c3adf97f724279c1/nixops/deployment.py#L572), do I need to create a similar function in my resource or is there another way to trigger the build for ${mypkgs.frontend}?

domenkozar:
I'll fix that shortly.

cessationoftime:
[...]

PR #719, opened 2017-08-19
Build machines before creating resources

domenkozar:
For a couple of reasons:

- machine nixos module could have nixos assertions that depend on resources
  and before this change it didn't trigger before resources were created
  and machine was brought up (and had attached resources like security
  groups for ec2), thus failing due to some misconfiguration

- provisioning resources might depend on built closures of machines

- such separation of concerns falls more into Nix eval/build/deploy model (full build before provisioning resources)

- if building machines failed for any reason, resources were already
  provisioned, leaving the deployed in somewhat broken state

There is one drawback I noticed:

- Remote building on the deployed machine is not supported anymore,
  but I don't think it ever could've worked since there's a chicken and
  egg problem when it comes to provisioning

Unblocks https://github.com/NixOS/nixops/issues/538

cc @AmineChikhaoui

### DIFF
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -885,79 +885,77 @@ class Deployment(object):
 
         self.logger.update_log_prefixes()
 
+        # Build the machine configurations.
+        bc = self.build_configs(dry_run=dry_run, repair=repair, include=include, exclude=exclude)
+
+        if build_only or dry_run:
+            return
+
+        # Record configs_path in the state so that the ‘info’ command
+        # can show whether machines have an outdated configuration.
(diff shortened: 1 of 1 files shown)


domenkozar:
What I'm mostly worried about here is that I missed a good reason for not doing this. I hope @edolstra and @rbvermaa can enlighten me if so :)

edolstra:
The reason is that building a machine can depend on the result of creating the resource. For example, you can refer to `${deployment.ec2.instanceId}` in a machine spec. This PR would make such specs fail to evaluate.

PR #728, opened 2017-09-20
Add domainType option to libvirt backend.

kquick:
The libvirt backend is currently hard-coded to use the kvm domain.  This is an efficient domain and good as the default, but not always available.

In my case I needed to run nixops to create a libvirt VM while running NixOS in a VMWare Fusion VM.  This patch allows:

    { example = { config, pkgs, ... }:
      { deployment.targetEnv = "libvirtd";
        deployment.libvirtd.domainType = "qemu";
       };
    }


domenkozar:
Thanks!

Issue #504, opened 2016-09-07
Python 3 Support


aneeshusa:
Does NixOps support Python 3? I couldn't tell from `setup.py` or the issues, but it looks like it's Python 2 only right now. (I'm waiting for #502 so I can try nixops out.)

I think this would be nice to go along with https://github.com/NixOS/nixpkgs/issues/18185.

domenkozar:
It doesn't support it yet.

aske:
I can try migrating NixOps to python3 (looks like we'll need to use boto3 library; need to check libraries used for non-AWS backends before that, though).

aske:
I'll make the PR to explicitly set python 2 version so we can move further with moving to python3 by default a bit later.

Alright, the rough roadmap right now looks like this:

1. [ ] Migrate to boto3
2. [ ] Upgrade/move to other libs for non-AWS backends
3. [ ] Migrate tests to nose2 (or py.test?)
4. [ ] Do the actual 2->3 migration

aske:
Also, cc @edolstra @rbvermaa I guess, in case they have any comments.

domenkozar:
@aske the plan sounds good. I'd keep nose2 for having a small scope for this somewhat big task.

domenkozar:
Boto3 migration guide: http://boto3.readthedocs.io/en/latest/guide/migration.html

aske:
I hope I'll be able to start working on migrating to boto3 this or next weekend (was a bit swamped with work).

bachp:
I don't think nixops needs to support python2 along side python3. I assume it is mostly installed vis nixpkgs which already has python3 and it can easily be switched there.

Please correct me if I'm wrong.

edolstra:
Yeah, I don't think there's any reason to keep supporting Python 2.

teto:
Any news on the python3 front ? is there a PR one can follow/help with ?

domenkozar:
First we need to migrate to boto3 as per checklist from @aske.

aske:
Hm, hackathon at NixCon looks like the perfect opportunity to finally start working on this.

domenkozar:
Small boto3 progress: https://github.com/NixOS/nixops/pull/749

davidak:
[The clock](https://pythonclock.org/) is ticking...

takeda:
[...]

PR #732, opened 2017-09-25
[RDY] Use libvirt bindings instead of parsing output

teto:
A bit of a WIP. 
When I try deploying my example, I get 
```
libvirt: Network Driver error : Network not found: no network with matching name 'default'
error: Network not found: no network with matching name 'default'
```
but my example.nix is very thin.

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -14,6 +14,11 @@ from nixops.backends import MachineDefinition, MachineState
 import nixops.known_hosts
 import nixops.util
 
+# https://www.redhat.com/archives/libvirt-users/2017-August/msg00011.html
+# def libvirt_callback(userdata, err):
+#     pass
+
+# libvirt.registerErrorHandler(f=libvirt_callback, ctx=None)
 
 class LibvirtdDefinition(MachineDefinition):
     """Definition of a trivial machine."""
@@ -59,11 +64,27 @@ class LibvirtdState(MachineState):
     def __init__(self, depl, name, id):
         MachineState.__init__(self, depl, name, id)
 
-        self.conn = libvirt.open(None)
-        if self.conn == None:
+        self.conn = libvirt.open('qemu:///system')
+        if self.conn is None:
             self.log('Failed to open connection to the hypervisor')
             sys.exit(1)
-        self.dom = self.conn.lookupByName(self.vm_id)
+        # print("VM id=", self._vm_id())
+        # self.dom = self.conn.lookupByName(self._vm_id())
+        self._dom = None
+        # try:
+        #     self.dom = self.conn.lookupByName(self._vm_id())
+        # except Exception as e:
+        #     self.log("Warning: %s" % e)
+
+    @property
+    def dom(self):
+
+        if self._dom is None:
+            try:
+                self._dom = self.conn.lookupByName(self._vm_id())
+            except Exception as e:
+                self.log("Warning: %s" % e)
+        return self._dom
 
     def get_ssh_private_key_file(self):
         return self._ssh_private_key_file or self.write_ssh_private_key(self.client_private_key)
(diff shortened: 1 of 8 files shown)


teto:
the 'default' network in libvirtd is visible only by root. How nixops can use the network in such conditions ? Is it supposed to ? Genuine question, due to bugs, I never reached the functional stage with libvirtd and nixops :(

teto:
I've been trying to install the changes with an overlay aka `nix-env -iA nixops-local ` but it fails when building the wheel:
```
/tmp/nix-build-nixops-1.6.2.drv-0/nixops/dist /tmp/nix-build-nixops-1.6.2.drv-0/nixops
nixops-_version_-py2-none-any.whl is not a valid wheel filename
```
It seems like the       `version='@version@',` is not substituted in my case. Is there anything special to do ?

Here is my overlay
```
   nixops-local = super.nixopsUnstable.overrideAttrs( oldAttrs: rec {

     # version = "2017-09-24";
     version = "1.6.2";
     name = "nixops-${version}";
     src = super.lib.cleanSource ~/nixops;
   });
```

teto:
So release.nix generates a tarball that is used in the derivation. It seems not straightforward to use the overlay so I stayed with ./dev-shell and succeeded making the branch define and run the vm. I will do some more testing as I need some specific libvirt tools (like possibly running my own kernel/initramfs)

teto:
This is getting better (aka work better).
I would like to convert some parts for python 3. Is there any work in progress for python ? or is it ok as long as it remains python2 compatible ?

bjornfor:
About python 3 support: https://github.com/NixOS/nixops/issues/504

teto:
I've removed unrelated changes and cleaned the code so it would be cool to have some other feedback. I've been using this without problems.

nixops/backends/libvirtd.py
```diff
@@ -181,22 +202,19 @@ def iface(n):
...
-            ip_with_subnet = lines.split()[i + 2]
-            return ip_with_subnet.split('/')[0]
+        """
+        return ip
+        """
+        # alternative is VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE
+        ifaces = self.dom.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE, 0)
+        if (ifaces == None):
+            self.log("Failed to get domain interfaces")
+            sys.exit(0)
```
bjornfor: Shouldn't this be exit(1) or something? Also, the current code doesn't call exit() at all. I think adding a comment (in the commit message) about why adding a process exit point in this function is a good thing. (I.e. is it a bug in the current code that it does _not_ call exit?)

teto: that's an oversight from my side, I wanted to put it back to "None" as in the original code and forgot it.

nixops/backends/libvirtd.py
```diff
@@ -222,18 +243,19 @@ def start(self):
             self.private_ipv4 = self._parse_ip()
         else:
             self.log("starting...")
-            self._logged_exec(["virsh", "-c", "qemu:///system", "start", self.vm_id])
+            self.dom.create()
```
bjornfor: Oh, not "start"? Weird API.

bjornfor:
I don't have rights to merge, but this looks good to me. (But see comments.)

bjornfor:
Please squash commits.

teto:
I think the squashing can be done on the github side ? if not I can do it ofc

nixops/backends/libvirtd.py
teto: yes it's weird https://www.redhat.com/archives/libvir-list/2010-June/msg00140.html

domenkozar:
@teto is this ready for review?

teto:
@domenkozar it's ready, it depends on https://github.com/NixOS/nixpkgs/pull/30950#issuecomment-342510936 though.

teto:
since I,ve rebased on master I can't destroy domains anymore. That's weird, I might try against a virtualbox to see if it's nixops or this Pr problem

teto:
seems unrelated in fact, if a deploiyment is not deployed yet, apparently I should use delete instead of destroy. I haven't changed the provider behavior so this must be a generic nixops heavior, unrelated to the libvirt provider.

teto:
I added a commit to check some returned values.

teto:
Anything I can do to help with merge ? I've some other libvirt improvements in queue :)

bjornfor:
It seems the fixup commit hasn't been squashed yet.

teto:
I mistakenly deleted the remote branch so I am forced to open a new PR https://github.com/NixOS/nixops/pull/783 which is this one rebased & squashed.

PR #630, opened 2017-03-20
Route53: Support private IP address for use in private clouds

roberth:

Issue #527, opened 2016-10-19
Make route53 independent of ec2


grahamc:
I (naively) expected to be able to use route53 on my other systems. Example:

``` nix
{
  nix-dev = { resources, pkgs, lib, ... }: {
    deployment.targetEnv = "hetzner";
    deployment.route53.accessKeyId = "foo";
    deployment.route53.hostName = "foo.my-zone";
  };
}
```

I expected foo.my-zone to be created in my route53 hosted zone for this hetzner machine.

roberth:
Maybe we can change the title to 'Make route53 independent of ec2'?

grahamc:
Done, @robertth.

domenkozar:
Is anyone looking into this? We (https://github.com/input-output-hk) would like to look into round-robin DNS routing using multi-value.

domenkozar:
Ok then :)

domenkozar:
This was implemented in https://github.com/NixOS/nixops/commit/97c0f9f696f55429ef7f98e106286750cc77e228

But it's not available in the manual yet (recordset, hosted zone, healthcheck).

domenkozar:
Added to the manual in 5ecfde3ece94b5163b5c1ef949bca92826fd1133

Issue #174, opened 2014-04-01
Route53 options should allow for multiple DNS entries for a single machine


ocharles:
We would like to use nginx with a few different virtual hosts on a single EC2 instance, but we have to manage these DNS entries outside of NixOps. It would be better if we could simply provide a list of DNS entries for a machine.

domenkozar:
Relevant to https://github.com/NixOS/nixops/issues/527

ocharles:
I think this has been fixed by https://github.com/NixOS/nixops/pull/832

PR #753, opened 2017-10-30
Add route53 record set resource

moretea:


lo1tuma:
This implements the suggested approach of a route53 RecordSet resource in #469, see also #630 and #527.
It doesn’t implement all route53 features yet but it would be very easy to add additional options in order to fix e.g. #174.
Using the route53 resource instead of the existing route53 functionality should also properly cleanup the DNS entries on destroy, see #147.

--------

A question I want raise is if it would make sense to document the deprecation of the old route53 functionality, so that it can be removed sometime in the future (next major release)?

nix/route53-recordset.nix
```diff
@@ -0,0 +1,87 @@
...
+    name = mkOption {
+      default = "nixops-${uuid}-${name}";
+      type = types.str;
+      description = "Name of the recordset.";
+    };
+
+    region = mkOption {
+      type = types.str;
+      description = "AWS region.";
+    };
```
lo1tuma: This option is currently not used as it is only required for latency-based resource record sets.

rbvermaa: @lo1tuma If the option is not used, should it be removed?

moretea: We've not implemented all potential behaviours for this resource. Just the ones they need at HolidayCheck. AFAICR other features do need to have a region specified.

moretea:
I don't think that I will spend any time on this PR any time soonish, so feel free to change it any way you want ;)

nix/route53-recordset.nix
lo1tuma: As @moretea said, for the current feature set it is not needed. There are some route53 features that require a region. For now I think we can just remove it or at least make it optional.

lo1tuma:
I’ve added another [comment to #469](https://github.com/NixOS/nixops/pull/469#issuecomment-340769909) because I think having route53 as a separate resource cannot solve the use-case I have.
So I guess we need some more thinking and discussions in how nixops should handle route53.

rbvermaa:
I am going to work on route53 support the next weeks, and will likely include your commit in the work and improve if necessary.

moretea:
Awesome :)

rbvermaa:
@lo1tuma The work I am doing should include your use-case btw. The most likely solution will be allowing machine resources to be used in the recordValue, with the option of using either public or private IP address of the instance/machine.

rbvermaa:
Closing this PR. I have used the commit in the route53 branch which should be merged soon.

PR #759, opened 2017-11-01
Support deployment keys in the container backend

johbo:
We did try to use the deployment keys in the container backend and found out that the following tweaks were needed.

With the change around `run_command` I am not really sure. The problem could also be worked around by changing the order of the shell commands like the following example shows:

```
johannes@dev-johbo: ~/wo/nixops work-johbo
$ a=b echo test; (echo test)
test
test

johannes@dev-johbo: ~/wo/nixops keys-container-backend
$ a=b (echo test); echo test
-bash: syntax error near unexpected token `('
```

The shell commands are constructed in the function `send_keys`: https://github.com/NixOS/nixops/blob/master/nixops/backends/__init__.py#L242

Depending on the order does feel wrong as well though. I am not sure if we really need the `HOME=/root` fragment in `run_command` for the container backend.

### DIFF
--- a/nixops/backends/container.py
+++ b/nixops/backends/container.py
@@ -120,6 +120,8 @@ class ContainerState(MachineState):
     def create(self, defn, check, allow_reboot, allow_recreate):
         assert isinstance(defn, ContainerDefinition)
 
+        self.set_common_state(defn)
+
         if not self.client_private_key:
             (self.client_private_key, self.client_public_key) = nixops.util.create_key_pair()
 
--- a/nixops/backends/container.py
+++ b/nixops/backends/container.py
@@ -186,6 +186,8 @@ class ContainerState(MachineState):
         self.log("starting container...")
         self.host_ssh.run_command("nixos-container start {0}".format(self.vm_id))
         self.state = self.STARTING
+        self.wait_for_ssh()
+        self.send_keys()
 
     def _check(self, res):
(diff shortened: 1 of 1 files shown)


nixops/backends/container.py
```diff
@@ -104,7 +104,7 @@ def wait_for_ssh(self, check=False):
     def run_command(self, command, **kwargs):
         command = command.replace("'", r"'\''")
         return self.host_ssh.run_command(
-            "nixos-container run {0} -- bash --login -c 'HOME=/root {1}'".format(self.vm_id, command),
+            "nixos-container run {0} -- bash --login -c 'HOME=/root; {1}'".format(self.vm_id, command),
```
kevincox: Should it be `export HOME=...` if it is its own statement?

domenkozar:
@johbo now that https://github.com/NixOS/nixops/pull/804 is merged, can you retest?

johbo:
Checked, #804 does the trick. There is a pending bit around the usage of `nixops start` when trying to start a container which needs deployment keys. This does seem to have a bigger issue though, since it typically would deadlock: `nixos-container start` waits until the container is up, which will never happen if the deployment keys are not yet in the container, so calling `send_keys` would have to be done in parallel to sneak the keys in so that the container can actually finish the startup...

For this PR we are done due to #804. I'll close this one.

PR #763, opened 2017-11-03
Upload key contents atomically. Fixes #762

### LINKED ISSUE #762: Race condition: nixops key service started before keys are uploaded
Keys are uploaded in a non-atomic fashion.

(I found this in my glusterfs setup where startup sometimes fails, complaining that the the SSL privkey used doesn't have a proper header; most likely it's empty, and a few milliseconds after that, it actually gets its contents.)

The `...-key` `inotify`-based service exits as soon as the file exists, doesn't wait until the contents are up. See [code here](https://github.com/NixOS/nixops/blob/279a92a5f4ea8b1d5b354addb30b3d2f134b8968/nix/keys.nix#L252-L260).

http://patchlog.com/patches/atomic-scp-and-sftp-upload/ says

> There's no way of knowing when the files have been fully transferred, both sftp and scp would create the files as soon as the transfer begins and will close then when it finishes

[`man scp`](https://linux.die.net/man/1/scp) confirms:

> If the target file does not yet exist, an empty file with the target file name is created, then filled with the source file contents. No attempt is made at "near-atomic" transfer using temporary files.

nh2:
See #762

### DIFF
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -250,7 +250,7 @@ in
           path = [ pkgs.inotifyTools ];
           preStart = ''
             (while read f; do if [ "$f" = "${name}" ]; then break; fi; done \
-              < <(inotifywait -qm --format '%f' -e create ${keyCfg.destDir}) ) &
+              < <(inotifywait -qm --format '%f' -e create,move ${keyCfg.destDir}) ) &
 
             if [[ -e "${keyCfg.path}" ]]; then
               echo 'flapped down'
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -232,9 +232,15 @@ class MachineState(nixops.resources.ResourceState):
                 raise Exception("Neither 'text' or 'keyFile' options were set for key '{0}'.".format(k))
 
             outfile = destDir + "/" + k
+            # We scp to a temporary file and then mv because scp is not atomic.
+            # See https://github.com/NixOS/nixops/issues/762
+            tmp_outfile = destDir + "/." + k + ".tmp"
             outfile_esc = "'" + outfile.replace("'", r"'\''") + "'"
-            self.run_command("rm -f " + outfile_esc)
-            self.upload_file(tmp, outfile)
+            tmp_outfile_esc = "'" + tmp_outfile.replace("'", r"'\''") + "'"
+            self.run_command("rm -f " + outfile_esc + " " + tmp_outfile_esc)
+            self.upload_file(tmp, tmp_outfile)
+            # For permissions we use the temporary file as well, so that
+            # the final outfile will appear atomically with the right permissions.
             self.run_command(
               ' '.join([
                 # chown only if user and group exist,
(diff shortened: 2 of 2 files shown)


nh2:
8 years later, looks like this is actually still racy:

In the code

```sh
(while read f; do if [ "$f" = "${name}" ]; then break; fi; done \
  < <(inotifywait -qm --format '%f' -e create,move ${keyCfg.destDir}) ) &

if [[ -e "${keyCfg.path}" ]]; then
  echo 'flapped down'
  kill %1
  exit 0
fi
wait %1
```

I believe there is a race condition:

`(while read ... < <$(inotifywait ...) ) &` starts a subshell as a background thread, in which a command substitution is run.

The following timline showcases a race:

* The key file `name` may exist in `/run/keys` or not (doesn't matter).
* The `*-key.service` unit may be running or not (doesn't matter).
* Nixops runs [`rm -f /run/keys/name`](https://github.com/NixOS/nixops/commit/918acdacc07d58ff44faa2fb8f30badc109d8216#diff-879c67fc09be9606286c85ea406a4ebab4e68b8877dd22230de5311354d9ef3dR240-R243). This deletes the file and thus stops the `*-key.service` unit (because that's waiting on `inotifywait -qq -e delete_self /run/keys/name`).
* Now some service already running on the system, e.g. `nginx`, restarts; if it depends on `name-key.service`, the restarting of `nginx.service` will also start `name-key.service` to wait for that key.
* So now the key file is gone and `*-key.service` is starting its `ExecStartPre`, which runs the above-quoted script.
* The subshell is started, but `inotifywait` is slow to start (say 10 seconds because the disk is slow).
* The main thread runs `if [[ -e "${keyCfg.path}" ]]; then` which is false, so the main thread continues blocking forever on the subshell (`wait %1`).
* Now NixOps runs [`self.upload_file(tmp, tmp_outfile)` and `self.run_command("mv " + tmp_outfile_esc + " " + outfile_esc)`](https://github.com/NixOS/nixops/commit/918acdacc07d58ff44faa2fb8f30badc109d8216#diff-879c67fc09be9606286c85ea406a4ebab4e68b8877dd22230de5311354d9ef3dR241-R263).
* Now `/run/keys/name` exists on the server.
* Now `inotifywait` starts up, and blocks forever on `-e create,move`. Infinite deadlock: `nixops deploy` hangs forever.

Potential solution:

Instead of `inotifywait -qm --format '%f' -e create,move ${keyCfg.destDir}`, run:

```
while true; do inotifywait -qm --format '%f' -e create,move --timeout 3 .; if [[ -e testfile ]]; then echo "flapped down" >&2; exit 0; fi; done
```

The `--timeout 3` allows us to check the existence of the file periodically, so that if it was created just before `inotifywait` got started, we can still notice it with some reasonably short delay.

## Repro tooling

`slow-inotifywait.sh`

```sh
#!/usr/bin/env bash
set -eu -o pipefail
echo "slow inotifywait starting" >&2 && sleep 5 && echo "starting now" >&2 && inotifywait "$@"
```

Using this instead of `inotifywait` allows to repro this situation manually.

## Side remark: Orphan `inotifywait` process

The script as-is makes `inotifywait` an orphan process:

When the `break` terminates the subshell, `inotifywait` keeps running until it notices another file change (may never happen), at which point it would output the file name to stderr, and terminate with `Broken pipe`. Because the subshell and the main thread exit, it the orphan `inotifywait` reparented to `init` (`systemd`).

systemd doesn't kill it despite the default [`KillMode=control-group`](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=)` because this orphan gets created in `ExecStartPre=`, after which `ExecStart=` runs (which waits until the key file gets deleted, potentially never), and thus in general in systemd orphans from `ExecStartPre` leak into the runtime of `ExecStart` and are never cleaned up until the whole unit get stopped (when `ExecStart` terminates).

This is ugly and confusing (because you have leftover `inotifywait` processes) but should not affect the race.

Issue #742, opened 2017-10-19
DigitalOcean deployments break on upgrade to unstable channel


dhess:
On all of my DigitalOcean deployments, upon upgrading to the unstable channel (and possibly earlier releases), the kernel's primary NIC name changes from `enp0s3` to `ens3`.

Because DO does not support DHCP, this means that upon the first reboot after upgrading to the unstable channel, the host loses network connectivity, because the deployed configuration tries to statically configure the no-longer-existing `enp0s3` interface. 

The only fix I've been able to come up with is to login to the droplet via the DO web console and bring the interface up by hand. Then, from the deployment host, add the new `networking` config to the machine definition, deploy (which breaks as soon as the new config is activated and the connection is terminated), reboot the droplet, and deploy again to finish the job.

pbogdan:
For reference this is mentioned in [release notes of 17.09](https://nixos.org/nixos/manual/release-notes.html#sec-release-17.09-incompatibilities) though I'm still not sure how to easily deal with it.

For brand new deployments I bumped `NIX_CHANNEL` in `nixops/data/nixos-infect` to 17.09 and ran `sed -i "s/enp0/en/g"` on `nixops/data/nixos-infect` & `nixops/backends/digital_ocean.py`.

dhess:
Thanks. It would be useful if this were mentioned in the DigitalOcean NixOps manual. 

Also, I'm not sure if NixOps releases track NixOS releases in any way, but it might be a good idea to change NixOps's bundled `nixos-infect` to use 17.09 going forward. It seems reasonable to assume that anyone using a recent NixOps version and making new DigitalOcean deployments will also be deploying 17.09 or later.

cleverca22:
if you switch over to this module, then nixos can configure the right IP on bootup every time
https://github.com/NixOS/nixops/issues/605

corpix:
Hello. NixOS newcomer is here! :)
Provisioned a DigitalOcean machine with `NixOps 1.5.2`, no network after reboot.
So I think the title could be changed from

> DigitalOcean deployments break on upgrade to unstable channel

to

> DigitalOcean deployments break after reboot

PR #765, opened 2017-11-06
digital ocean: enp0s3 -> ens3

disassembler:
This addresses #742 although would break anything with older kernels with the enp0s3 device.

### DIFF
--- a/nixops/backends/digital_ocean.py
+++ b/nixops/backends/digital_ocean.py
@@ -8,7 +8,7 @@ the infect itself, another after we pushed the nixos image.
 
 I hit a few subtle problems along the way:
 * DO doesn't do dhcp so we have to hard-code the network configuration
-* Ubuntu still uses eth0, 1 etc, not enp0s3 etc so we have a network
+* Ubuntu still uses eth0, 1 etc, not ens3 etc so we have a network
   link name change after the reboot.
 * I had to modify nixos-infect to reflect the network link name changes,
   and to not reboot to avoid ssh-interruption and therefore errors.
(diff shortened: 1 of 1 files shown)


Mic92:
could this be checked?

rbvermaa:
Yeah, we should be able to set the name based on the NixOS version (system.nixosVersion), right?

Mic92:
From my understanding this happens in ubuntu and not nixos? So checking `nixosVersion` should be not the correct solution.

domenkozar:
It happens on digitalocean running nixos.

coretemp:
Hard coding such constants seems to be a flawed approach requiring a ton of maintenance(they change scheme every few years). I have little doubt that there is a way to do these things dynamically and in a way that is only kernel dependent (e.g. Linux or FreeBSD) and not on a particular version. 

Perhaps it could be done in a platform independent way by using some cross-platform tools. 

Another thing I don't really like is that a PR is opened in which the word "breaks" occurs. OTOH, perhaps we should be happy that the disassembler took the time to improve something to the DigitalOcean user community. I suppose what it comes down to is whether it would constitute a net "improvement" (which is why it isn't merged yet). If breaking something means we don't merge it, we should just be honest about that and thank disassembler for the attempt and close the PR. I don't know what the correct response is, but having 5 people (including me) look at it, seems to be a waste of resources. One qualified reviewer is economically optimal according to research. While no money changes hands, there are such limits as time and attention.

I vote to close this PR, just to draw a line.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #773, opened 2017-11-16
fix parsing error in virtualbox backend

corngood:
The default virtualbox config outputs options such as:

    videocapopts=

These were failing to parse with:

    error: string index out of range


lihop:
:+1: Thanks! I encountered this issue with the trivial virtualbox example and this patch resolves it for me.

PR #786, opened 2017-11-28
Fix iam roles permissions

### LINKED ISSUE #785: AWS: update iam roles on ec2 instance
refer to [1]

currently, boto (both 2 and 3) lack API's for DescribeIamInstanceProfileAssociations/AssociateIamInstanceProfile/DisassociateIamInstanceProfile

so there is no way to modify the iam roles attached to an instance after deploy

once boto has been updated, the code in nixops to use it can be written

[1] https://github.com/NixOS/nixops/pull/786/commits/a52d957e82d89640a4c77b382280c3bf62852f4c

cleverca22:
this PR adds proper error messages to the IAM permission errors, rather then silently assuming things dont exist
it also documents how to solve #785 
the AWS policy that gives the minimum required to manage IAM roles is in [1]

[1] https://gist.github.com/cleverca22/4d4a846bcca3a406e03d8b29926b057d

### DIFF
--- a/nixops/backends/ec2.py
+++ b/nixops/backends/ec2.py
@@ -919,6 +919,7 @@ class EC2State(MachineState, nixops.resources.ec2_common.EC2CommonState):
             # instance has been provisioned.
             self._cancel_spot_request()
         else: # not self.vm_id
+            pass
             #if defn.instance_profile.startswith("arn:") :
             #common_args['instance_profile_arn'] = defn.instance_profile
             #else:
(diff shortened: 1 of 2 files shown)


nixops/backends/ec2.py
```diff
@@ -918,6 +918,17 @@ def create(self, defn, check, allow_reboot, allow_recreate):
             # Cancel spot instance request, it isn't needed after the
             # instance has been provisioned.
             self._cancel_spot_request()
+        else: # not self.vm_id
+            #if defn.instance_profile.startswith("arn:") :
```
copumpkin: Doesn't this need a `pass`?

cleverca22:
@copumpkin ping

copumpkin:
Looks good to me, but I don't use NixOps so don't put too much weight on that. I do deal with a lot of AWS IAM though 😄

PR #799, opened 2017-12-07
Replace use of select() by poll(). Fixes #798.

### LINKED ISSUE #798: nixops uses select()
Nixops currently [uses](https://github.com/NixOS/nixops/blob/c5a0e288b0c671a959e1488b94ed3c6b5f4fbdf4/nixops/util.py#L98) the `select()` syscall, which will fail when any fd given to it is a number bigger than 1023.

I suspect this can lead to random failures, as there's no code in nixops that guarantees that at no given time less than 1024 files are opened.

The simplest fix is probably to use [`poll()`](https://docs.python.org/3/library/select.html#poll-objects) instead.

nh2:
`select()` only works with FD numbers <= 1023.

Since we have no explicit control over FD numbers,
`select()` can fail when there are many files open.

This commit uses `poll()` instead, which doesn't suffer
from this restriction.

`poll()` doesn't work on Windows, but that's not
a supported platform for Nixops anyway.

### DIFF
--- a/nixops/util.py
+++ b/nixops/util.py
@@ -43,6 +43,34 @@ class CommandFailed(Exception):
         return "{0} (exit code {1})".format(self.message, self.exitcode)
 
 
+def poll_file_like_objects(objs, timeout=None, eventmask=None):
+    """
+    Like `select.select()`, but implemented using `poll()`.
+    Given `objs` must be a list of file-like objects that with
+    a `fileno()` method.
+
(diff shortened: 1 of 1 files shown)


grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #804, opened 2017-12-08
Fix #803 containers keys upload

### LINKED ISSUE #803: nixops send-keys does nothing when deploying to NixOS containers
Container deployments that wait for `...-key.service` hang forever, because the keys are never uploaded.

`nixops send-keys` does nothing for container deployments.

On first look, this is because there's this:

https://github.com/NixOS/nixops/blob/c5a0e288b0c671a959e1488b94ed3c6b5f4fbdf4/nixops/backends/__init__.py#L48

`storeKeysOnMachine` defaulting to `True` even though the manual release notes say for nixops 1.2 ["Changed default value of `deployment.storeKeysOnMachine` to false"](https://github.com/NixOS/nixops/blob/c5a0e288b0c671a959e1488b94ed3c6b5f4fbdf4/doc/manual/release-notes.xml#L463).

However, even after setting this to `True`, `nixops send-keys` does nothing, because `get_keys()` [here](https://github.com/NixOS/nixops/blob/c5a0e288b0c671a959e1488b94ed3c6b5f4fbdf4/nixops/backends/__init__.py#L215) returns `{}` for unknown reason. I can see the keys being set correctly in the `MachineDefinition` [here](https://github.com/NixOS/nixops/blob/c5a0e288b0c671a959e1488b94ed3c6b5f4fbdf4/nixops/backends/__init__.py#L36) if I print `self.keys` there.

nh2:
This fixes the multiple issues I've found in #803 about stuff being broken in `containers.py`.

See commit messages for details.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -45,7 +45,7 @@ class MachineState(nixops.resources.ResourceState):
     ssh_pinged = nixops.util.attr_property("sshPinged", False, bool)
     ssh_port = nixops.util.attr_property("targetPort", 22, int)
     public_vpn_key = nixops.util.attr_property("publicVpnKey", None)
-    store_keys_on_machine = nixops.util.attr_property("storeKeysOnMachine", True, bool)
+    store_keys_on_machine = nixops.util.attr_property("storeKeysOnMachine", False, bool)
     keys = nixops.util.attr_property("keys", {}, 'json')
     owners = nixops.util.attr_property("owners", [], 'json')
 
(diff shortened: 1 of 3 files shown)


domenkozar:
@nh2 https://github.com/NixOS/nixops/pull/759 also calls `send_keys()`, I'd assume that's also needed.

kevincox:
Overall looks good. Pinging @johbo about the `send_keys()` call.

johbo:
Let's merge this, we can add send_keys() also with another PR. This chunk looks good to me. 

@mbld ping

nh2:
> #759 also calls `send_keys()`, I'd assume that's also needed.

@domenkozar @johbo @kevincox Hmm, for me it seems to have worked without an extra call to `send_keys()`.

johbo:
If I remember correctly, then the call to `send_keys()` was only needed when using `nixops start` to start an already existing container. Not 100% sure at the moment though, will have to give it a try.

nh2:
> If I remember correctly, then the call to send_keys() was only needed when using nixops start to start an already existing container.

Ah, I see. I haven't tried `nixops start` with my fixes yet, only `deploy`.

PR #811, opened 2017-12-11
libvirtd: added kernel/initrd/cmdline support

teto:
This allows ot generate the following domain part, which is useful for
kernel testing (e.g. no need to redeploy the VM).
```
  <os>
    <type arch='x86_64' machine='pc-i440fx-2.10'>hvm</type>
    <kernel>/home/teto/mykernel/build/arch/x86_64/boot/bzImage</kernel>
    <cmdline>root=/dev/sda1 earlycon=ttyS0 console=ttyS0</cmdline>
    <boot dev='hd'/>
  </os>
```

### DIFF
--- a/nixops/backends/libvirtd.py
+++ b/nixops/backends/libvirtd.py
@@ -170,7 +170,7 @@ class LibvirtdState(MachineState):
             return [
                 '<os>',
                 '    <type arch="x86_64">hvm</type>',
-                "    <kernel>%s</kernel>" % defn.kernel or '',
+                "    <kernel>%s</kernel>" % defn.kernel,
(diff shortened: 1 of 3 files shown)


nixops/backends/libvirtd.py
```diff
@@ -163,14 +166,22 @@ def iface(n):
                 '    </interface>',
             ]).format(n)
 
+        def _make_os(defn):
+            return [
+                '<os>',
+                '    <type arch="x86_64">hvm</type>',
+                "    <kernel>%s</kernel>" % defn.kernel or '',
```
domenkozar: or is redundant here if you only allow string type in Nix

teto: Right, fixed it in a separate commit to make review easier

PR #817, opened 2017-12-18
Add RDS DB security group resource and make it possible to attach it to an rds instance

AmineChikhaoui:
Previously, rds instances are created with the default rds db security group which as far as I can see don't have any rules by default, so the rds instance resource is quite unusable atm.

This adds RDS db security groups as resources as well as an option to specify a list of security groups an RDS instance can run in.


AmineChikhaoui:
Marking this as WIP since I'll be rewriting it with the diff engine (as it was merged recently)

PR #820, opened 2017-12-20
Update nixos-infect to 5d8c2ed (changed 17.03 to 17.09)

mhsjlw:
I believe there should be a better way of doing this than syncing up with `nixos-infect` every once in a while. Especially since the Linode pull request now maintains it's own separate copy of `nixos-infect`. I'm open to suggestions

### DIFF
--- a/nixops/data/nixos-infect
+++ b/nixops/data/nixos-infect
@@ -1,53 +1,12 @@
 #! /usr/bin/env bash
 
-# Use Digital Ocean Droplet image:
-#
-#     Fedora 24 x64
-#     Ubuntu 16.04 x64
-#     Debian 8.5 x64
-#
-# YMMV with any other hoster + image combination.
+# More info at: https://github.com/elitak/nixos-infect
 
-# nixos-infect is so named because of the high likelihood of rendering a system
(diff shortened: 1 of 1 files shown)


GlennS:
Seems like a good point.

As a first step - and especially since I've made the mess worse - I could:

1. Document what I had to do to nixos-infect to make it work with Linode
2. Instead of maintaining a separate copy of the file, make it into a patch that you can apply to the Digital Ocean version

Would that be helpful?

mhsjlw:
Yup, that sounds like an awesome idea

rbvermaa:
Committed in e93ef69

GlennS:
I will use this as the basis for updating my Linode pull request.

At present my patch looks like: https://pastebin.com/g2hQMnqA

But there is still some unknown problem that I need to debug.

GlennS:
Looks like this was reverted? https://github.com/NixOS/nixops/commit/72109e508b660e7fedb33d1c908513f680937f42

mhsjlw:
> Still need to test upgrade from 17.03 bootstrapped droplet

Issue #827, opened 2017-12-30
libvirtd: image derivation fails due to mknod


emmanuelrosa:
When using Nixops 1.5.2 to deploy using the libvirtd backend the VM image fails to build due to a failure when executing `mknod`. The command produces the message *mknod: /dev/vda1: File exists*. I believe that at that point `mknod` returns an exit status, which causes the build to fail. Here's a sample deployment/network:

```
let
    manual = (import <nixpkgs/nixos/release.nix> {}).manual.x86_64-linux;
in
{
    network.description = "My web server";

    webserver = { config, pkgs, lib, nixos, ... }:
    {
        deployment = {
            targetEnv = "libvirtd";
            libvirtd.headless = true;
        };

        networking.firewall.allowedTCPPorts = [ 2015 ];

        services.caddy = {
            enable = true;
            config = ''
                0.0.0.0 {
                    root "${manual}/share/doc/nixos"
                }
            '';
        };
    };
}
```

And here's an attempt to deploy it:

```
$ nixops deploy -d mywebserver --show-trace
webserver> these derivations will be built:  
webserver>   /nix/store/mghrxvcr3b6f2yw8b15ysb15chr6wbw3-libvirtd-image.drv                                                                                                                                                                                                      
webserver>   /nix/store/cm35z2qc7vx51qpgpcg0lvsvymy9sr7j-libvirtd-ssh-image.drv
webserver> building path(s) ‘/nix/store/chzyrfnb5k84fff69q881yrv4m8wgjrr-libvirtd-image’
webserver> Formatting '/nix/store/chzyrfnb5k84fff69q881yrv4m8wgjrr-libvirtd-image/image', fmt=qcow2 size=10737418240 cluster_size=65536 lazy_refcounts=off refcount_bits=16
webserver>                                                                
SeaBIOS (version rel-1.11.0-0-g63451fca13-prebuilt.qemu-project.org)
webserver>                                                                                                                                                                                                                                                                       
webserver>                                             
webserver> iPXE (http://ipxe.org) 00:03.0 C980 PCI2.10 PnP PMM+2FF912F0+2FEF12F0 C980
                                                                    
webserver>                                                      
webserver>                                              
webserver> Booting from ROM...                                
webserver> Probing EDD (edd=off to disable)... o                    
loading kernel modules...                                             
webserver> mounting Nix store...                                    
webserver> mounting host's temporary directory...      
[...]

PR #826, opened 2017-12-29
Get rid of `mknod /dev/vda1` thanks to the introduction of `devtmpfs`

Ma27:
see NixOS/nixpkgs@0d27df2 for further reference

With this call the build using `nix-build nix/libvirtd-image.nix` failed
as the build script attempted to recreate `/dev/vda1`:

```
loading kernel modules...
node1> mounting Nix store...
node1> mounting host's temporary directory...
node1> starting stage 2 (/nix/store/r32wfcabkjn235wdr29c8d5bl89rl5hn-vm-run-stage2)
node1> Information: You may need to update /etc/fstab.
node1>
node1> Information: You may need to update /etc/fstab.
node1>
node1> mknod: /dev/vda1: File exists
node1> [    0.822831] reboot: Power down
node1> builder for '/nix/store/wds74n3rh3cfidy16rsnr7fk0hlbb2iz-libvirtd-image.drv' failed with exit code 1
node1> cannot build derivation '/nix/store/kk872nl1mldmqym0gpry0zmnpnhmwcdl-libvirtd-ssh-image.drv': 1 dependencies couldn't be built
node1> error: build of '/nix/store/kk872nl1mldmqym0gpry0zmnpnhmwcdl-libvirtd-ssh-image.drv' failed
error: command ‘['nix-build', '-I', 'nixops=/nix/store/i9j7bm77bl0ad4wqqbfqmc3w60n9zxgr-nixops-1.5.2/share/nix/nixops', '--arg', 'networkExprs', '[ "/home/ma27/Projects/node1/nixops/libvirtd.nix" "/home/ma27/Projects/node1/nixops/cluster.nix" ]', '--arg', 'args', '{}', '--argstr', 'uuid', u'376c50fa-eca7-11e7-8960-024295226d2c', '--argstr', 'deploymentName', u'node1', '<nixops/eval-machine-info.nix>', '--arg', 'checkConfigurationOptions', 'false', '-A', 'nodes.node1.config.deployment.libvirtd.baseImage', '-o', '/tmp/nixops-tmpx5uITs/libvirtd-image-node1']’ failed on machine ‘node1’ (exit code 100)
```

Since the addition of `tmpfs` to the VM builds these to commands can be
omitted.

PR #828, opened 2017-12-30
libvirtd: fix mknod error

emmanuelrosa:
This changeset modifies the libvirtd image Nix expressions to only call `mknod` when the block device doesn't already exist, therefore avoiding a build failure.

ref #827


Ma27:
see #826 
Since https://github.com/NixOS/nixpkgs/commit/0d27df2 we don't need these `mknod` statements anymore

emmanuelrosa:
Ah, I see.

So if the `mknod` statements are removed, then a `NixOps` minor release would have to be cut (ex. v1.6) so that NixOS Hummingbird can continue with v1.5.x, since it doesn't have the updated build-support/vm changes.

I created this changeset on top of the v.1.5.2 tag, thinking that it would work for Hummingbird and Impala.

domenkozar:
Fixed in https://github.com/NixOS/nixops/pull/826

PR #832, opened 2018-01-09
Route53 [WIP]

rbvermaa:
Still need to add option to allow machines to be used as part of the record values.


ocharles:
This looks fantastic! Please also don't forget to update the manual :smile:

Issue #766, opened 2017-11-09
nixops for azure broken on 17.09 and unstable


jbgi:
Azure deployment does not work anymore because of some changes in the new version of the `adal` python lib packaged for 17.09+ .

It looks access to functions `acquire_token_with_client_credentials`/`acquire_token_with_username_password` needs a different import:
`error: 'module' object has no attribute 'acquire_token_with_username_password'` at https://github.com/NixOS/nixops/blob/ab5256aec4d5cb06e3142ce011c95459736642b2/nixops/azure_common.py#L275

pvgoran:
Authenticating via service principal doesn't work as well (a similar message about the `acquire_token_with_username_password` attribute is shown).

This is a huge problem, and it wasn't addressed in any way for more than a month. Is NixOps even maintained?..

hyphon81:
It seems `adal.acquire_token_with_username_password` is wrong.
For example,
```
context = adal.AuthenticationContext(self.authority_url)
context.acquire_token_with_username_password
```
is right.
https://github.com/NixOS/nixops/blob/ab5256aec4d5cb06e3142ce011c95459736642b2/nixops/azure_common.py#L275-L278

However, acquire_token_with_username_password required the OAuth client id.
https://github.com/AzureAD/azure-activedirectory-library-for-python/blob/a73579d84983012e546c26d308d4894590c9fea5/adal/authentication_context.py#L133

And, Microsoft doesn't recommended auth with username and password,
so, I think, on NixOps codes, it needs to reconsider authentication method for Azure and nix expressions for deploying to Azure.
https://github.com/AzureAD/azure-activedirectory-library-for-python/issues/50
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-authentication-scenarios

hyphon81:
Now, if we use Azure with NixOps, we should prepare the web app server for authentication.
https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-authentication-scenarios

Issue #616, opened 2017-03-05
How to set up azure credentials?


spinus:
I have troubles to provide credentials.
Currently I took just network from examples and I'm having problem with launching it.

My username is my email (not microsoft email or anything like that), password set by env variable.
Authority URL: https://management.azure.com/, but I also tried https://login.windows.net/TENANT_ID (where tenant is properly set, this was mentioned in docs)
and subscription is set in config as well.

with https://management.azure.com/ I have an error:
```
def-group.........................> Failed getting access to Azure resource group 'nixops-0686a4eb-01ed-11e7-a282-0242bf42bf99-def-group'
error: Auth failure: Could not determine tenant.
```
but with https://login.windows.net/TENANT_ID
```
def-group.........................> Failed getting access to Azure resource group 'nixops-0686a4eb-01ed-11e7-a282-0242bf42bf99-def-group'
error: Auth failure: Error:Error:Error:Server returned error in RSTR - ErrorCode: InvalidRequest : FaultMessage: Invalid Request token_response:None token_response:None token_response:None
```
My user is admin, and only user (I'm using free subscription to test azure).

If anybody has any experience with setting this up, I could update documentation to make it more straight forward.

colemickens:
The Authority URL is going to `https://login.microsoftonline.com/common` or `https://login.microsoftonline.com/{TENANT_ID}`, but I don't think you really need to actually be setting it. I'm pretty sure I just set tenant id, user/pass.

You will likely also need to use a Tenant-specific email address instead of raw username/password. (@Phreedom do you remember)? I'd recommend you create a Service Principal and give it appropriate permissions rather than using your own personal account credentials as well.

spinus:
@colemickens thank you.
Do you know how to get tenant specific user email? (either web or azure cli)

I was thinking about setting up service principle but if I understand that correctly it's bound to an app, and an app requires URL which is used to authentication in the app, is that right? (my idea was to create nixops app which would be just API client, without providing any service) What is the way to do it?

colemickens:
You'd have to go into your AAD Tenant and add a user account for yourself. You can probably just use MSA credentials. I think that's what @Phreedom was doing.

Yes, you can make an AAD "Application" to get a Service Principal. The application AppID is the `ClientID` and the SP password is the `ClientSecret` that you'd use to login. I had one "App/SP" that I used for my NixOps operations, yes. That's a good model.

[...]

PR #853, opened 2018-01-26
Fix for accessing to Azure.

hyphon81:
#616 #766 
Now, we couldn't access to Azure with NixOps.
I had accessed to Azure with this codes, so, I would like to update it.


lblasc:
Works like described in updated manual, azure backend is useless without this patch. Thanks!

rbvermaa:
@hyphon81 Thanks for the fix, I couldn't test it before due to not having access to Azure anymore (fixed now).  @lblasc Thanks for confirming the code works. Merged in b267e2b

PR #857, opened 2018-01-27
Don't use port open check to determine if reboot completed. Fixes #856.

### LINKED ISSUE #856: waiting for the machine to finish rebooting does not work when the machine reboots too quickly
Using `nixops deploy --force-reboot`.

Good case:

```
nixos-cachecache-hetzner> closures copied successfully
machine1> updating GRUB 2 menu...
machine1> rebooting...
machine1> waiting for the machine to finish rebooting....[down]............................................................[up]
machine1> activation finished successfully
nixos-cachecache-hetzner> deployment finished successfully
```

Bad case:

```
machine1> setting custom nix.conf options
machine1> copying closure...
nixos-cachecache-hetzner> closures copied successfully
machine1> updating GRUB 2 menu...
machine1> rebooting...
packet_write_wait: Connection to 78.46.150.201 port 22: Broken pipe
machine1> waiting for the machine to finish rebooting............................................................................................................................
```

Note the `packet_write_wait` in the above.

The machine is perfectly up and running in that case.

I suspect there is a race condititon that if the machine shuts down so quickly that `packet_write_wait` appears before `waiting for the machine to finish rebooting` appears, then the reboot detection doesn't work.

nh2:
See #856

The old approach, waiting for the machine to not having an open
port, and then waiting for it to be open again, was insufficient,
because of the race condition that the machine rebooted so quickly
that the port was immediately open again without nixops noticing
that it went down. I experienced this on a Hetzner cloud server.

The new approach checks the `last reboot` on the remote side
to change, which is not racy.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -3,6 +3,7 @@
 import os
 import re
 import subprocess
+import time
 
 import nixops.util
 import nixops.resources
@@ -182,16 +183,53 @@ class MachineState(nixops.resources.ResourceState):
             reboot_command = "systemctl reboot"
         self.run_command(reboot_command, check=False)
         self.state = self.STARTING
+        # Note we always reset the connection to ensure the SSH control master
+        # connection is stopped.
+        # While an orderly `reboot` should make the remote kernel send RST to
+        # all TCP connections, we don't want to rely on that here.
         self.ssh.reset()
 
     def reboot_sync(self, hard=False):
         """Reboot this machine and wait until it's up again."""
+
+        # To check when the machine has finished rebooting in a race-free
+        # manner, we compare the output of `last reboot` before and after
+        # the reboot. Once the output has changed, the reboot is done.
+        def get_last_reboot_output():
+            # Note `last reboot` does not exist on older OSs like
+            # the Hetzner rescue system, but that doesn't matter much
+            # because all we care about is when the output of the
+            # command invocation changes.
+            # We use timeout=10 so that the user gets some sense
+            # of progress, as reboots can take a long time.
+            return self.run_command('last reboot --time-format iso | head -n 1', capture_stdout=True, timeout=10).rstrip()
+
+        # The server might be off.
+        # In that case, we just set `pre_reboot_last_reboot_output` to some string
+        # so that it changes when it comes up.
+        try:
+            pre_reboot_last_reboot_output = get_last_reboot_output()
+        except nixops.ssh_util.SSHConnectionFailed:
+            self.log("server is unreachable, so reboot completion detection may not be accurate if it's a network problem")
+            pre_reboot_last_reboot_output = "server is off or couldn't be contacted"
+
         self.reboot(hard=hard)
-        self.log_start("waiting for the machine to finish rebooting...")
-        nixops.util.wait_for_tcp_port(self.get_ssh_name(), self.ssh_port, open=False, callback=lambda: self.log_continue("."))
-        self.log_continue("[down]")
-        nixops.util.wait_for_tcp_port(self.get_ssh_name(), self.ssh_port, callback=lambda: self.log_continue("."))
-        self.log_end("[up]")
+
+        self.log_start("waiting for reboot to complete...")
+        while True:
+            last_reboot_output = None
+            try:
+                last_reboot_output = get_last_reboot_output()
+            except (nixops.ssh_util.SSHConnectionFailed, nixops.ssh_util.SSHCommandFailed):
+                # We accept this because the machine might be down,
+                # and show an 'x' as progress indicator in that case.
+                self.log_continue("x")
+            if last_reboot_output is not None and last_reboot_output != pre_reboot_last_reboot_output:
+                break
+            self.log_continue(".")
+            time.sleep(1)
+        self.log_end("done.")
+
         self.state = self.UP
         self.ssh_pinged = True
         self._ssh_pinged_this_time = True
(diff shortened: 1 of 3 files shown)


nh2:
Changed output with this patch:

```
nixos-cachecache-hetzner> closures copied successfully
machine1> updating GRUB 2 menu...
machine1> rebooting...
packet_write_wait: Connection to 1.2.3.4 port 22: Broken pipe
machine1> waiting for reboot to complete...done.
machine1> activation finished successfully
```

nh2:
Alternative output that can be created by this:

```
nixos-cachecache-hetzner> closures copied successfully
machine1> updating GRUB 2 menu...
machine1> rebooting...
machine1> waiting for reboot to complete...packet_write_wait: Connection to 78.46.150.201 port 22: Broken pipe

machine1> mux_client_request_session: read from master failed: Broken pipe
machine1> done.
machine1> activation finished successfully
```

So the output is not as pretty as before due to the errors, but at least it works without race-condition.

nh2:
I pushed a small fix (`def reboot()` being overridden in subclasses).

nixops/backends/__init__.py
```diff
@@ -182,16 +183,40 @@ def reboot(self, hard=False):
...
+        while True:
+            last_reboot_output = None
+            try:
+                last_reboot_output = get_last_reboot_output()
+            except nixops.ssh_util.SSHCommandFailed:
+                # We accept this because the machine might be down,
+                # and show an 'x' as progress indicator in that case.
+                self.log_continue("x")
+            if last_reboot_output is not None and last_reboot_output != pre_reboot_last_reboot_output:
+                break
```
domenkozar: Shouldn't this call `ssh.reset()` eventually?

nh2: Based on my current understanding, this is not needed, because the only thing `reset()` is call [`shutdown()`](https://github.com/nh2/nixops/blob/37bc5afde9531265b4e09efd5532b0097f4c94f3/nixops/ssh_util.py#L99) which exits the control master process, and I think that gets cleaned up automatically when the connection dies due to the machine rebooting.

nh2:
All good? This PR is the required base of my new PR #948, so it would be cool if we could finish this one.

nixops/backends/__init__.py
```diff
@@ -182,16 +183,40 @@ def reboot(self, hard=False):
...
-        nixops.util.wait_for_tcp_port(self.get_ssh_name(), self.ssh_port, open=False, callback=lambda: self.log_continue("."))
-        self.log_continue("[down]")
-        nixops.util.wait_for_tcp_port(self.get_ssh_name(), self.ssh_port, callback=lambda: self.log_continue("."))
-        self.log_end("[up]")
+
+        # To check when the machine has finished rebooting in a race-free
+        # manner, we compare the output of `last reboot` before and after
+        # the reboot. Once the output has changed, the reboot is done.
+        def get_last_reboot_output():
+            return self.run_command('last reboot --time-format iso | head -n1', capture_stdout=True).rstrip()
```
aszlig: Both the Hetzner rescue system and the actual NixOS system are using `systemd`, so maybe it's a better idea to use `systemd-analyze` because it [fails whenever bootup is not finished](https://github.com/systemd/systemd/blob/v238/src/analyze/analyze.c#L294). At least that would avoid the current vs. last string comparison.

nh2: I don't think that can work because when NixOS is waiting for nixops keys to be uploaded (so, right here), you'll get

```
# systemd-analyze
Bootup is not yet finished. Please try again later.
```

nh2:
Updated the PR, I noticed I have to not only catch `nixops.ssh_util.SSHCommandFailed`, but also `nixops.ssh_util.SSHConnectionFailed` explicitly.

nh2:
Made another small improvement so that `nixops reboot --hard` also works when the machine is off.

nixops/backends/__init__.py
aszlig: Oh, right... you're correct and after all we don't care about a truly finished reboot but only want to make sure the machine has rebooted *after all*, so `last reboot` should work.

nixops/backends/__init__.py
aszlig: Hm, in theory you're right, but if the machine didn't manage to properly close the TCP socket (for example due to a hard reboot), the control master process is still alive.

aszlig:
@nh2: Approved, but see my [last comment](https://github.com/NixOS/nixops/pull/857#discussion_r191063993).

nixops/backends/__init__.py
```diff
@@ -182,16 +183,46 @@ def reboot(self, hard=False):
...
+        # manner, we compare the output of `last reboot` before and after
+        # the reboot. Once the output has changed, the reboot is done.
+        def get_last_reboot_output():
+            # Note `last reboot` does not exist on older OSs like
+            # the Hetzner rescue system, but that doesn't matter much
+            # because all we care about is when the output of the
+            # command invocation changes.
+            # We use timeout=10 so that the user gets some sense
+            # of progress, as reboots can take a long time.
+            return self.run_command('last reboot --time-format iso | head -n1', capture_stdout=True, timeout=10).rstrip()
```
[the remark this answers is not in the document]
coretemp: `head -n1` ->  `head -n 1` http://pubs.opengroup.org/onlinepubs/9699919799/utilities/head.html

nh2:
OK, I've updated the commits to address the remaining comments.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #866, opened 2018-02-06
add labels option to be applied to GCE instances

AmineChikhaoui:
This seems to be implemented properly in <https://github.com/apache/libcloud/pull/1115>
but not released yet from what I see. So this is a temproary
implementation and we should modify it once a new libcloud version is
available.

### DIFF
--- a/nix/common-gce-options.nix
+++ b/nix/common-gce-options.nix
@@ -1,0 +1,16 @@
+{ lib }:
+
+with lib;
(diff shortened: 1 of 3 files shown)


AmineChikhaoui:
@rbvermaa Fine with this ?

AmineChikhaoui:
@rbvermaa ping

PR #871, opened 2018-02-19
get release from config.system.nixos.release as fallback

dtzWill:
Not sure why but recently started seeing eval problems
complaining about lack of nixosVersion, this fix
tries to get "config.system.nixos.release" after
failing to get config.system.nixosRelease,
inspired by recent plymouth change:

0500cf79afeb3e0c61458b202b361064397a7776

I've no idea when/if/why nixosVersion should be used,
hopefully it's the most legacy of the options...?

Sorry for less-than-thorough background research
on this PR, hopefully it's still useful!

### DIFF
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -303,7 +303,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
         { inherit (v.config.deployment) targetEnv targetPort targetHost encryptedLinksTo storeKeysOnMachine alwaysActivate owners keys hasFastConnection;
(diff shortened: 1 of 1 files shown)


edolstra:
Alternatively, we could revert the option rename. I don't think there was a very compelling motivation for renaming them in the first place.

ledettwy:
> Sorry for less-than-thorough background research on this PR, hopefully it's still useful!

Well, it fixed the problem for me, so I'd call it useful :)

dtzWill:
Well we certainly should do one or the other, I have no preference.  No reason to keep nixops broken, right?

Mic92:
The attributes where changed here: https://github.com/NixOS/nixops/pull/871#issuecomment-366997997

rbvermaa:
As the rename was not rolled back, I applied a similar patch from another PR.

PR #874, opened 2018-02-23
ec2/eval-machine-info: Change config.system.nixosVersion to config.system.nixos.version, because the former doesn't exist in newer releases of nixos

lejonet:
I bumped into this bug while doing development work on a nixpkgs and wanting to pull in latest master of nixpkgs to base my work on.
All of a sudden I was getting an error that nixops couldn't find config.system.nixosVersion, even tho I had the relevant entries in rename.nix, and I found in the version module that the variable had been renamed to config.system.nixos.version.

I've verified that with this patch, nixops works again when running from nixpkgs master, so my guess is that its just a matter of time before this change trickles down to the channels and thus breaking nixops.

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -150,7 +150,7 @@ let
     then "/dev/" + builtins.substring 12 100 dev
     else dev;
 
-  nixosVersion = builtins.substring 0 5 config.system.nixosVersion;
+  nixosVersion = builtins.substring 0 5 config.system.nixos.version;
 
   amis = import <nixpkgs/nixos/modules/virtualisation/ec2-amis.nix>;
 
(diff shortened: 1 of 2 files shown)


lejonet:
I've verified that nixops works on 17.09 stable branch with this patch.

cleverca22:
duplicates #875 ?

lejonet:
Yeah, its more or less a duplicate of https://github.com/NixOS/nixops/pull/875

teto:
This fixes it for me on nixos-unstable.

rbvermaa:
Although the rename should work, at some point that will be removed, so I will apply the patch.

rbvermaa:
Ugh, wasn't paying attention and merged wrong thing. Fixing the commit now.

cleverca22:
https://gist.github.com/cleverca22/65d8f09a8440961aab12da205242c57a is what i did to fix it if anybody wants to compare

PR #875, opened 2018-02-24
Look for release version in config.system.nixos.release

### LINKED ISSUE #873: NixOps does not work with Nix 2.0
Let's round up assorted bugs here. Only one I've spotted so far, but I'll probably find more.

- [ ] `attribute ‘nixosVersion’ missing, at nix/eval-machine-info.nix:284`
  These attributes have been renamed to `nixos.version` and so forth.

Baughn:
Fixes #873, maybe.

### DIFF
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -303,7 +303,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
         { inherit (v.config.deployment) targetEnv targetPort targetHost encryptedLinksTo storeKeysOnMachine alwaysActivate owners keys hasFastConnection;
-          nixosRelease = v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
+          nixosRelease = v.config.system.nixos.release or v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
           azure = optionalAttrs (v.config.deployment.targetEnv == "azure")  v.config.deployment.azure;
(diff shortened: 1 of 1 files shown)


cleverca22:
just ran into this myself, and ive narrowed down the cause some more
nixos moved the option, and has mkRenamedOptionModule for backwards compat

nixops has been updated to look in the new location

but older nixpkgs, lack the forwards compat changes, and just fail hard

cleverca22:
i think i found the problem
the `"config"` in these entries is a typo and leads to the rename not even applying

https://github.com/NixOS/nixpkgs/blob/2e6b796761672e0e3ed685487007bb0d99126d91/nixos/modules/rename.nix#L192-L198

cleverca22:
and i think the alias only works for setting the old name

it doesnt appear to work for reading the old name

nix/eval-machine-info.nix
```diff
@@ -303,7 +303,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
         { inherit (v.config.deployment) targetEnv targetPort targetHost encryptedLinksTo storeKeysOnMachine alwaysActivate owners keys hasFastConnection;
-          nixosRelease = v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
+          nixosRelease = v.config.system.nixos.release or v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
```
Mic92: maybe it should include an `or` for older nixops releases.

Baughn: I thought that's what I was doing...

Mic92: right

rbvermaa:
Pushed a different change to master yesterday.

PR #877, opened 2018-02-25
Add a check function for the EBS resource.

### LINKED ISSUE #876: Attatched volumes don't get marked as missing after check.
If a volume get deleted manually, it doesn't get marked as **MISSING** after `check` operation. Also if the volume was attached to an instance we get an obscure error when **nixops** try to attach it again using `deploy`. This is the minimal expression to reproduce it.
```
let
  region = "us-east-1";
  zone = "us-east-1d";
in
{
  resources.ec2KeyPairs.my-key-pair =
    { inherit region; };

  resources.ec2SecurityGroups.ssh-security-group = {
    inherit region;
    description = "Security group for NixOps tests";
    rules = [ {
      fromPort = 22;
      toPort = 22;
      sourceIp = "0.0.0.0/0";
    } ];
  };

  machine =
  {resources, lib, ...}:
  {
    deployment.targetEnv = "ec2";
    deployment.ec2 = {
      inherit region zone;
      ami = "ami-40bee63a";
      securityGroups = [ resources.ec2SecurityGroups.ssh-security-group ];
      instanceType = "t1.micro";
      blockDeviceMapping."/dev/sdf".disk = resources.ebsVolumes.data_volume;
      keyPair = resources.ec2KeyPairs.my-key-pair ;
      };

  };

  resources.ebsVolumes.data_volume = {
      inherit region zone;
      size = 1;
      volumeType = "gp2";
  };


}
```
After deploying the network, delete the volume manually using
```
aws ec2 delete-volume --volume-id VOLUME_ID --region us-east-1
```
Don't forget to detach it before deleting 
```
aws ec2 detach-volume --volume-id VOLUME_ID  --force 
```
If we run a `check` operation, *nixops* notice that the volume is missing but it doesn't mark it as such
```
[nix-shell:~/git/nixops]$ nixops check -d volume_bug 
Machines state:
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
| Name    | Exists | Up  | Reachable | Disks OK | Load avg.      | Units                                  | Notes                                                      |
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
| machine | Yes    | Yes | Yes       | No       | 0.04 0.28 0.21 | proc-sys-fs-binfmt_misc.mount [failed] | volume ‘VOLUME_ID‘ not attached to ‘/dev/xvdf’             |
|         |        |     |           |          |                |                                        | volume ‘VOLUME_ID’ no longer exists                        |
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
Non machines resources state:
+--------------------+--------+
| Name               | Exists |
+--------------------+--------+
| data_volume        | Yes    |
| my-key-pair        | Yes    |
| ssh-security-group | Yes    |
+--------------------+--------+
```
And since the volume wasn't marked as deleted it won't be created in the next deploy operation. Leading to the following error.
```

[nix-shell:~/git/nixops]$ nixops deploy -d volume_bug --allow-recreate
machine...........> warning: device ‘/dev/xvdf’ was manually detached!
machine...........> warning: volume ‘VOLUME_ID’ has disappeared; will create an empty volume to replace it
error: 'NoneType' object has no attribute 'status'
```
Please note that I'm working on this, so please don't create a PR for it

Chakerbh:
If missing the volume will be marked as such, after a check
operation. Also show more meaningful error message when a volume
disappear. This fix #876.

### DIFF
--- a/nixops/resources/cloudwatch_metric_alarm.py
+++ b/nixops/resources/cloudwatch_metric_alarm.py
@@ -135,7 +135,6 @@ class CloudwatchMetricAlarmState(nixops.resources.ResourceState):
         cfg['InsufficientDataActions'] = map(resolve_action, defn.insufficient_data_actions)
 
         if self.put_config != cfg or check:
-            pprint(cfg)
             client.put_metric_alarm(**cfg)
             with self.depl._db:
                 self.state = self.UP


nixops/resources/ebs_volume.py
```diff
@@ -141,6 +141,10 @@ def create(self, defn, check, allow_reboot, allow_recreate):
                 states=['available', 'in-use'])
             self.state = self.UP
 
+    def check(self):
+        volume = nixops.ec2_utils.get_volume_by_id(self.connect(self.region), self.volume_id)
+        if volume is None:
+            self.state = self.MISSING
```
AmineChikhaoui: we probably need a warning that the volume is no longer available here as well.

nixops/resources/ebs_volume.py
```diff
@@ -141,6 +141,10 @@ def create(self, defn, check, allow_reboot, allow_recreate):
                 states=['available', 'in-use'])
             self.state = self.UP
 
+    def check(self):
```
AmineChikhaoui: hm isn't the method name _check ?

AmineChikhaoui: nevermind, checked again, this should work as well.

nixops/backends/ec2.py
```diff
@@ -506,6 +506,8 @@ def create_after(self, resources, defn):
 
     def attach_volume(self, device, volume_id):
         volume = nixops.ec2_utils.get_volume_by_id(self.connect(), volume_id)
+        if not volume:
+            raise Exception("volume {0} doesn't exist".format(volume_id))
```
AmineChikhaoui: Maybe you can suggest running `nixops check` here to reconcile the state with the real-world state.

nixops/resources/ebs_volume.py
Chakerbh: Well, the disks check doesn't include notes like the machine one. But after updating the volume state, the check table show it as missing
```
Machines state:
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
| Name    | Exists | Up  | Reachable | Disks OK | Load avg.      | Units                                  | Notes                                                      |
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
| machine | Yes    | Yes | Yes       | No       | 0.00 0.00 0.00 | proc-sys-fs-binfmt_misc.mount [failed] | volume ‘vol-096cbe0647ee88eae’ not attached to ‘/dev/xvdf’ |
|         |        |     |           |          |                |                                        | volume ‘vol-096cbe0647ee88eae’ no longer exists            |
+---------+--------+-----+-----------+----------+----------------+----------------------------------------+------------------------------------------------------------+
Non machines resources state:
+--------------------+--------+
| Name               | Exists |
+--------------------+--------+
| data_volume        | No     |
| my-key-pair        | Yes    |
| ssh-security-group | Yes    |
+--------------------+--------+
```
The only thing that I can add is a message like this 
```
data_volume.......> Volume VOLUME_ID is missing!
``` 
Before the check tables

AmineChikhaoui: I see, yeah seems the existent warnings are fine already.

Chakerbh:
Ping @rbvermaa

Issue #847, opened 2018-01-23
Rely on qemu agent


teto:
I sometimes get:
```
client> waiting for SSH...
Traceback (most recent call last):
  File "/home/teto/nixops/scripts/nixops", line 983, in <module>
    args.op()
  File "/home/teto/nixops/scripts/nixops", line 406, in op_deploy
    max_concurrent_activate=args.max_concurrent_activate)
  File "/home/teto/nixops/nixops/deployment.py", line 1036, in deploy
    self.run_with_notify('deploy', lambda: self._deploy(**kwargs))
  File "/home/teto/nixops/nixops/deployment.py", line 1028, in run_with_notify
    f()
  File "/home/teto/nixops/nixops/deployment.py", line 1036, in <lambda>
    self.run_with_notify('deploy', lambda: self._deploy(**kwargs))
  File "/home/teto/nixops/nixops/deployment.py", line 975, in _deploy
    nixops.parallel.run_tasks(nr_workers=-1, tasks=self.active_resources.itervalues(), worker_fun=worker)
  File "/home/teto/nixops/nixops/parallel.py", line 44, in thread_fun
    result_queue.put((worker_fun(t), None, t.name))
  File "/home/teto/nixops/nixops/deployment.py", line 948, in worker
    r.create(self.definitions[r.name], check=check, allow_reboot=allow_reboot, allow_recreate=allow_recreate)
  File "/home/teto/nixops/nixops/backends/libvirtd.py", line 144, in create
    self.start()
  File "/home/teto/nixops/nixops/backends/libvirtd.py", line 254, in start
    self._wait_for_ip(0)
  File "/home/teto/nixops/nixops/backends/libvirtd.py", line 229, in _wait_for_ip
    ip = self._parse_ip()
  File "/home/teto/nixops/nixops/backends/libvirtd.py", line 216, in _parse_ip
    ifaces = self.dom.interfaceAddresses(libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE, 0)
  File "/nix/store/zj4dq8ipqiwakj0hql5kdrsql11laf0g-python2.7-libvirt-3.10.0/lib/python2.7/site-packages/libvirt.py", line 1347, in interfaceAddresses
    if ret is None: raise libvirtError ('virDomainInterfaceAddresses() failed', dom=self)
libvirt.libvirtError: pointeur de réseau invalide dans virNetworkGetDHCPLeases

```
[...]

PR #824, opened 2017-12-28
libvirtd: Add support for remote libvirt URIs

erosennin:
This PR adds support for deploying to remote libvirtd hosts via `qemu+ssh://...` or `qemu+tcp://...` URIs.

- [x] Make the URI configurable instead of hardcoging it to `qemu:///system`.
- [x] Use libvirt API to upload the disk image instead of using the local image directory.
- [x] Use libvirt API to determing the QEMU executable instead of using `/run/current-system/sw/bin/qemu-system-x86_64`.
- [x] Error handling.
- [x] Backward compatibility.
- [x] Update documentation and examples.

PR #881, opened 2018-02-28
[WIP] Qemu agent

teto:
I do some testing on the TCP stack and in general to support static IP VMs etc it might be best to rely on the VM Agent.

The only missing part is the script to start the agent in the VM. There seems to be one at https://github.com/NixOS/nixpkgs/issues/34722 so I hope to  revisit this once it appears in nixpkgs.
Related issue https://github.com/NixOS/nixops/issues/847


teto:
for whatever reason, I get plenty of errors like
```
File "/nix/store/dwd62qwwr4jzrcpp6nfrgq9mzrnjp73m-python2.7-libvirt-4.1.0/lib/python2.7/site-packages/libvirt.py", line 1347, in interfaceAddresses
    if ret is None: raise libvirtError ('virDomainInterfaceAddresses() failed', dom=self)
```
which I believe could be solved with the agent. I believe @mbrgm has a working agent ? do you intend to share it ? I would hate to duplicate the work :(

mbrgm:
Yes, it's working and I intend to share it. However, it's based on #824 and I don't know how to handle this in a PR... Should I just put my branch on top of #824 (this is the current state), create a PR and then rebase whenever #824 changes?

teto:
Sure you can open a PR prefixed with [WIP] (work in Progress) and rebase once #824 is merged. This prevents people from reimplementing things already being developed and let some users test it in advance. The dev activity is not that huge on nixops so it shouldn't be a burden. 
If you don't feel like opening a PR yet that's fine but I would love if possible to copy your code (I searched in your repo to no avail) since I've had tons of network problems when trying to do source routing xD

teto:
sorry to nag you but may I ask when you intend to make it public ? I have a tight deadline so I might reimplement it myself (not like it's too much work) but I would rather improve what you have if need be.

teto:
closing this as a better PR is on the way

PR #900, opened 2018-03-08
Add new option "NIXOPS_SSHOPTS"

andrewchambers:
This new option is an escape hatch allowing
you to pass arbitrary flags to ssh. This
was added to help do things like bootstrap
a 'none' deployment in an unusual environment.

### DIFF
--- a/doc/manual/nixops.xml
+++ b/doc/manual/nixops.xml
@@ -199,6 +199,14 @@ cloud.</para>
 
   </varlistentry>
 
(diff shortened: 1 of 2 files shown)


andrewchambers:
I have removed nixops as a dependency from my project and thus will no longer track this issue.

aij:
Isn't this still a good idea though?

Sorry the maintainers seem to have gone missing. (hopefully only temporarily...)

PR #901, opened 2018-03-08
Allow non-subdomains in EC2 Route53

ElvishJerricco:
This is part issue part PR. I don't actually have any idea why this was like this to begin with, but it seems deliberate. But this change works for me, so I'm opening a PR to see what ought to be done.

For some reason, you can't use `deployment.ec2.route53` to bind a non-subdomain like `foo.com` to an EC2 instance.


ElvishJerricco:
Ah, I hadn't seen `recordValues`.

Issue #883, opened 2018-02-28
postprocess the libvirt domain.xml ?


teto:
Some of my (dev) VMs crash from times to times (business as usual). I would like to generate from nixops a domain with settings as described in https://gymnasmata.wordpress.com/2010/12/02/setting-up-gdb-to-work-with-qemu-kvm-via-libvirt/ (e.g. add in the top-level tag the xml rules, and add the qemu command line tags), which is impossible to do without editing the domain xml via virsh.
It's not realistic to support all libvirt options but maybe we could have a postprocess command ? (but it becomes impure I guess). I suppose most people would just use the plain Qemu command but it can be tough to write it at some point and I like the `nixops deploy`. Is there any alternative to just maintaining my fork ?

teto:
Or the other way around,  would it be ok to have a setting to load an alternate base XML ?

PR #903, opened 2018-03-08
[RDY] Format domain XML from a file to allow for more flexibility

teto:
Related to https://github.com/NixOS/nixops/issues/883.

The PR adds a  deployment.libvirtd.template setting so that users can upload their own domain.xml. They can use a few variables that will be replaced such as {interfaces} (look at domain.xml.in for an example). 
I wonder if it wouldn't be best to accept a string instead of a filename ?

I need to test it a bit more.

### DIFF
--- a/nix/libvirtd.nix
+++ b/nix/libvirtd.nix
@@ -53,6 +53,32 @@ in
       '';
     };
 
+    deployment.libvirtd.template = mkOption {
+      type = types.str;
+      default = ''<domain type="{domain_type}">
+          <name>{name}</name>
+          <memory unit="MiB">{memory_size}</memory>
+          <vcpu>{vcpu}</vcpu>
+          {os}
+          <devices>
+              <emulator>{emulator}</emulator>
+            <disk type="file" device="disk">
+              <driver name="qemu" type="qcow2"/>
+              <source file="{diskPath}"/>
+              <target dev="hda"/>
+            </disk>
+            {interfaces}
+            <input type="keyboard" bus="usb"/>
+            <input type="mouse" bus="usb"/>
+            {extra_devices}
+          </devices>
+          {extra_domain}
+        </domain>'';
+      description = ''
+        Template for the libvirt domain. You can use all other parameters.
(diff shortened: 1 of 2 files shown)


teto:
I changed the parameter to accept a string instead as it seems more "nix-like", having an extra file is a bother.
Eventually this setting could replace extraDevicesXML / extraDomainXML.

nixops/backends/libvirtd.py
```diff
@@ -185,38 +186,19 @@ def _make_os(defn):
...
-
-        return domain_fmt.format(
-            self._vm_id(),
-            defn.memory_size,
-            qemu,
-            self._disk_path(defn),
-            defn.vcpu,
-            defn.domain_type
+        result = defn.domain_tpl.format(
+            name=self._vm_id(),
```
teto: Here I tried to replace the `memory_size=defn.memory_size,` by `**defn` (more elegant) but couldn't succeed.

teto:
I want to use shared folders with libvirt (http://rabexc.org/posts/p9-setup-in-libvirt), if there is any interest I can provide a proper interface   `deployment.libvirtd.sharedFolders` but  this PR would allow me to make it work inbetween.

teto:
bump. This should be backwards compatible.

teto:
@rbvermaa what do you think ?

nix/libvirtd.nix
```diff
@@ -53,6 +53,32 @@ in
       '';
     };
 
+    deployment.libvirtd.template = mkOption {
```
AmineChikhaoui: I guess I'm confused what's the point of having this as a nixops option ?
What if someone changes this to:
```
deployment.libvirtd.template = ''
<other template that doesn't use the same variables>
''
```
That'll basically make it inconsistent with what the python code expects

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #906, opened 2018-03-11
Use proper buildscript

tomberek:


AmineChikhaoui:
Not sure how is this still showing in the diff, but the change is already done as far as I can see https://github.com/NixOS/nixops/blob/master/maintainers/build-virtualbox-image.sh#L3.

PR #941, opened 2018-05-03
Fixed a bug creating route53 record sets

boxyoman:


boxyoman:
I was just stupid. I just used it wrong.

PR #948, opened 2018-05-08
Hetzner partitioning script

nh2:
This PR is based on top of PR #857 and thus includes its first commit; that PR should be merged before this one.

This PR modernises how the hetzner backend gets its inputs (using `config` instead of custom XML parsing), and then adds 3 options via which you can do custom partitioning. This is needed e.g. when you want to make an ext4 with external journal, using a combination of SSDs and HDDs that are common at Hetzner.

CC @aszlig @cleverca22

### DIFF
--- a/nix/hetzner.nix
+++ b/nix/hetzner.nix
@@ -90,12 +90,181 @@ with lib;
         part swap --recommended --label=swap --fstype=swap --ondisk=vda
         part / --fstype=ext4 --label=root --grow --ondisk=vda
       '';
-      type = types.lines;
+      type = types.nullOr types.lines;
       description = ''
         Specify layout of partitions and file systems using Anacondas Kickstart
         format. For possible options and commands, please have a look at:
 
         <link xlink:href="http://fedoraproject.org/wiki/Anaconda/Kickstart"/>
+
+        If Kickstart is not sufficient for your partitioning needs,
+        consider the <option>partitioningScript</option> option instead.
+
+        The <option>partitions</option> and <option>partitioningScript</option>
+        options are mutually exclusive.
+      '';
+    };
+
+    partitioningScript = mkOption {
+      type = types.nullOr types.lines;
+      default = null;
+      example = ''
+        # Example custom partitioningScript
+        # that creates an ext4 with external journal, across two RAID1s
+        # (one on HDDs, one on SSDs).
+
+        set -x
+        set -euo pipefail
+
+        # If the SSD doesn't support the RZAT (Return Zero After Trim) feature,
+        # we can't do the `lazy_journal_init=1` journal creation speedup
+        # below, so fail early in that case.
+        # Note that as per
+        #   https://github.com/torvalds/linux/blob/e64f638483a21105c7ce330d543fa1f1c35b5bc7/drivers/ata/libata-core.c#L4242-L4250
+        # TRIM in general is optional and thus this would be unsafe,
+        # but the kernel announces RZAT only for a whitelist of devices
+        # that are known to execute TRIM when requested.
+        #
+        # Note that this is probably not needed when the ext4 journal is on top
+        # of an mdadm RAID (because that one likely guarantees to read zeros from
+        # freshly initialised RAID arrays, but I haven't checked that this really
+        # works), but we keep it in here just in case it doesn't work or
+        # somebody wants to run the journal NOT on top of a RAID.
+        #
+        # TODO Fall back to slow `lazy_journal_init=1` if RZAT isn't supported.
+        if hdparm -I /dev/sda | grep -i 'Deterministic read ZEROs after TRIM'; then echo "RZAT supported, can use lazy_journal_init=1 safely"; else echo "RZAT not supported on /dev/sda, cannot use lazy_journal_init=1 safely, exiting" 1>&2; exit 1; fi
+        if hdparm -I /dev/sdb | grep -i 'Deterministic read ZEROs after TRIM'; then echo "RZAT supported, can use lazy_journal_init=1 safely"; else echo "RZAT not supported on /dev/sdb, cannot use lazy_journal_init=1 safely, exiting" 1>&2; exit 1; fi
+
+        # Stop RAID devices if running, otherwise we can't modify the disks below.
+        test -b /dev/md0 && mdadm --stop /dev/md0
+        test -b /dev/md1 && mdadm --stop /dev/md1
+
+        # Zero out SSDs with TRIM command, so that `lazy_journal_init=1` can be safely used below.
+        blkdiscard /dev/sda
+        blkdiscard /dev/sdb
+
+        # Create BIOS boot partition and main partition for each SSD and HDD.
+        # Note Hetzner does use BIOS, not UEFI.
+        # We use GPT because these disks could be too large for MSDOS partitions (e.g. 10TB disks).
+        parted --script -a optimal /dev/sda -- mklabel gpt mkpart primary 1MiB 2MiB set 1 bios_grub on mkpart primary 2MiB '100%'
+        parted --script -a optimal /dev/sdb -- mklabel gpt mkpart primary 1MiB 2MiB set 1 bios_grub on mkpart primary 2MiB '100%'
+        parted --script -a optimal /dev/sdc -- mklabel gpt mkpart primary 1MiB 2MiB set 1 bios_grub on mkpart primary 2MiB '100%'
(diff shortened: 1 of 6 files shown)


nixops/backends/hetzner.py
```diff
@@ -51,16 +51,38 @@ def get_type(cls):
...
-                 ("create_sub_account", "createSubAccount", "bool"),
-                 ("robot_user", "robotUser", "string"),
-                 ("robot_pass", "robotPass", "string"),
-                 ("partitions", "partitions", "string")]
-        for var, name, valtype in attrs:
-            node = x.find("attr[@name='" + name + "']/" + valtype)
-            setattr(self, var, xml_expr_to_python(node))
+
+        self.main_ipv4 = config["hetzner"]["mainIPv4"]
+        assert type(self.main_ipv4) is str
```
FRidh: Use descriptors instead? Possibly http://traitlets.readthedocs.io.

nh2: These are nice ideas, but here I'm simply unifying the style to use the same as the other backends (such as the EC2 one, https://github.com/NixOS/nixops/blob/f10999a111fba80e1dc597b7e14069ea620832af/nixops/backends/ec2.py#L39).

If we want to push for a change in style in general, we should probably tackle that separately for all backends and in coordination with whoever takes care of the those.

Regarding traitlets, I'm all for more type checking, but as far as I can tell it's an extra dependency so I can't tell whether or not we want to add it to nixops.

nh2:
TODO for myself:

Fix `example`:

* [ ] `mkpart primary` is misleading for GPT tables; as per https://www.gnu.org/software/parted/manual/html_node/mkpart.html#mkpart:
  > part-type is one of ‘primary’, ‘extended’ or ‘logical’, and may be specified only with ‘msdos’ or ‘dvh’ partition tables
* [ ] Switch to `MB` instead of `MiB`, add note
  ```
        # Note we use "MB" instead of "MiB" because otherwise `--align optimal` has no effect;
        # as per documentation https://www.gnu.org/software/parted/manual/html_node/unit.html#unit:
        # > Note that as of parted-2.4, when you specify start and/or end values using IEC
        # > binary units like "MiB", "GiB", "TiB", etc., parted treats those values as exact
  ```

nix/hetzner.nix
```diff
@@ -90,12 +90,177 @@ with lib;
...
       '';
-      type = types.lines;
+      type = types.nullOr types.lines;
       description = ''
         Specify layout of partitions and file systems using Anacondas Kickstart
         format. For possible options and commands, please have a look at:
 
         <link xlink:href="http://fedoraproject.org/wiki/Anaconda/Kickstart"/>
+
+        If the Kickstart is not sufficient for your partitioning needs,
```
aszlig: Probably remove the `the`: `If Kickstart is not sufficient...`

nix/hetzner.nix
```diff
@@ -90,12 +90,177 @@ with lib;
...
+
+        Note as of writing, Hetzner uses BIOS, not UEFI, so if you want
+        to use GPT partition tables (which you need in case you want to
+        make partitions larger than 2 TiB) you will likely have to make
+        a BIOS boot partition
+        (<link xlink:href="http://fedoraproject.org/wiki/Anaconda/Kickstart"/>).
+
+        Where possible, use the simpler "partitions" option instead of this option.
+
+        The "partitions" and "partitioningScript" options are mutually exclusive.
```
aszlig: `<option>partitions</option>` and `<option>partitioningScript</option>`

nix/hetzner.nix
```diff
@@ -90,12 +90,177 @@ with lib;
...
+        and after formatting, to mount the root filesystem at /mnt.
+
+        This option is required when "partitioningScript" is used.
+      '';
+    };
+
+    filesystemInfo = mkOption {
+      type = types.nullOr types.attrs;
+      default = null;
+      example = {
```
aszlig: Maybe use `literalExample` here, otherwise this ends up like this: 

```nix
{ boot = { loader = { grub = { devices = [ "/dev/sda" "/dev/sdb" "/dev/sdc" "/dev/sdd" ] ; } ; } ; } ; fileSystems = { / = { fsType = "ext4"; label = "root"; options = [ "journal_path=/dev/disk/by-label/rootjournal" "data=journal" "errors=remount-ro" ] ; } ; } ; swapDevices = [ ] ; }
```

nix/hetzner.nix
```diff
@@ -90,12 +90,177 @@ with lib;
...
+              "data=journal"
+              "errors=remount-ro"
+            ];
+          };
+        };
+      };
+      description = ''
+        Override the filesystem info obtained from the machine after partitioning.
+
+        This option is required when "partitioningScript" is used, but can also
```
aszlig: Again, `<option/>`.

nixops/backends/hetzner.py
aszlig: Apart from that, those types should already be checked by the NixOS module system, so I think most of those asserts aren't really needed.

nixops/backends/hetzner.py
```diff
@@ -223,13 +247,27 @@ def _bootstrap_rescue_for_existing_system(self):
...
+                          mount_script=None,
+                          fs_info=None):
         """
         Bootstrap everything needed in order to get Nix and the partitioner
         usable in the rescue system. The keyword arguments are only for
         partitioning, see reboot_rescue() for description, if not given we will
         only mount based on information provided in self.partitions.
+
+        Exactly one of `partitions` and `partitioning_script` must be given as
+        non-None value.
```
aszlig: I'd check this via the NixOS module system as well.

nix/hetzner.nix
nh2: Done

nix/hetzner.nix
nh2: Done

nix/hetzner.nix
nh2: Done

nix/hetzner.nix
nh2: Done

nh2:
Addressed a couple comments, some other ones still outstanding

nixops/backends/__init__.py
```diff
@@ -184,16 +185,46 @@ def reboot(self, hard=False):
...
+
+        pre_reboot_last_reboot_output = get_last_reboot_output()
+
+        # We set reset=False so that we can continue running `last reboot`
+        # remotely; when the reboot happens, our SSH connection will be reset
+        # by the remote side instead.
+        self.reboot(hard=hard, reset=False)
+
+        self.log_start("waiting for reboot to complete...")
+        while True:
```
coretemp: This doesn't look robust, because perhaps Hetzner did something wrong and it doesn't actually ever come up. I see lots of people who write `while True: ` loops, but in reality, if it hasn't started after 3 minutes, you will start to wonder what's going on and you would have to manual work. 

As such, this level of automation is an improvement over having nothing, but there is room for improvement.

nixops/backends/__init__.py
```diff
@@ -184,16 +185,46 @@ def reboot(self, hard=False):
...
+                last_reboot_output = get_last_reboot_output()
+            except (nixops.ssh_util.SSHConnectionFailed, nixops.ssh_util.SSHCommandFailed):
+                # We accept this because the machine might be down,
+                # and show an 'x' as progress indicator in that case.
+                self.log_continue("x")
+            if last_reboot_output is not None and last_reboot_output != pre_reboot_last_reboot_output:
+                break
+            self.log_continue(".")
+            time.sleep(1)
+        self.log_end("done.")
```
coretemp: In a log it would be more useful see a line that says what actually was `done`.

aszlig: @coretemp: This is done in [log_start](https://github.com/NixOS/nixops/pull/948/files/3ca62f276bf1f964f9ded828be9eaa556bcfa6c2#diff-6ea77fd00d5f092f8150c2398eaf53d8R213), which results in something like `waiting for reboot to complete.....xxxx....done.`. However I guess we could also refactor the logging mechanism to allow for context managers so this becomes more clear.

nixops/backends/hetzner.py
coretemp: Yes, but not a reason not to merge, is it?

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #971, opened 2018-06-20
Regenerate ec2-properties.nix

lo1tuma:
This updates the `nix/ec2-properties.nix` file generated by `nix/generate-ec2-properties.py`.
Motivation behind this is that there are new instance types like `m5.*` or `c5.*`.


nix/ec2-properties.nix
```diff
@@ -1,64 +1,97 @@
 {
   "c1.medium" = { cores = 2; memory = 1740; allowsEbsOptimized = false; };
-  "c1.xlarge" = { cores = 8; memory = 7168; allowsEbsOptimized = true; };
-  "c3.2xlarge" = { cores = 8; memory = 15360; allowsEbsOptimized = true; };
-  "c3.4xlarge" = { cores = 16; memory = 30720; allowsEbsOptimized = true; };
+  "c1.xlarge" = { cores = 8; memory = 7168; allowsEbsOptimized = false; };
```
lo1tuma: I’m not quite sure why this has been changed. Is there a bug in `nix/generate-ec2-properties.py`?

rbvermaa: It looks like some of the newer instance types that have EBS optimization on by default [1] do not really support disabling it, basically making disabling it a no-op. Not sure why the c3's changed though. 

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSOptimized.html

rbvermaa: I will review generate-ec2-properties.py today, and check if I can see an obvious bug.

rbvermaa: Looking at the JSON that AWS produces, it seems that they do not provide the ebsOptimized flag anymore in the old generation models in the pricing JSON file.

domenkozar:
https://github.com/NixOS/nixops/pull/1120

PR #972, opened 2018-06-21
DO NOT MERGE: Nix 2 out of memory workaround

nh2:
This is a workaround for the `out of memory` issue introduced with Nix 2 and present in NixOS 18.03 images (e.g. NixOS 18.03 AMIs).

It includes my fix from https://github.com/NixOS/nix/pull/2206

## How to use this workaround

Use this branch and set

```
deployment.ec2.nixosAmiVersion = "17.09";
```

for your AWS machines.

That will ensure the machine gets initially deployed with 17.09 which doesn't have the memory problem, will then get nix 2 with my patch deployed, so that in no of the two states is the problem present.

---

This is probably not intended to be merged, at least not with the hard nix override in the second patch.

However, if we made that override an option, then we could merge this.

Right now this is just to present a publicly visible workaround for people to use, until there's an official `nix` release with the fix.

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -150,7 +150,10 @@ let
     then "/dev/" + builtins.substring 12 100 dev
     else dev;
 
-  nixosVersion = builtins.substring 0 5 (config.system.nixos.version or config.system.nixosVersion);
+  nixosVersion =
+    if cfg.nixosAmiVersion != null
+      then cfg.nixosAmiVersion
+      else builtins.substring 0 5 (config.system.nixos.version or config.system.nixosVersion);
 
   amis = import <nixpkgs/nixos/modules/virtualisation/ec2-amis.nix>;
 
@@ -233,6 +236,17 @@ in
       description = ''
         EC2 identifier of the AMI disk image used in the virtual
         machine.  This must be a NixOS image providing SSH access.
+      '';
+    };
+
+    deployment.ec2.nixosAmiVersion = mkOption {
+      example = "17.09";
+      type = types.nullOr types.str;
+      description = ''
+        NixOS version to choose AMIs from.
+        Has no effect if <option>ami</option> is set.
+        Useful if you want the base image to be a different NixOS version
+        than the system you will eventually deploy.
       '';
     };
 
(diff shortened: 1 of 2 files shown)


AmineChikhaoui:
I was just testing t2.nano/t2.micro instances with `nix.package = pkgs.nixUnstable` (Nix 2.1) on NixOS 18.03 and it seems to work just fine :)

@nh2 I'm going to close this PR as it seems that it's probably not relevant anymore, please feel free to re-open if you think I missed anything. Thanks

edolstra:
I don't think `nixosAmiVersion` solves the problem well because it only helps on the initial deployment. Any subsequent deployment will use Nix 2.0 from the new system, so it will have the same OOM issues.

However, I think `nixosAmiVersion` might be useful for other use cases, such as testing NixOS upgrades.

AmineChikhaoui:
@nh2 Ok right, seems the expression that I have didn't reproduce the initial issue, I was under the impression t2.nano/micro won't work at all even with a minimal expression but I guess I need to add few "large" packages to trigger the issue.

AmineChikhaoui:
and yeah most likely the issue will persist unless Nix is built and activated before copying the closure which doesn't happen.
Most likely we will need to wait for NixOS 18.09 to be officially released as I don't think shipping the ec2 AMIs with nixUnstable would be a good idea.

nh2:
> Any subsequent deployment will use Nix 2.0 from the new system, so it will have the same OOM issues.

I don't understand this argument. The idea here is that `deployment.ec2.nixosAmiVersion` would always be used in combination with specifying a patched version of nix that doesn't have the issue.

The initial boot from AMI will bring up a version of nix that is so old that it doesn't have the problem. The first deployment will then instal whatever version of nix that I specify, which of course would have the patches in to fix the issue.

At no point would a 2.0 version without the issue patched be installed on the target system.

coretemp:
Is this now obsolete? If so, please close this.

nh2:
> Is this now obsolete? If so, please close this.

No (please read the thread, this was already asked and answered above).

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #987, opened 2018-08-06
Add an argument to enable IP forwarding in GCE

Chakerbh:
Setting `deployment.gce.canIpForward` to **True** enable IP forwarding in the instance. IP forwarding can only be enabled when creating the instance for the first time. It can't be changed later on.


Chakerbh:
cc @AmineChikhaoui

PR #991, opened 2018-08-19
nix/auto-luks.nix: remove reference to auto-generated passphrase

eqyiel:
If this feature exists, it doesn't work right now due to the assertion here: https://github.com/NixOS/nixops/blob/cf47cfdf75c515967597ffd43019c5f32e7f9e41/nix/auto-luks.nix#L102

### DIFF
--- a/nix/auto-luks.nix
+++ b/nix/auto-luks.nix
@@ -44,16 +44,13 @@ with utils;
             default = "";
             type = types.str;
             description = ''
-              The passphrase (key file) used to decrypt the key to access
-              the volume.  If left empty, a passphrase is generated
-              automatically; this passphrase is lost when you destroy the
-              machine or underlying device, unless you copy it from
-              NixOps's state file.  Note that unless
+              The passphrase (key file) used to decrypt the key to access the
+              volume.  Note that unless
               <option>deployment.storeKeysOnMachine</option> is set to
-              <literal>false</literal>, the passphrase is stored in the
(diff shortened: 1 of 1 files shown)


AmineChikhaoui:
This feature exists and does work properly, the nixops backends (python parts) takes care of generating the passphrases and creating the physical spec with the appropriate block device mappings + passphrases. In EC2 backend for example setting `encrypt = true;` in the filesystem config will trigger the generation of a passphrase.

eqyiel:
@AmineChikhaoui thanks for pointing out that there's a difference between `deployment.autoLuks.<name>.passphrase` and `fileSystems.*.ec2.passphrase`.  

If I understand you correctly, I should be using `fileSystems.*.ec2` instead of editing `deployment.autoLuks.<name>.*` directly (because it gets passed there anyway)?

https://github.com/NixOS/nixops/blob/cf47cfdf75c515967597ffd43019c5f32e7f9e41/nix/ec2.nix#L465-L472

Even so, I think this change should be made because it's not possible to not set a passphrase if you use `deployment.autoLuks.<name>` directly.

That, is, this description should not change:
https://github.com/NixOS/nixops/blob/cf47cfdf75c515967597ffd43019c5f32e7f9e41/nix/ec2.nix#L115-L128

but this one should:
https://github.com/NixOS/nixops/blob/cf47cfdf75c515967597ffd43019c5f32e7f9e41/nix/auto-luks.nix#L43-L58

PR #993, opened 2018-08-24
Add a resource for GCE routes

Chakerbh:
An example of a route is 
```
  resources.gceRoutes.lb-default = {
    name = "lb-default";
    priority = 800;
    destination = "8.8.8.8/32";
    tags = [ "workers" ];
  };

```

### DIFF
--- a/nix/gce-routes.nix
+++ b/nix/gce-routes.nix
@@ -29,7 +29,8 @@ with import ./lib.nix lib;
 
       destination = mkOption {
         example ="1.1.1.1/32";
-        type = types.nullOr types.str;
+        type = types.nullOr (types.either types.str (resource "machine"));
+        apply = x: if x == null || (builtins.isString x) then x else "res-" + x._name;
         description = "The destination IP range that this route applies to. If the destination IP of a packet falls in this range, it matches this route.";
       };
 
@@ -43,7 +44,8 @@ with import ./lib.nix lib;
       nextHop = mkOption {
         default = null;
         example = "NAT-gateway";
-        type = types.nullOr types.str;
+        type = types.nullOr (types.either types.str (resource "machine"));
+        apply = x: if x == null || (builtins.isString x) then x else "res-" + x._name;
         description = "Next traffic hop, Leave it empty for the default internet gateway.";
       };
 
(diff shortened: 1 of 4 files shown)


nixops/resources/gce_route.py
```diff
@@ -0,0 +1,181 @@
...
+
+                if defn.nextHop.startswith("res-"):
+                    try:
+                        nextHop_name = self._get_machine_property(defn.nextHop[4:], "machine_name")
+                        defn.nextHop = self.connect().ex_get_node(nextHop_name)
+                    except AttributeError:
+                        raise Exception("nextHop can only be a GCE machine.")
+                        raise
+                    except libcloud.common.google.ResourceNotFoundError:
+                        raise Exception("The machine {0} isn't deployed, it need to be before it's added as nextHop".format(nextHop_name))
```
AmineChikhaoui: Can we use create_after method instead ?

nix/gce-routes.nix
```diff
@@ -0,0 +1,65 @@
...
+
+      network = mkOption {
+        example = "my-custom-network";
+        default = "default";
+        type = types.str;
+        description = "Name of the network, defaults to <literal>default</literal>.";
+      };
+
+      destination = mkOption {
+        example ="1.1.1.1/32";
```
AmineChikhaoui: nitpick: add space afer =

nix/gce-routes.nix
```diff
@@ -0,0 +1,65 @@
...
+        type = types.nullOr (types.either types.str (resource "machine"));
+        apply = x: if x == null || (builtins.isString x) then x else "res-" + x._name;
+        description = "The destination IP range that this route applies to. If the destination IP of a packet falls in this range, it matches this route.";
```
AmineChikhaoui: nitpick: can you use '' '' for long descriptions instead
```
description = ''
   foo
''
```

nixops/resources/gce_route.py
Chakerbh: I added create_after, but I'm keeping this in case we use --include the flag to only deploy the route. In that case the deployment the above exception will be raised.

PR #1003, opened 2018-09-13
Take NIX_SSHOPTS into account

yorickvP:


yorickvP:
Envisioned usage:
```
nixops deploy --include builder resources
NIX_SSHOPTS="-J $(nixops private-ip builder)" nixops deploy --exclude builder
```

PR #1006, opened 2018-09-14
support ipv6 for scp with "none" backend

phryneas:
Motivation:

Currently, secrets cannot be deployed when `deployment.targetHost` contains an ipv6.

This is because, while sshing to a ipv6 has the same syntax, `scp` requires square brackets around the ip. This adds square brackets for calls to scp around any ip with a colon.

I'm no python developer, so please take this PR with a grain of salt. Fortunately, it's not very big ;)

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -271,7 +271,7 @@ class MachineState(nixops.resources.ResourceState):
     def get_keys(self):
         return self.keys
 
-    def get_ssh_name(self, scp=False):
+    def get_ssh_name(self):
         assert False
 
     def get_ssh_flags(self, scp=False):
@@ -390,7 +390,7 @@ class MachineState(nixops.resources.ResourceState):
         cmdline = ["scp"] + self.get_ssh_flags(True) + master.opts
         if recursive:
             cmdline += ['-r']
-        cmdline += [source, "root@" + self.get_ssh_name(True) + ":" + target]
+        cmdline += [source, "root@" + self.get_ssh_name() + ":" + target]
         return self._logged_exec(cmdline)
 
     def download_file(self, source, target, recursive=False):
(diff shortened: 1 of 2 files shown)


phryneas:
Looks like another `get_ssh_name` is being called - I didn't account for the other backends. Are you definitely using the ǹone` backend?

flokli:
I observed the regression with the hetzner provider, didn't test deploying secrets with the 'none' provider so far.

phryneas:
Ah, okay, I need to update the signature to `def get_ssh_name(self, scp=False):` in the other providers, too, then.

I'll update this PR in the next days. In the meantime, just update `def get_ssh_name(self):` in the hetzner provider to `def get_ssh_name(self, scp=False):`, that should solve it for you for now.

flokli:
> In the meantime, just update `def get_ssh_name(self):` in the hetzner provider to `def get_ssh_name(self, scp=False):`, that should solve it for you for now.

Looking at the other implementations of `get_ssh_name`, shouldn't the `[%s]` wrapping logic better be placed into a helper function in `nixops/backends/__init__.py`, used by both `upload_file` and `download_file`, instead of passing that `ssh` parameter through all backends, and applying logic there?

phryneas:
Yup, I had the same thought. I'm quite busy right now, but I'll try to update this PR in the next weeks. It's not like it hasn't been lying around for a while already ;)

flokli:
Sure, whenever you find the time :-)

flokli:
@phryneas - poke, could you update this PR?

phryneas:
@flokli oops, sorry. There you go.

flokli:
LGTM.

CI is red due to the current nixpkgs breakage with azure-mgmt-compute-4.4.0 (unrelated, fixed by https://github.com/NixOS/nixpkgs/pull/56775).

PR #1011, opened 2018-09-23
Support persistent spot requests

Chakerbh:
This means when the instance gets interrupted, the request is still
valid, and if the termination behaviour is set to "stop", the instance
will start automatically when the request get fulfilled again by AWS.

There's no way to achieve this with Boto2, so I changed the call that
create the spot request to Boto3. I did some refactoring, so the same
call is used when creating both On demand and spot instances.

an example:

```
{
    region ? "us-east-1"
    , accessKeyId ? "AKIA...."
, ...
}:
{
machine = {resources, ...}:
  { 
      deployment.targetEnv = "ec2";
      deployment.ec2.accessKeyId = accessKeyId;
      deployment.ec2.region = "us-east-1";
      deployment.ec2.instanceType = "m3.large";
      deployment.ec2.keyPair = resources.ec2KeyPairs.keypair;


      deployment.ec2.spotInstancePrice = 240;
      deployment.ec2.spotInstanceRequestType = "persistent";
      deployment.ec2.spotInstanceInterruptionBehavior = "stop";


      deployment.ec2.tags  = {
          Name = "persistant_spot_instance";
      };
      deployment.ec2.securityGroups = ["admin" "default"];
      fileSystems."/data" = {
        fsType = "xfs";
        device =  "/dev/xvdf";
        autoFormat = true;
        ec2.size = 5;
        ec2.volumeType = "gp2";
      };
    };
  resources.ec2KeyPairs.keypair = { inherit region accessKeyId; };
}
```

### DIFF
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -390,6 +390,23 @@ in
       '';
     };
 
+    deployment.ec2.spotInstanceRequestType = mkOption {
+      default = "one-time";
+      type = types.str;
+      description = ''
+        The type of the spot instance request. It can be either "one-time" or "persistent".
+      '';
+    };
+
+    deployment.ec2.spotInstanceInterruptionBehavior = mkOption {
+      default = "terminate";
+      type = types.str;
+      description = ''
+        Whether to terminate, stop or hibernate the instance when it gets interrupted.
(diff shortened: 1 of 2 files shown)


AmineChikhaoui:
Closing as the work is picked up in #1102

Issue #696, opened 2017-06-29
StrictHostKeyChecking=no is insecure


elitak:
The hardcoded `-o StrictHostKeyChecking=no` everywhere is a big SecOps no-no. It's quite feasible an attacker could wind up with an IP address you neglect to change after relinquishing, and have an entire host config hand-delivered to him to inspect for vulnerabilities. He wouldn't be able to MITM the the deployment, but obtaining what is essentially a dump of the host's whole filesystem is still pretty disastrous, from a defensive standpoint.

Ideallly, as @oconnorr suggested in #262, nixops would have its own known_hosts mapping cached, but short of that, I'd like to at least have the option of turning on StrictHostKeyChecking=yes permanently somehow. It really should be at least =ask by default, too, but I expect handling the user interaction could cause some grief. I'd much rather have to `ssh-keygen -R $host; ssh $host :` every time a host changes its keys, than compromise a big part of ssh's integrity for a small convenience.

domenkozar:
For context:

```
     StrictHostKeyChecking
	     If	this flag is set to ``yes'', ssh(1) will never automatically
	     add host keys to the ~/.ssh/known_hosts file, and refuses to con-
	     nect to hosts whose host key has changed.	This provides maximum
	     protection	against	trojan horse attacks, though it	can be annoy-
	     ing when the /etc/ssh/ssh_known_hosts file	is poorly maintained
	     or	when connections to new	hosts are frequently made.  This
	     option forces the user to manually	add all	new hosts.  If this
	     flag is set to ``no'', ssh	will automatically add new host	keys
	     to	the user known hosts files.  If	this flag is set to ``ask'',
	     new host keys will	be added to the	user known host	files only
	     after the user has	confirmed that is what they really want	to do,
	     and ssh will refuse to connect to hosts whose host	key has
	     changed.  The host	keys of	known hosts will be verified automati-
	     cally in all cases.  The argument must be ``yes'',	``no'',	or
	     ``ask''.  The default is ``ask''.
```

elitak:
Incidentally, the wording of that paragraph is quite misleading, especially: `The host keys of known hosts will be verified automatically in all cases.`. That sounds to me like ssh will refuse to connect on a mismatch with the setting =no, but it does anyway. Only password auth and agent forwarding are forcefully disabled when there's a mismatch.

[...]

PR #1023, opened 2018-10-20
Partially mitigate StrictHostKeyChecking=no issue

aszlig:
From issue #696:

> The hardcoded -o StrictHostKeyChecking=no everywhere is a big SecOps no-no. It's quite feasible an attacker could wind up with an IP address you neglect to change after relinquishing, and have an entire host config hand-delivered to him to inspect for vulnerabilities. He wouldn't be able to MITM the the deployment, but obtaining what is essentially a dump of the host's whole filesystem is still pretty disastrous, from a defensive standpoint.

I by myself have been guilty of using this (added this to the Hetzner backend), because I did actually misunderstand the meaning of setting this option to `no`. My understanding was that it will refuse to connect whenever an existing host key is different from that in known hosts.

However, this turns out to be only true for keyboard interactive or password authentication and if we're using pubkey auth, OpenSSH will happily connect.

Now the real fix for this (already deploying with a pre-generated host key) is a bit more involved, but we can mitigate this for now, because since OpenSSH 7.5 there is the `accept-new` option to `StrictHostKeyChecking`, which does exactly what I thought "no" would do:

 >If this flag is set to ``accept-new'' then ssh will automatically add new host keys to the user known hosts files, but will not permit connections to hosts with changed host keys.

### DIFF
--- a/nix/ssh-tunnel.nix
+++ b/nix/ssh-tunnel.nix
@@ -87,7 +87,7 @@ with lib;
         remoteCommand = mkAddrConf v.remoteTunnel v.remoteIPv4 v.localIPv4;
 
       in "ssh -i ${v.privateKey} -x"
-       + " -o StrictHostKeyChecking=no"
+       + " -o StrictHostKeyChecking=accept-new"
        + " -o PermitLocalCommand=yes"
        + " -o ServerAliveInterval=20"
        + " -o LocalCommand='${localCommand}'"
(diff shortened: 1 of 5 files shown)


aszlig:
@edolstra, @rbvermaa: Is there a case where we need to rely on OpenSSH outside of nixpkgs?

kampka:
I am much happier with this than with the original behavior.
Since this is still not a partial fix, I think it would be a good idea to add some documentation that this is happening and add a quick warning / howto for security conscious users.
At least no one is going to be surprised about it then.

edolstra:
@aszlig No, don't think so.

PR #1024, opened 2018-10-21
Allow specifying arbitrary SSH configuration for nodes

grahamc:
I'd like to deploy to nodes only available behind a bastion server. This patch letsyou do this:

```
  my-node.deployment.sshConfigOptionsFile = ./ssh-config;
```

and ./ssh-config:

```
ProxyCommand ssh 10.5.3.111 -A -W %h:%p
ForwardAgent yes
LogLevel VERBOSE
```

Initial testing shows this works, I'll do further testing tomorrow.

### DIFF
--- a/nixops/backends/hetzner.py
+++ b/nixops/backends/hetzner.py
@@ -186,12 +186,11 @@ class HetznerState(MachineState):
             # so only wait for the reboot to finish when deploying real
             # systems.
             self.log_start("waiting for rescue system...")
-            while self.try_ssh():
-                self.log_continue(".")
+            dotlog = lambda: self.log_continue(".")  # NOQA
+            self.wait_for_ssh(callback=dotlog, up=False)
             self.log_continue("[down]")
 
-            while not self.try_ssh():
-                self.log_continue(".")
+            self.wait_for_ssh(callback=dotlog, up=True)
             self.log_end("[up]")
         self.state = self.RESCUE
 
--- a/nixops/backends/hetzner.py
+++ b/nixops/backends/hetzner.py
@@ -9,7 +9,6 @@ import subprocess
 from hetzner.robot import Robot
 
 from nixops import known_hosts
-from nixops.util import wait_for_tcp_port, ping_tcp_port
 from nixops.util import attr_property, create_key_pair, xml_expr_to_python
 from nixops.ssh_util import SSHCommandFailed
 from nixops.backends import MachineDefinition, MachineState
@@ -187,10 +186,12 @@ class HetznerState(MachineState):
             # so only wait for the reboot to finish when deploying real
             # systems.
             self.log_start("waiting for rescue system...")
-            dotlog = lambda: self.log_continue(".")  # NOQA
-            wait_for_tcp_port(ip, 22, open=False, callback=dotlog)
+            while self.try_ssh():
+                self.log_continue(".")
             self.log_continue("[down]")
-            wait_for_tcp_port(ip, 22, callback=dotlog)
+
+            while not self.try_ssh():
+                self.log_continue(".")
             self.log_end("[up]")
         self.state = self.RESCUE
 
@@ -650,7 +651,9 @@ class HetznerState(MachineState):
         """
         self.log_start("waiting for system to shutdown... ")
         dotlog = lambda: self.log_continue(".")  # NOQA
-        wait_for_tcp_port(self.main_ipv4, 22, open=False, callback=dotlog)
+        while self.try_ssh():
+            self.log_continue(".")
+
         self.log_continue("[down]")
 
         self.state = self.STOPPED
@@ -684,7 +687,7 @@ class HetznerState(MachineState):
             return
 
         if self.state in (self.STOPPED, self.STOPPING):
-            res.is_up = ping_tcp_port(self.main_ipv4, 22)
+            res.is_up = self.try_ssh()
             if not res.is_up:
                 self.state = self.STOPPED
                 res.is_reachable = False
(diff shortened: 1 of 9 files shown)


grahamc:
I can now force reboot nodes and have it wait properly for the machine to be up. A couple more TCP waiting patches to make.

coretemp:
Are you sure that this also handles the ssh-for-each case? If so, please merge. If not, please try again :)

grahamc:
```
$ nixops ssh-for-each hostname
mac7........> mac7
mac6........> mac6
mac5........> mac5
mac4........> mac4
mac3........> mac3
mac2........> mac2
mac1........> mac1
mac8........> mac8
```

grahamc:
OK I think this is good to go:

```
$ nixops deploy --force-reboot --include mac1 
mac1........> copying closure...
personal> closures copied successfully
mac1........> rebooting...
mac1........> waiting for the machine to finish rebooting...[down]channel 0: open failed: connect failed: Connection refused
stdio forwarding failed
ssh_exchange_identification: Connection closed by remote host

mac1........> could not connect to ‘root@192.168.2.101’, retrying in 1 seconds...
mac1........> activation finished successfully
personal> deployment finished successfully
```

nixops/ssh_util.py
```diff
@@ -299,5 +299,28 @@ def run_command(self, command, flags=[], timeout=None, logged=True,
...
+
+
+    def wait_for_ssh(self, user=None, attempts=-1, timeout=5, callback=None):
+        """Wait until the remote's SSH is up."""
+        n = 0
+        while True:
+            try:
+                self.run_command('true', timeout=5, user=user, flags=['-q'])
+                return True
+            except nixops.ssh_util.SSHConnectionFailed:
```
aszlig: Why not re-use `try_ssh` here instead of doing roughly the same thing here again?

nixops/ssh_util.py
```diff
@@ -299,5 +299,28 @@ def run_command(self, command, flags=[], timeout=None, logged=True,
             else:
                 return res
 
+    def try_ssh(self, user=None, timeout=-1):
```
aszlig: The `timeout` keyword argument is unused here and later hardcoded to `1`.

nixops/backends/hetzner.py
```diff
@@ -650,7 +651,9 @@ def _wait_stop(self):
         """
         self.log_start("waiting for system to shutdown... ")
         dotlog = lambda: self.log_continue(".")  # NOQA
-        wait_for_tcp_port(self.main_ipv4, 22, open=False, callback=dotlog)
+        while self.try_ssh():
```
aszlig: This is a bit racy, because the timeout for `run_command` is only the SSH connect timeout, so if the machine is no longer reachable after the key exchange `try_ssh` could hang for minutes.

Btw. the same problem could occur in https://github.com/NixOS/nixops/pull/857, hence cc @nh2.

nixops/backends/hetzner.py
```diff
@@ -187,10 +186,12 @@ def _wait_for_rescue(self, ip):
             # so only wait for the reboot to finish when deploying real
             # systems.
             self.log_start("waiting for rescue system...")
-            dotlog = lambda: self.log_continue(".")  # NOQA
-            wait_for_tcp_port(ip, 22, open=False, callback=dotlog)
+            while self.try_ssh():
+                self.log_continue(".")
```
aszlig: All these constructs sound a bit repetitive, what about refactoring it so that we can use a single `wait_for_ssh` instead?

asymmetric:
@grahamc are you still planning on working on this?

grahamc:
The commits since 0192a61182c96ae3e3e38c66aa5b93aa9c7727e2 address almost all the feedback from @aszlig, but I'm not sure what to do about the timeout you raise.

Additionally, I no longer require this PR and am actually not even able to really test it as I don't have any bastion use cases.

dhess:
I'll give this a try sometime in the next few weeks. I plan to put my new deployments behind a bastion host.

nixops/backends/hetzner.py
aszlig: If we'd switch to Python 3.x (I think it's long overdue), we could use [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) with the `timeout` argument for `logged_exec` and in `SSHMaster`, but in the interim it's still possible in Python 3.4 and lower using [`wait`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) and [`kill`](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.kill). We already have a `timeout` argument, so I'd either rename the argument or add another `command_timeout` argument. Note that we can't simply get rid of the connect timeout and just use the command timeout, because there might be long-running operations, like switching to a new configuration.

dhess:
The bastion functionality works perfectly, so that's very nice, as I couldn't get this to work without the patch (NixOps hangs on `waiting for SSH...`).

However, I did have one case where `nixops reboot` did not detect that the host had come back up, so this seems to confirm @aszlig's comment about the potential race condition. (I have had at least one other reboot work fine, however.)

peterhoeg:
I'd love this functionality as well. Good to merge with @dhess's blessing?

dhess:
Unfortunately, I think it's not ready. `nixops reboot` frequently loses track of the rebooted host.

I may have time to look into a fix for this in the next few weeks.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

wmertens:
:laughing:  Fun to see the conversation between you and you @grahamc ;)

Issue #833, opened 2018-01-11
Deploy disconnected my Hetzner machine again


nh2:
Similar to #640, but this time I didn't add any extra package myself like `dnsmasq` in that ticket; this time I only updated packages to a newer version on top of `release-17.09` (from `a2845aa0` to `df6f8d4c`).

In nixops I see

```
...
node-3............> updating GRUB 2 menu...
node-3............> stopping the following units: alsa-store.service, audit.service, consul.service, consulAfterTinc.service, cpufreq.service, dnsmasq.service, glusterVolumeRunningCheckWatchdog.service, glusterd.service, glustereventsd.service, kmod-static-nodes.service, network-link-eth0.service, network-local-commands.service, network-setup.service, nix-daemon.service, nix-daemon.socket, nscd.service, rngd.service, stolon-keeper.service, stolon-proxy.service, stolon-sentinel.service, systemd-journal-catalog-update.service, systemd-modules-load.service, systemd-sysctl.service, systemd-timesyncd.service, systemd-tmpfiles-clean.timer, systemd-tmpfiles-setup-dev.service, systemd-udev-trigger.service, systemd-udevd-control.socket, systemd-udevd-kernel.socket, systemd-udevd.service, tinc.myvpn.service, my-gluster-files-mount.service
node-3............> NOT restarting the following changed units: getty@tty1.service, systemd-backlight@backlight:acpi_video0.service, systemd-journal-flush.service, systemd-logind.service, systemd-random-seed.service, systemd-remount-fs.service, systemd-tmpfiles-setup.service, systemd-update-utmp.service, systemd-user-sessions.service, user@0.service
node-3............> activating the configuration...
node-3............> warning: user ‘postgres’ has unknown group ‘postgres’
node-3............> setting up /etc...
node-3............> restarting systemd...
node-3............> setting up tmpfiles
node-3............> reloading the following units: dbus.service, dev-hugepages.mount, dev-mqueue.mount, firewall.service, sys-fs-fuse-connections.mount, sys-kernel-debug.mount
node-3............> restarting the following units: network-addresses-eth0.service, nginx.service, sshd.service, systemd-journald.service
Timeout, server 1.2.3.4 not responding.
```

CC @basvandijk, @FPtje, @aszlig, @domenkozar, @fpletz, @groxxda

basvandijk:
Bummer!

[...]

PR #1032, opened 2018-10-28
hetzner: Don't use localCommands, set interface device of default gateway

nh2:
Should help #833.

Original localCommands patch by aszlig.
Fixups and interface device stuff from me.

Note that this new way of doing it without localCommands is backwards
incompatible with NixOS < 18.03, see:
https://nixos.org/nixos/manual/release-notes.html#sec-release-18.03-notable-changes

As described on

    https://wiki.hetzner.de/index.php/Netzkonfiguration_Debian/en#IPv4

Hetzner requires that even traffic to the local subnet goes via the gateway.

NixOps already had code for that (set in `localCommands`), but it wasn't enough:
The

    ip route replace default  via "GATEWAY_IP"  proto static

generated by `network-setup.service` in NixOS's `network-interface-scripted.nix`
fails with

    RTNETLINK answers: Network is unreachable

because all routes added so far had `via "GATEWAY_IP"`, but
the kernel didn't know how to actually reach GATEWAY_IP because
there was no static route for that IP, see.

    https://serverfault.com/questions/581159/unable-to-add-a-static-route-sioaddrt-network-is-unreachable/581162#581162

This commit adds an explicit static route to the default gateway, on the
relevant network device.

This allows us to generate the following instead:

    ip route replace GATEWAY_IP dev THE_DEVICE proto static
    ip route replace default  via "GATEWAY_IP" dev THE_DEVICE proto static

so that the kernel knows how to reach the gateway.

An example of what routes look like in `ip route` after deploying and 18.03
machine, where 1.2.3.4 is the server's IP and 1.2.3.1 is the gateway:

    default via 1.2.3.1 dev eth0                         src 1.2.3.4 metric 202
    1.2.3.0/27          dev eth0 proto kernel scope link src 1.2.3.4 metric 202
    1.2.3.1             dev eth0 proto static scope link

For IPv6, the link-local `fe80::1` is the gateway, as described on

    https://wiki.hetzner.de/index.php/Netzkonfiguration_Debian/en#IPv6

### DIFF
--- a/nixops/backends/hetzner.py
+++ b/nixops/backends/hetzner.py
@@ -464,9 +464,12 @@ class HetznerState(MachineState):
         """
         Return the default gateway of the currently running machine.
         """
-        cmd = "ip route list | sed -n -e 's/^default  *via  *//p'"
-        cmd += " | cut -d' ' -f1"
-        return self.run_command(cmd, capture_stdout=True).strip()
+        default_gw_cmd = "ip route list | sed -n -e 's/^default  *via  *//p'"
+        default_gw_output = self.run_command(default_gw_cmd, capture_stdout=True).strip()
+        default_gw_output_split = default_gw_output.split(' ')
+        gw_ip = default_gw_output_split[0]
+        gw_dev = default_gw_output_split[2]
+        return (gw_ip, gw_dev)
 
     def _get_nameservers(self):
         """
@@ -499,13 +502,11 @@ class HetznerState(MachineState):
         """
         udev_rules = []
         iface_attrs = {}
-        extra_routes = []
-        ipv6_commands = []
 
         server = self._get_server_by_ip(self.main_ipv4)
 
         # Global networking options
-        defgw = self._get_default_gw()
+        defgw_ip, defgw_dev = self._get_default_gw()
         v6defgw = None
 
         # Interface-specific networking options
@@ -519,41 +520,52 @@ class HetznerState(MachineState):
 
             udev_rules.append(self._get_udev_rule_for(iface))
 
-            ipv4, prefix = result
+            ipv4addr, prefix = result
             iface_attrs[iface] = {
-                'ipAddress': ipv4,
-                'prefixLength': int(prefix),
+                'ipv4': {
+                    'addresses': [
+                        {'address': ipv4addr, 'prefixLength': int(prefix)},
+                    ],
+                },
             }
 
             # We can't handle Hetzner-specific networking info in test mode.
             if TEST_MODE:
                 continue
 
-            # Extra route for accessing own subnet
-            net = self._calculate_ipv4_subnet(ipv4, int(prefix))
-            extra_routes.append(("{0}/{1}".format(net, prefix), defgw, iface))
+            # Extra route for accessing own subnet for this interface
+            # (see https://wiki.hetzner.de/index.php/Netzkonfiguration_Debian/en#IPv4),
+            # but only if it's not the interface for the default gateway,
+            # because that one will already get such a route generated
+            # by NixOS's `network-setup.service`. See also:
+            #   https://github.com/NixOS/nixops/pull/1032#issuecomment-433741624
+            if iface != defgw_dev:
+                net = self._calculate_ipv4_subnet(ipv4addr, int(prefix))
+                iface_attrs[iface]['ipv4'] = {
+                    'routes': [{
+                        'address': net,
+                        'prefixLength': int(prefix),
+                        'via': defgw_ip,
+                    }],
+                }
 
-            # IPv6 subnets only for eth0 (XXX: more flexibility here?)
-            v6addr_command = "ip -6 addr add '{0}' dev '{1}' || true"
+            # IPv6 subnets only for eth0
+            v6subnets = []
             for subnet in server.subnets:
                 if "." in subnet.net_ip:
                     # skip IPv4 addresses
                     continue
-                v6addr = "{0}/{1}".format(subnet.net_ip, subnet.mask)
-                ipv6_commands.append(v6addr_command.format(v6addr, iface))
-                assert v6defgw is None or v6defgw == subnet.gateway
-                v6defgw = subnet.gateway
-
-        # Extra routes
-        route4_cmd = "ip -4 route change '{0}' via '{1}' dev '{2}' || true"
-        route_commands = [route4_cmd.format(network, gw, iface)
-                          for network, gw, iface in extra_routes]
-
-        # IPv6 configuration
-        route6_cmd = "ip -6 route add default via '{0}' dev eth0 || true"
-        route_commands.append(route6_cmd.format(v6defgw))
-
-        local_commands = '\n'.join(ipv6_commands + route_commands) + '\n'
+                v6subnets.append({
+                    'address': subnet.net_ip,
+                    'prefixLength': int(subnet.mask)
+                })
+                assert (v6defgw is None or
+                        v6defgw.get('address') == subnet.gateway)
+                v6defgw = {
+                    'address': subnet.gateway,
+                    'interface': defgw_dev,
+                }
+            iface_attrs[iface]['ipv6'] = { 'addresses': v6subnets }
 
         self.net_info = {
             'services': {
@@ -561,9 +573,12 @@ class HetznerState(MachineState):
             },
             'networking': {
                 'interfaces': iface_attrs,
-                'defaultGateway': defgw,
+                'defaultGateway': {
+                    'address': defgw_ip,
+                    'interface': defgw_dev,
+                },
+                'defaultGateway6': v6defgw,
                 'nameservers': self._get_nameservers(),
-                'localCommands': local_commands,
             }
         }
 


nh2:
CC @basvandijk @aszlig @fpletz

basvandijk:
Good stuff!

nh2:
I see one last issue in `journalctl` when the server boots:

```
Oct 28 20:00:14 builder systemd[1]: Starting Address configuration of eth0...
Oct 28 20:00:14 builder network-addresses-eth0-start[910]: adding address 2a01:123:123:123::/64... done
Oct 28 20:00:14 builder network-addresses-eth0-start[910]: adding route 1.2.3.4/27... failed
Oct 28 20:00:14 builder systemd[1]: network-addresses-eth0.service: Main process exited, code=exited, status=1/FAILURE
Oct 28 20:00:14 builder systemd[1]: network-addresses-eth0.service: Failed with result 'exit-code'.
Oct 28 20:00:14 builder systemd[1]: Failed to start Address configuration of eth0.
```

nix/hetzner-bootstrap.nix
```diff
@@ -5,7 +5,6 @@ let
 
   nixpart = python2Packages.nixpart0.override {
     useNixUdev = false;
-    udevSoMajor = 0;
```
aszlig: Ah, that's actually something I wanted to break up into a separate commit, but it was left in that stash.

nixops/backends/hetzner.py
```diff
@@ -464,9 +464,12 @@ def _get_default_gw(self):
...
         Return the default gateway of the currently running machine.
         """
-        cmd = "ip route list | sed -n -e 's/^default  *via  *//p'"
-        cmd += " | cut -d' ' -f1"
-        return self.run_command(cmd, capture_stdout=True).strip()
+        default_gw_cmd = "ip route list | sed -n -e 's/^default  *via  *//p'"
+        gw_ip_cmd = default_gw_cmd + " | cut -d' ' -f1"
+        gw_dev_cmd = default_gw_cmd + " | cut -d' ' -f3"
+        gw_ip = self.run_command(gw_ip_cmd, capture_stdout=True).strip()
+        gw_dev = self.run_command(gw_dev_cmd, capture_stdout=True).strip()
```
aszlig: Probably would make sense to use only one single command using `spli`t instead of the `cut`.

tests/hetzner-backend/default.nix
```diff
@@ -1,7 +1,7 @@
 { system, nixops }:
 
 with import <nixpkgs/nixos/lib/testing.nix> { inherit system; };
-with import <nixpkgs/nixos/lib/qemu-flags.nix>;
+with import <nixpkgs/nixos/lib/qemu-flags.nix> { inherit pkgs; };
```
aszlig: Same here...

aszlig:
@nh2: Did you test whether this works on a server with additional subnets?

nh2:
@aszlig I think the

```
    "interfaces": {
      "eth0": {
        "ipv4": {
          "routes": [
            {
              "via": "1.2.3.1",
              "prefixLength": 27,
              "address": "1.2.3.0"
            }
          ]
        },
```

that we're generating now is unnecessary.

It gets placed into `network-addresses-eth0` and results in the `adding route 1.2.3.4/27... failed` from above with `RTNETLINK answers: Network is unreachable`:

```
[root@builder:~]# cat $(systemctl cat network-addresses-eth0.service | grep ExecStart | cut -d= -f2)
#! /nix/store/lw7xaqhakk0i1c631m3cvac3x4lc5gr5-bash-4.4-p12/bin/bash -e
state="/run/nixos/network/addresses/eth0"
mkdir -p $(dirname "$state")

echo "1111:1111:1111:1111::/64" >>  $state
echo -n "adding address 1111:1111:1111:1111::/64... "
if out=$(ip addr add "1111:1111:1111:1111::/64" dev "eth0" 2>&1); then
  echo "done"
elif ! echo "$out" | grep "File exists" >/dev/null 2>&1; then
  echo "failed: $out"
  exit 1
fi


state="/run/nixos/network/routes/eth0"
mkdir -p $(dirname "$state")

echo "1.2.3.0/27" >> $state
echo -n "adding route 1.2.3.0/27... "
if out=$(ip route add "1.2.3.0/27"  via "1.2.3.1" dev "eth0" proto static 2>&1); then
  echo "done"
elif ! echo "$out" | grep "File exists" >/dev/null 2>&1; then
  echo "failed: $out"
  exit 1
fi
```

This is because `network-addresses-eth0.service` runs before `network-setup.service` can set up the direct route to the gateway -- it does that already in https://github.com/NixOS/nixpkgs/blob/6f07d27625f28219095c273b97ded3cc416e9d46/nixos/modules/tasks/network-interfaces-scripted.nix#L120.

aszlig:
@rbvermaa, @edolstra: Do we care about supporting NixOS versions older than 18.03? Or what is the general policy on which NixOS versions we support?

aszlig:
@nh2: Also, what's missing here (and what I wanted to implement when testing this) is a migration function that migrates the old `net_info` into the new one. This of course would need additional testing.

nixops/backends/hetzner.py
nh2: Done

nix/hetzner-bootstrap.nix
nh2: Done

tests/hetzner-backend/default.nix
nh2: Done

nh2:
I've pushed a change 

> It gets placed into `network-addresses-eth0` and results in the `adding route 1.2.3.4/27... failed` from above with `RTNETLINK answers: Network is unreachable`:

I've pushed a fix for that:

Only generate the route if it's not for the device that's the default gateway.

nh2:
> Also, what's missing here (and what I wanted to implement when testing this) is a migration function that migrates the old `net_info` into the new one.

@aszlig I've written an ad-hoc one that goes via `sqlite3` and `jq` that I used to migrate my servers.

Perhaps the logic in there is useful to some extent for making the proper migration script.

https://gist.github.com/nh2/7707e0069df1d4bebc6572c40c85b673

nh2:
@flokli In general I agree that there's no point not to set netmask `/32` given what Hetzner writes.

I considered doing that, but then opted to first make the _minimal_ change in this PR that fixes the issue.

I would recommend implementing the `/32` chance as a separate PR on top of this, unless anybody can come up with whether `/32` would make any problems.

flokli:
I was suggesting the /32 for this PR, as we started discussing about migrations, and then we'd probably need to worry about that a second time, when changing to the /32.

I run a such a network configuration on some Hetzner servers for quite a while now and consider it much less flaky.

nh2:
@flokli This PR doesn't do migrations, I'd probably add that with a separate PR as well.

I care first about that _new_ Hetzner deployments work, and second that old ones can be migrated.

I am hesitant to auto-migrate nixops Hetzner users from `/27` to `/32` because who knows what their network config is.

There must be _some_ reason why Hetzner doens't just set it up as `/32` -- perhaps it's for the case where you buy multiple IPs for multiple servers, and put a separate switch in between (they have a service for that)?

Again, I'm all for the `/32` approach if nobody comes up with a counter case, but it would ideally not hold back this PR just so that we have some solution that works with NixOS out of the box without losing network connectivity now and then, as soon as possible.

flokli:
> I am hesitant to auto-migrate nixops Hetzner users from `/27` to `/32` because who knows what their network config is.
Same should apply for not auto-migrating existing deployments without asking back ;-)
 
> There must be _some_ reason why Hetzner doens't just set it up as `/32` -- perhaps it's for the case where you buy multiple IPs for multiple servers, and put a separate switch in between (they have a service for that)?
I guess it a) grew historically b is easy to setup on some systems, and even when not adding the `/26` or `/27` route, most of the internet still works :laughing: 

Additional IPv4 adresses are routed via the main IPv4 adress, so you can either add it with netmask /32 on any interface, or route it somewhere else internally (for example, as a peer route via a tunnel link)

> Again, I'm all for the `/32` approach if nobody comes up with a counter case, but it would ideally not hold back this PR just so that we have some solution that works with NixOS out of the box without losing network connectivity now and then, as soon as possible.
Fine, so let's merge this, and add some migration assistance and the `/32` routing as two separate PRs.

aszlig:
@flokli: I tested this with a new server just now and routing to the same subnet doesn't work as it's missing the gateway. So I guess I'll add the `/32` prefix to make sure that the machine always uses the gateway instead of sending directly, that way we shouldn't need the extra route.

aszlig:
@flokli: Actually, never mind about the `doesn't work`... because I've tested this against another host that was misconfigured.

nh2:
This came up for me yesterday.

## NixOps Hetnzer machines currently are generated wrong, failing to ping machines near them

NixOps currently still uses a "normal" subnet for Hetzner (e.g. `/24`) because PR #1070 (to change it to `/32`) was not merged (auto-closed because of the NixOps plugin rearchitecture). This causes Hetzner machines deployed by NixOps to be unable to ping Hetzner machines that are on the same subnet (e.g. probably in the same rack).

## Current state

@aszlig kindly sat down with me and together we embarked on a quest to find out what NixOS and NixOps should best do. I post here our notes for posterity.

First, a summary of what NixOS and NixOps currently do:

* NixOps determines the IP, subnet size, and gateway, from the Hetzner Rescue System, e.g. IP/subnet `1.2.3.4/24` and gateway `1.2.3.1`.
  It generates entries accordingly the NixOS config:
  ```nix
  networking.interfaces.eth0.ipv4.addresses = [ { address = "1.2.3.4`; prefixLength = 24; } ]`
  networking.defaultGateway = "1.2.3.1";
  ```
* NixOS:
  * Generates the `ip addr` command
    ```sh
    ip addr replace "1.2.3.4/24" dev "eth0"
    ```
    based on [`networking.interfaces.<name>.ipv4.addresses`](https://search.nixos.org/options?channel=24.11&show=networking.interfaces.%3Cname%3E.ipv4.addresses&from=0&size=50&sort=relevance&type=packages&query=networking.interfaces.%3Cname%3E.ipv4.addresses) in [`network-interfaces-scripted.nix`](https://github.com/NixOS/nixpkgs/blame/b134951a4c9f3c995fd7be05f3243f8ecd65d798/nixos/modules/tasks/network-interfaces-scripted.nix#L206).
  * Generates the `ip route` commands
    ```sh
    # Set the default gateway.
    ip route replace 1.2.3.1 dev eth0  proto static

    ip route replace default  via "1.2.3.1"  dev eth0 proto static
    ```
    based on the global (that is, not per-IP-address) [`networking.defaultGateway`](https://search.nixos.org/options?channel=24.11&from=0&size=50&sort=relevance&type=packages&query=defaultGateway) in [`network-interfaces-scripted.nix`](https://github.com/NixOS/nixpkgs/blame/b134951a4c9f3c995fd7be05f3243f8ecd65d798/nixos/modules/tasks/network-interfaces-scripted.nix#L131-L138) into `network-setup.service`. This was first added by @fpletz [here](https://github.com/NixOS/nixpkgs/commit/d0435ba032473f89fad8b6425bb3feee3d6fd718#diff-344943ac406571508ecc8d0311e237519c34548ffaf4eba83a15758a1ced65a4R126).

This current behaviour is wrong because if the machine tries to contact `1.2.3.2`, the packets will be sent directly to the local network (switch), instead of to the router (which Hetzner [requires](https://docs.hetzner.com/robot/dedicated-server/network/net-config-debian-ubuntu/#ipv4)).

The Hetzner docs give as an example

```
iface eth0 inet static
  address 192.168.0.250
  netmask 255.255.255.255
  gateway 192.168.0.1
  pointopoint 192.168.0.1
```

In PR

* https://github.com/NixOS/nixops/pull/1070

@aszlig proposed as a fix that NixOps should instead generate `prefixLength = 32`.

## Equivalence of `/32`-pointopoint versus `/32` with route?

The question that arises from that is whether that is fully equivalent to the `pointopoint` that Hetzner docs suggest; that is, is

```
ip addr add 1.2.3.4/32 dev eth0
ip route add 1.2.3.1 dev eth0
```

equivalent to Hetzner's suggested

```
ip addr add 1.2.3.4/32 peer 1.2.3.1 dev eth0
```

?

I also posted that question here: https://unix.stackexchange.com/questions/793114/difference-between-pointopoint-32-and-32-with-added-route

(Note `peer` in `ip address` is equivalent to `pointopoint` in `ifconfig`/`/etc/network/interfaces`, and I use `ip ... add` instead of `ip ... replace` above to remove any confusion about `replace`; `replace` basically just allows the command to not fail if these entries already exists; `replace` will also actually _add_ more entries instead of replacing e.g. based on the IP address unless the command lines are fully identical, e.g. running `ip addr replace 1.2.3.4/32 peer 1.2.3.1 dev eth0` after `ip addr replace 1.2.3.4/32 dev eth0` will result in 2 IP address entries, it cannot add/remove the `peer` flag for an existing entry.)

### Notes from our investigation

We first wanted to check what

```
ip route replace 1.2.3.1 dev eth0
ip route replace default via "1.2.3.1" dev eth0
```

does exactly.

For `ip route replace 1.2.3.1 dev eth0`:

* This entry generally tells Linux that anything in the specified network shoud be switched/ARPed directly, instead of being sent to a router (that would be specified using `via ROUTER_IP`).
  Without this (with no route set at all), we would get `no route to host` when trying to connect to ANY IP, and the later invocation further down of `ip route ... via SOME_IP` would also fail with `SOME_IP: no route to host`.
  This motivates why `ip route replace 1.2.3.1 dev eth0` is placed before `ip route replace default via "1.2.3.1" dev eth0` in NixOS.
* A thing that confused us is the syntax `ip route replace 1.2.3.1 dev eth0`.
  Usually `ip route` is used with a subnet, e.g. `1.2.3.1/24`.
  When just `1.2.3.1` is specified on the CLI, does that have the effect of `/0` or `/32`?
  * We checked in the `ip` code this causes `bitlen = 0` which suggests it behaves like `1.2.3.1/0`.
    But that is pretty weird, because it would mean "all IPs ever are in the local network", so the `default` route afterwards would be useless, as it would never be used (as the default route handles all cases that aren't in the local network).
    It would also mean that NixOS's chosen syntax
    `ip route replace 1.2.3.1 dev eth0` were unnecessary, because if it means `ip route replace 1.2.3.1/0 dev eth0`, we might as well write a hardcoded `ip route replace 0.0.0.0/0 dev eth0` since that, too would refer to `/0` (for `/0`, it should matter what address you write). So why would NixOS bother to splice in the actual IP in [the code](https://github.com/NixOS/nixpkgs/commit/d0435ba032473f89fad8b6425bb3feee3d6fd718#diff-344943ac406571508ecc8d0311e237519c34548ffaf4eba83a15758a1ced65a4R126)?
  * Investigating this more, from `man ip route`
    ```
       ip route { add | del | change | append | replace } ROUTE
       ROUTE := NODE_SPEC [ INFO_SPEC ]

       NODE_SPEC := [ TYPE ] PREFIX [ tos TOS ] [ table TABLE_ID ] [ proto RTPROTO ] [ scope
               SCOPE ] [ metric METRIC ] [ ttl-propagate { enabled | disabled } ]
    ```
    so `replace` needs a `PREFIX`, and for a prefix theroretically the IP shouldn't matter.
  * But contradicting this idea is:
    `iproute.c` seems to do something with the address:
     ```c
      if (dst.bytelen)
        addattr_l(&req.n, sizeof(req),
            RTA_DST, &dst.data, dst.bytelen);
     ```
     If the address doesn't matter and only the `PREFIX` (e.g. subnet `/..`) is important, why does it use it?
  * @aszlig found we should figure out what the kernel does with the address then, and if `/0` has special meaning there.
    He checked the kernel code:
    In the kernel, `SOMEIP/0` seems to be treated as "no prefix was set" -- the same handling as if `SOMEIP/32` was given.
    `/0` and `/32` create the same RT_NETLINK packet.
    https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/net/ipv4/ipmr.c?id=eff5f16bfd87ae48c56751741af41a825d5d4618#n2758
    * This heavily suggests that not specifying a `/..` (`ip route replace 1.2.3.1 dev eth0`) behaves conceptually like `/32` (`ip route replace 1.2.3.1/32 dev eth0`).
      This would make much more sense, because it makes the `default` route in NixOS useful, and makes the line mean "just the gateway is reachable directly".

### Testing

From our testing on 3 Hetzner servers (of which 2 are in the same Hetzner subnet), we found no visible difference between `/32`-pointopoint versus `/32` with route.

### Hetzner's statement

Hetzner support's opinion on whether the two are equivalent is:

> Setting point to point means that no matter what happens the traffic will always go to the peer first. By defining it as /32 it is very similar since the address will be the only entity in the subnet and will go to the defined gateway.

### Equivalence

From this we conclude that

    ip addr 1.2.3.4/32 peer 1.2.3.1 dev eth0
    
is likely equivalent to

    ip addr add 1.2.3.4/32 dev eth0
    ip route replace 1.2.3.1 dev eth0

but the former is slightly simpler.

Perhaps that could be determined with `bpftrace` on the trie that's involved in handling this in the kernel [TODO for us to write up this idea more concretely].


## What NixOps/NixOS should do

This means we have 2 options:

* Add to NixOS the ability to configure `peer` by adding a `networking.interfaces.<name>.ipv4.addresses.*.peer` option. (This should then ideally be implemented by both `network-interface-scripted.nix` and `network-interfaces-systemd.nix`.)
  * Then make NixOps generate code that sets `.peer`.
    * Drawback: NixOps would have to wait for the above to land in NixOS, and generating usage of that new option would make NixOps incompatible with older NixOS versions, unless we add some conditional config generation for this.
* Just change the `prefixLength` to `/32` in NixOps's generation, proposed in the autoclosed PR #1070

The latter seems like the simpler change to me.

But adding `networking.interfaces.<name>.ipv4.addresses.*.peer` to NixOS seems like a good idea independently.

PR #1039, opened 2018-11-02
Check types wit mypy

Mic92:
Motivation: Restore maintainability that was complained about in https://github.com/NixOS/nixpkgs/pull/48378#issuecomment-435213505

Mypy is optional typing system for python. It was designed by Guido van Rossum, the creator of python
and used on many big code bases such as Dropbox for python2 to python3 migrations.
Personally I was able to migrate a 6000 LOC non-trivial application with little test coverage in less then a day because it had type annotations. Please read the commit message descriptions for more information about the individual changes. Most types can be interfered automatically. To make a method type checked a type annotation has to be added. This also increases readability.
If this pull request is accepted I will make an effort to add semi-automatically types using the existing tests and https://github.com/dropbox/pyannotate.
I think mypy makes also sense beyond python3 migrations for a project like nixops where unittests are pretty hard due to dependencies on many external systems such as cloud apis.

### DIFF
--- a/nixops/azure_common.py
+++ b/nixops/azure_common.py
@@ -10,6 +10,7 @@ import requests
 from nixops.util import attr_property, check_wait
 import nixops.resources
 
+from typing import Dict
 from azure import *
 
 from azure.mgmt.common import SubscriptionCloudCredentials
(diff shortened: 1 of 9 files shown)


Mic92:
I wonder why there is no CI support in this repository. I could add travis ci, if you want to build at least the `release.nix`.

Mic92:
cc @domenkozar

Mic92:
Anything left to do here?

PR #1041, opened 2018-11-04
fix release build w.r.t. mypy

Mic92:
Mypy was added to the wrong build output.
I was testing the wrong `release.nix` attribute when testing this.
I also added travis support for the same reason to avoid this in future.


Mic92:
https://github.com/NixOS/nixops/pull/1042

PR #1042, opened 2018-11-04
Travis support for better code review

Mic92:
This is complementary for github review
as it provides instead feedback in the pr interface.
Since all it does is building release.nix the
maintenance effort should minimal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: nix
+script: nix-build --quiet release.nix -A build.x86_64-linux
--- a/nixops/diff.py
+++ b/nixops/diff.py
@@ -2,8 +2,11 @@ import os
 import json
 import itertools
 
-from typing import Any, Callable, Optional, List, Dict
+from typing import Any, Callable, Optional, List, Dict, Union, AnyStr
 import nixops.util
+from nixops.deployment import Deployment
+from nixops.logger import MachineLogger
+from nixops.state import StateDict
 
 class Diff(object):
     """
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,2 @@
+language: nix
+script: nix build -f release.nix build.x86_64-linux
```
Mic92: I had to use this instead of `nix-build` because it produced too much output.

Mic92:
There is currently I am investigating since it works on my nixpkgs revision: https://travis-ci.com/Mic92/nixops/jobs/156237797#L4428

```console
$ python2 -c 'import boto.ec2'
```

.travis.yml
Mic92: Found the `nix-build --quiet` option.

Mic92:
Builds now: https://travis-ci.com/Mic92/nixops/builds/90943057

Mic92:
Builds now https://travis-ci.com/Mic92/nixops/builds/90943180

nixops/diff.py
```diff
@@ -163,7 +183,9 @@ def _default_handle(self):
         raise NotImplementedError
 
     def get_deps(self):
+        # type: () -> List[Handler]
         return self._dependencies
 
     def get_keys(self,*keys):
+        # type: (*Any) -> List[str]
```
Mic92: Strings can be str and unicode here?

nixops/diff.py
```diff
@@ -143,7 +163,7 @@ def retrieve_def(d):
 
 class Handler(object):
     def __init__(self, keys, after=None, handle=None):
-        # type: (Dict[str, Dict[str,str]], Optional[List], Optional[Callable]) -> None
+        # type: (List[str], Optional[List], Optional[Callable]) -> None
```
Mic92: same question here

domenkozar:
lgtm

domenkozar:
I've enabled travis-ci for this repository.

domenkozar:
Mosts tests are functional, so you need configured backends.

tomberek:
Add `typing` to release.nix?

Mic92:
@tomberek see https://github.com/NixOS/nixops/blob/d5ad09cd3db5ed096d1bd17fbe5a0790b434f2c7/release.nix#L99

Issue #959, opened 2018-05-25
resources.sshKeyPairs empty in deployment


rehno-lindeque:
```nix
{
  # Create a public / private keypair
  resources.sshKeyPairs.build-master = {};

  machine = { resources, lib, ... }: {
    deployment = {
      targetEnv = "container";
      keys = {
        # Create a file at /run/keys/id_build_master that may be used as an identity by ssh (for example, sshKey in nix.buildMachines)
        "id_build_master" = {
          # This unexpectedly creates an empty file in /run/keys/id_build_master:
          text = lib.traceVal "${resources.sshKeyPairs."build-master".privateKey}";
        };
      };
    };
  };
}
```

```
$ nixops create -d experiment container.nix
created deployment ‘2f45bdf5-5fd3-11e8-b372-a45e60e8054f’
2f45bdf5-5fd3-11e8-b372-a45e60e8054f

$ nixops deploy -d experiment
trace: 
machine.....> building initial configuration...
```

Aside from my personal use-case (generating a file that can be used with ssh on the target machine), I find it surprising that I am able to reference `resources.sshKeyPairs."build-master".privateKey` and yet get back an empty string for my trouble.

ElvishJerricco:
I just got hit by this. It looks like NixOps evaluates the config in two passes in order to get enough info to properly pass to all the modules. The keys being deployed are being extracted from the first pass, which has less information e.g. no `privateKey` / `publicKey` in `resources.sshKeyPairs.<name>`. Is this a regression?

ElvishJerricco:
Should keys be resent in phase 2? I worry that there may be differences in the keys between phases that could lock things out of things.

exarkun:
The API documentation (https://nixos.org/nixops/manual/#idm140737316403040) specifically says you can read the value of publicKey and privateKey with these attributes: "You can access the generated keypair by using the following options."

Is this just a bug?  Is there some circumstance under which the documentation is actually correct?

ElvishJerricco:
Deployment info like machines' IPs and generated keys aren't available in phase 1. The nix expressions are re-evaluated in phase 2 with these values plugged in. The issue is that `deployment.keys` are deployed from the phase 1 evaluation, when the generated keys aren't present. Maybe they should be deployed from the phase 2 evaluation?

tomberek:
[...]

PR #1048, opened 2018-11-16
init: arbitrary output resources

tomberek:
Using the ssh_keypair as a model, this allows one to create whatever JSON via a shell script in order to allow customized resources.

Example use:
```
resources.output.myKeys.func = ''
    tinc --batch -c $NIXOPS_OUTPUT_DIR generate-ed25519-keys > /dev/null
    jq '{pub:$pub,priv:$priv}' --null-input --rawfile pub $NIXOPS_OUTPUT_DIR/ed25519_key.pub --rawfile priv $NIXOPS_OUTPUT_DIR/ed25519_key.priv
'';
```


nix/output.nix
```diff
@@ -0,0 +1,30 @@
+{ config, lib, name ? "default-output", ... }:
+with lib;
+{
+  options = {
+    name = mkOption {
+      default = "${name}";
+      type = types.str;
+      description = "Name of the output.";
+    };
+    func = mkOption {
```
Mic92: Instead of `func` I just would call this `script`.

nix/output.nix
```diff
@@ -0,0 +1,30 @@
...
+        Text of a shell script which will produce a JSON value.
+        Warning: This uses shell features and is potentially dangerous.
+        Environment variables: 
+            NIXOPS_OUTPUT_DIR is a temp directory available for use.
+        '';
+    };
+    value = mkOption {
+      default = null;
+      type = types.nullOr types.str;
+      description = "Result of running function";
```
Mic92:
```suggestion
      description = "Result of running script";
```

nix/output.nix
```diff
@@ -0,0 +1,30 @@
...
+      default = "${name}";
+      type = types.str;
+      description = "Name of the output.";
+    };
+    func = mkOption {
+      default = null;
+      type = types.nullOr (types.str);
+      #type = types.nullOr (types.either types.str types.path);
+      description = ''
+        Text of a shell script which will produce a JSON value.
```
Mic92: I wonder if we should limit this to shell. I could imagine with a high-level programming languages
it would be easier to generate json in a safe way. Also one could leverage libraries.

tomberek:
Thanks for the feedback, I’ll incorporate the thoughts over the weekend sometime. I do expect that having resources available in phase 1 is going to be more important going forward: #959

Any thoughts on the direction of the design? At the moment I normally accomplish a similar thing by just using the file system and putting deployments into their own folder and use relative paths. With key/value store we can continue to use the “resource” paradigm when we have more backends. Another consideration is that it is sometimes cumbersome to output proper json from the command line.

The net effect of this design is that you can include any supporting scripts directly into nix and they are run for you automatically.

tomberek:
Another thought, use nix-shell in some way to manage the local dependencies that may be in the script? Or just allow people to write whatever they want?

Mic92:
First of all I am not in the NixOps team, so I won't be able to merge this pull request in the end.
The design makes sense to me. I first thought those outputs could be build by nix derivations, but then we would end up with secrets in the nix store, which is not what we want. 

Rather then a shell script I would change it to accept the path to an executable program. This could be a script or whatever the user wants to use.
The question is if it possible to have something like this:

```nix
resources.output.myKeys.command = pkgs.writeScript "foo.py" ''
#!${python}/bin/python
# do whatever
'';

resources.output.myKeys2.command = pkgs.writeScript "foo.sh" ''
#!${bash}/bin/bash
# do whatever
'';
```

Then one could use whatever fits the best to generate the output.
We already have many helpers in nixpkgs to generate programs on the fly and we might
have more in the future: https://github.com/NixOS/nixpkgs/pull/49290

An alternative could be the following to allow additional dependencies 
on the fly by using a nix-shell in the shebang:

```nix
resources.output.myKeys2.command = pkgs.writeScript "foo.sh" ''
#!/usr/bin/env nix-shell -p
#!nix-shell -p python3.pkgs.foobar -p python3 nix -i python3
'';
```

nixops/resources/output.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def __init__(self, depl, name, id):
+        self.id = id
+        nixops.resources.ResourceState.__init__(self, depl, name, id)
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
+        if self.value is None:
+            self.name = defn.name
+            self.state = self.UP
+            output_dir = tempfile.mkdtemp(prefix="nixops-output-tmp")
```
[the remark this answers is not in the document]
tomberek: revised with try-finally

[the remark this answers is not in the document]
tomberek: also working type annotations ...

tomberek:
Wireguard example:
```
resources.output.wg_keys.script = ''
  wg genkey | tr -d '\n' | tee $out/privatekey | wg pubkey | tr -d '\n' > $out/publickey
  jq '{pub:$pub,priv:$priv}' --null-input --rawfile pub $out/publickey --rawfile priv $out/privatekey
'';
```

nixops/resources/output.py
```diff
@@ -0,0 +1,104 @@
...
+        # type: () -> Optional[str]
+        if self.value is not None:
+            # Avoid printing any potential secret information
+            return "{0}-{1}".format(self.name,hashlib.sha256(self.value).hexdigest()[:32])
+        else:
+            return None
+
+    def __init__(self, depl, name, id):
+        # type: (Any,str,int) -> None
+        # Any is: nixops.deployment.Deployment
```
Mic92: You can actually use import for `nixops.deployment.Deployment` instead of `Any`.

nixops/resources/output.py
```diff
@@ -0,0 +1,104 @@
...
+        self.id = id
+        nixops.resources.ResourceState.__init__(self, depl, name, id)
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
+        # type: (Any,bool,bool,bool) -> None
```
Mic92: Here too

nixops/resources/output.py
```diff
@@ -0,0 +1,104 @@
...
+        # type: () -> (str)
+        return "output"
+
+    @classmethod
+    def get_resource_type(cls):
+        # type: () -> (str)
+        return "output"
+
+    def __init__(self, xml):
+        # type: (Any) -> None
```
Mic92: This should be `xml.etree.elementtree.Element`.

nixops/resources/output.py
tomberek: fixed

nixops/resources/output.py
tomberek: fixed

nixops/resources/output.py
tomberek: fixed

tomberek:
Also have tested using this to serialize (via tar and base 64) an output to re-serialize after garbage collection:

```
resources_ssl= {
     output."ssl_test".script =''
       temp=$(nix-store -r $(nix-instantiate --quiet -E "(with import <nixpkgs>{}; custom-certs \"$RANDOM\")") | tr -d '\n')
       tar -C $temp -cf $out/temp.tar . >&2
       cat $out/temp.tar | base64
     '';
   };
```
This set's up a new set of SSL CA's and certificates. The result is serialized to a value stored in the statefile. The following in `deployment.keys` will extract it, re-create the original output from the statefile during deployment:
```
"test2" = let a = resources.output.ssl_test.value;
                     b = builtins.toFile "test2" a;
                     c = builtins.trace b (pkgs.runCommandCC "thing" {} ''
                       mkdir $out
                       pushd $out
                       cat ${b} | base64 -d | tar x
                       '');
         in
         if a == null then {} else
         {
         text = some-function c;
       };
```

nix/output.nix
```diff
@@ -0,0 +1,30 @@
...
+    script = mkOption {
+      default = null;
+      type = types.nullOr (types.str);
+      #type = types.nullOr (types.either types.str types.path);
+      description = ''
+        Text of a script which will produce a JSON value.
+        Warning: This uses shell features and is potentially dangerous.
+        Environment variables: 
+            $out is a temp directory available for use.
+        '';
```
aszlig: nitpick: Indentation, also please note that descriptions are written in Docbook, so maybe use [`<warning/>`](https://tdg.docbook.org/tdg/5.0/warning.html) and probably [`<envar/>`](https://tdg.docbook.org/tdg/5.0/envar.html) here.

nix/output.nix
```diff
@@ -0,0 +1,30 @@
...
+{
+  options = {
+    name = mkOption {
+      default = "${name}";
+      type = types.str;
+      description = "Name of the output.";
+    };
+    script = mkOption {
+      default = null;
+      type = types.nullOr (types.str);
```
aszlig: Parentheses not needed here.

nix/output.nix
```diff
@@ -0,0 +1,30 @@
+{ config, lib, name ? "default-output", ... }:
+with lib;
+{
+  options = {
+    name = mkOption {
+      default = "${name}";
```
aszlig: `default = name;` should be enough.

nixops/resources/output.py
```diff
@@ -0,0 +1,110 @@
...
+            try:
+                output_dir = tempfile.mkdtemp(prefix="nixops-output-tmp")
+                self.log("Running shell function for output ‘{0}’...".format(defn.name))
+                env = {} # type: Dict[str,str]
+                env.update(os.environ)
+                env.update({"out":output_dir})
+                res = subprocess.check_output(
+                        [defn.script],
+                        env=env,
+                        shell=True)
```
aszlig: Please note that this will make the script dependent on the system's shell. For example if you're using NixOps on a Debian machine, you'll end up getting `dash` instead of `bash`. So I'd probably wrap the script with `stdenv.shell` or `pkgs.bash` so that if the `script` is using bashisms it will still work.

One possibility would be to add another option to the resource, like eg. `executable`, which defaults to using something like eg. `pkgs.writeScript "somename" "#!${pkgs.stdenv.shell}\n${config.script}"`. That way you only need to run the `executable` here without the need to set `shell=True`.

tomberek: Incorporated all the other feedback. I'm not sure how to do this one. Whenever I try to do this, it seems the writeScript gets evaluated, but not instantiated.

tomberek:
This still functions on top of master. Also works well with #1054 

TODO: I'd like to create a similar thing, but where the resource is created and stored on the remote machines. Send-keys would then retrieve things like public keys  (pub/priv created on remote machine) and send public side to other machines that need it.

AmineChikhaoui:
@tomberek is there anything left to do in this PR from your side ?
@aszlig are you ok with the state of the PR at this point ?

release.nix
```diff
@@ -105,7 +105,7 @@ rec {
         export PATH=$(pwd)/scripts:${openssh}/bin:$PATH
       '';
 
-      doCheck = true;
+      doCheck = false;
```
aszlig: Wait, why!?

nixops/resources/output.py
aszlig: Can you post an output of that with `--debug`?

nixops/deployment.py
```diff
@@ -653,6 +653,11 @@ def build_configs(self, include, exclude, dry_run=False, repair=False):
             os.environ['NIX_CURRENT_LOAD'] = load_dir
 
         try:
+            # Re-evaluate info.machines in order to have phys_expr (IP's, resources, etc) available for keys
```
AmineChikhaoui: Shouldn't this be on a separate PR ? seems like a different feature

nixops/backends/ec2.py
```diff
@@ -261,7 +261,7 @@ def resource_id(self):
 
 
     def address_to(self, m):
-        if isinstance(m, EC2State): # FIXME: only if we're in the same region
+        if isinstance(m, EC2State) and m.region == self.region and m.zone == self.zone:
```
AmineChikhaoui: This is also unrelated to the PR afaik

nixops/resources/output.py
```diff
@@ -0,0 +1,112 @@
...
+        return "output"
+
+    @classmethod
+    def get_resource_type(cls):
+        # type: () -> (str)
+        return "output"
+
+    def __init__(self, xml):
+        # type: (Element) -> None
+        nixops.resources.ResourceDefinition.__init__(self, xml)
```
AmineChikhaoui: Can you get rid of the xml parsing ?
If you don't override __init__, the definition should be in a dictionary that is available through `defn.config` iirc

nixops/resources/output.py
```diff
@@ -0,0 +1,112 @@
...
+    @property
+    def resource_id(self):
+        # type: () -> Optional[str]
+        if self.value is not None:
+            # Avoid printing any potential secret information
+            return "{0}-{1}".format(self.name,hashlib.sha256(self.value).hexdigest()[:32])
+        else:
+            return None
+
+    def __init__(self, depl, name, id):
```
AmineChikhaoui: not sure why this override is needed ?

nixops/resources/output.py
```diff
@@ -0,0 +1,112 @@
...
+        # type: (Deployment,str,int) -> None
+        self.id = id
+        nixops.resources.ResourceState.__init__(self, depl, name, id)
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
+        # type: (OutputDefinition,bool,bool,bool) -> None
+        if (defn.script is not None ) and (self.script != defn.script) or self.value is None:
+            self.name = defn.name
+            try:
+                output_dir = tempfile.mkdtemp(prefix="nixops-output-tmp")
```
AmineChikhaoui:
```suggestion
                output_dir = nixops.util.SelfDeletingDir(tempfile.mkdtemp(prefix="nixops-output-tmp"))
```
I think this will automatically take care of deleting the directory on exit

nixops/deployment.py
tomberek: it's in another PR, rebasing....

nixops/backends/ec2.py
tomberek: it's in another PR, rebasing....

nixops/resources/output.py
tomberek: Later on i detect changes between self and defn. I'm not 100% sure it is needed. I'd welcome thoughts on this.

nixops/resources/output.py
tomberek: I'm not sure either, but removing it causes:
```
...
File "/home/dev/nixops/nixops/resources/__init__.py", line 76, in _set_attrs
    (self.id, n, v))
AttributeError: 'OutputState' object has no attribute 'id'
```

tomberek:
Added a test. Runs with `python tests.py tests.functional.test_output_creates -A vbox`

nixops/resources/output.py
tomberek: incorporated in latest commit.

nixops/resources/output.py
AmineChikhaoui: Yes you can still compare. The config will be available in `defn.config` inside the ResourceState.create method, so you could pull `script` from example using `defn.config['script']`
See and example of a resource implementation that does that https://github.com/NixOS/nixops/blob/master/nixops/resources/sns_topic.py

nixops/resources/output.py
AmineChikhaoui: Hm not sure what's causing that, but pretty sure the whole `__init__` override isn't needed :-)

AmineChikhaoui: Ok I looked at this and it's caused by the `name =  nixops.util.attr_property("name",None)` declaration as name is used by nixops internally for persisting the state so it got confused.
So just switch from e.g name to `outputName`

tests/functional/single_machine_outputs.nix
```diff
@@ -0,0 +1,16 @@
...
+      script = ''
+        #!/bin/sh
+        echo -n '"12345"'
+      '';
+    };
+  };
+  machine = {resources, ...} : {
+    imports = [ <nixpkgs/nixos/modules/profiles/minimal.nix> ];
+    deployment = {
+      keys."secret.key".text = resources.output.thing.value;
```
AmineChikhaoui: This requires the 3 evaluations you added, so I suggest another example/test ?

nixops/resources/output.py
```diff
@@ -0,0 +1,112 @@
...
+
+import os
+import nixops.util
+import nixops.resources
+
+import tempfile
+import shutil
+import subprocess
+import hashlib
+
```
AmineChikhaoui: nitpick: remove extra line

nixops/resources/output.py
```diff
@@ -0,0 +1,112 @@
...
+        # type: () -> Dict[str,str]
+        return {'value': self.value}
+
+    def destroy(self, wipe=False):
+        # type: (bool) -> bool
+        if self.depl.logger.confirm("are you sure you want to destroy {0}?".format(self.name)):
+            self.log("destroying...")
+        else:
+            raise Exception("can't proceed further")
+            return False
```
AmineChikhaoui: I guess `return False` isn't really needed as it would exit with raising an exception

nixops/resources/output.py
tomberek: @AmineChikhaoui That worked. Updated. Thanks!

nixops/resources/output.py
tomberek: updated.

nixops/resources/output.py
tomberek: remove

nixops/resources/output.py
tomberek: removed

tests/functional/single_machine_outputs.nix
tomberek: Added a new test.

AmineChikhaoui:
@tomberek Btw didn't you have a version where it used `pkgs.writeScriptBin` ? That idea sounds better, I'm not sure how easy would it be to define the script as a "package" basically with all needed dependencies without relying on the system.
Did you have problems while trying such approach ?

Also maybe an overkill, but I wish we could do such scripts in isolation (inside nix builds ? in chroots?), the idea of allowing random scripts is a bit scary to me :) But that's probably more complicated.

nixops/deployment.py
```diff
@@ -309,13 +309,13 @@ def evaluate_args(self):
         except subprocess.CalledProcessError:
             raise NixEvalError
 
-    def evaluate_config(self, attr):
+    def evaluate_config(self, attr,extra_exprs=[]):
```
AmineChikhaoui: Note that the 3 evals diff is still here even after the rebase

tomberek:
I never got the writeScriptBin approach to work. It had to do with the order of instantiation, so the drv got created, but never realized. I decided to leave that as a future improvement. 

Regarding the three-phase: yes, most use cases include using an output resource in deployment.keys such that it’s most useful with it. I can remove it if you think it’s cleaner that way. (I’d have to modify the tests, but that’s ok)

Another approach is to redesign the whole thing in a better way, I’d appreciate any ideas. Especially if we include a method to generate the resources remotely.

AmineChikhaoui:
> I never got the writeScriptBin approach to work. It had to do with the order of instantiation, so the drv got created, but never realized. I decided to leave that as a future improvement.

Can you share the nix/output.nix that had the issue ? I want to try it out as well.

> Regarding the three-phase: yes, most use cases include using an output resource in deployment.keys such that it’s most useful with it. I can remove it if you think it’s cleaner that way. (I’d have to modify the tests, but that’s ok)

I'm just not sure if adding yet another eval is a good idea, it will make things slower in my opinion. I'd like to get more opinions on that (perhaps in a separate PR): @edolstra @rbvermaa @grahamc @aszlig @domenkozar  (randomly pinging people :-) ), any opinions ?

tomberek:
@AmineChikhaoui : I've got some of the "executable" work here if you want to play around with it:
https://github.com/tomberek/nixops/blob/v2_outputs/nix/output.nix
https://github.com/tomberek/nixops/blob/v2_outputs/nixops/resources/output.py

Note: it's not in a working state.

AmineChikhaoui:
I gave it a try to use a generated executable (which uses `pkgs.writeScriptBin`) but I've hit some infinite recursions issues during eval. See gist: https://gist.github.com/AmineChikhaoui/65c93cb4c6934f186c0afc818b88b7c6.

Not sure yet why it would be very different from using a regular NixOS module. I know NixOps is a bit different in the evaluation but it still uses `evalModules` under the hood.
@aszlig Please let me know if you spot an obvious issue and see that we can still use a nix pkg for the script instead of just using the host environment which I would like to avoid if possible.

aszlig:
@AmineChikhaoui: This looks a lot like it's recursing into `<nixpkgs/pkgs/top-level/all-packages.nix>`. Can you post at which stage you get the infinite recursion?

AmineChikhaoui:
@aszlig Ok, I think this happens in `Deployment::evaluate_args()` which is in my network:
```
nix-instantiate --read-write-mode \
 --arg networkExprs '[ "/home/amine/src/nixops/output.nix" ]' \
 --arg args {} --argstr uuid 9ad8bea1-6215-11e9-a024-02420a47088d \
 --argstr deploymentName out nix/eval-machine-info.nix \
 --eval-only --xml --strict --arg checkConfigurationOptions false -A info
```
with @grahamc's nix patch https://github.com/NixOS/nix/pull/2761, I get the following output (not sure if relevant for our case though):
```
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/check-meta.nix:22:5 3730 130 28.6923
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/check-meta.nix:215:9 2145 65 33
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/systems/doubles.nix:24:50 4768 18 264.889
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:271:5 236 10 23.6
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/systems/inspect.nix:58:45 539 24 22.4583
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:98:6 23 1 23
/home/amine/src/nixops/nix/eval-machine-info.nix:350:21 158 5 31.6
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/types.nix:348:23 15 1 15
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:370:19 216 12 18
/home/amine/src/nixops/nix/eval-machine-info.nix:184:9 643 12 53.5833
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/top-level/stage.nix:206:11 2042 6 340.333
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/make-derivation.nix:108:95 2217 77 28.7922
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/make-derivation.nix:106:14 188 6 31.3333
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:399:7 94 5 18.8
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/linux/default.nix:350:9 49 1 49
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:455:21 299 15 19.9333
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:360:14 313 15 20.8667
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/attrsets.nix:356:46 1077 52 20.7115
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/strings.nix:59:16 50 1 50
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:458:16 300 15 20
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:192:8 19 1 19
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/strings.nix:123:60 99 2 49.5
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/check-meta.nix:214:9 118672 65 1825.72
/home/amine/src/nixops/nix/eval-machine-info.nix:192:9 18 1 18
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/make-derivation.nix:108:33 3653 77 47.4416
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/pkgs/stdenv/generic/make-derivation.nix:102:36 158 6 26.3333
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/systems/parse.nix:174:12 8819 19 464.158
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:96:36 43 1 43
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:243:23 299 12 24.9167
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:218:9 184 7 26.2857
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/attrsets.nix:125:18 107 3 35.6667
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/options.nix:110:10 15 1 15
/nix/store/5kbkahsc6kz480syx8i063n5ys2k0pd0-nixexprs.tar.xz/lib/modules.nix:350:17 282 15 18.8
```

AmineChikhaoui:
Btw the first eval `Deployment::evaluate_config()` seems to be wrong as well as it doesn't produce anything basically:
```
[nix-shell:~/src/nixops]$ nixops deploy -d out --debug
nix-instantiate --read-write-mode -I nixops=/home/amine/src/nixops/nixops/../nix --arg networkExprs [ "/home/amine/src/nixops/output.nix" ] --arg args {} --argstr uuid 9ad8bea1-6215-11e9-a024-02420a47088d --argstr deploymentName out <nixops/eval-machine-info.nix> --eval-only --xml --strict --arg checkConfigurationOptions false -A info.network
XML output of nix-instantiate:
<?xml version='1.0' encoding='utf-8'?>
<expr>
  <attrs>
  </attrs>
</expr>
```

edolstra:
I agree with @aszlig, `output` is a pretty vague name. Seeing `resources.output` doesn't really give an intuitive idea of what kind of resource it is.

tomberek:
@edolstra I'm not committed to any particular name. Any suggestions?

- resources.local (in anticipation of resources.remote)
- resources.static

@aszlig recommendations:
- resources.commandOutput
- resources.customOutput
- resources.custom

I'm thinking resources.custom would be good. It could later come in multiple types "local" and "remote", "vault", "s3", "url", etc. specifying where the resource is created and/or stored.

I don't want to bikeshed on this, so if there's a consensus, strong feeling, no response for a day, or a first response for one of these I will refactor/test/push a new version.

AmineChikhaoui:
Soo @edolstra mumbled something about --xml vs --json when I asked about the issue but I thought it's not related then I tried that and it's actually the root cause of the infinite recursion issue I was seeing.
https://gist.github.com/AmineChikhaoui/a028d72187ff07370d41421281b8df20 :man_shrugging: 

I wish we can nuke the xml parsing we have there and only use json but that will require an update to all the resources that do that in the `ResourceDefinition` classes.
On the other hand, I'm not sure why the behavior is different using --xml vs --json, that should probably be fixed in Nix.

AmineChikhaoui:
> @aszlig recommendations:
> 
>     * resources.commandOutput
> 
>     * resources.customOutput
> 
>     * resources.custom
Using `resources.commandsOutput` seems reasonable to me.

AmineChikhaoui:
Forgot to post an update here:
I gave up on the packaged script approach: even if the xml was generated properly nixops currently only builds the machines part, we can change that maybe but we're also blocked by the xml usage, in order to get rid of it there should be a major refactoring of the resources/backends using it.

So I guess we will go with the current approach.
@tomberek can you take care of the resource name change and also remove the 3rd evaluation, we can discuss adding it in the other PR that you have open.

tomberek:
@AmineChikhaoui : updated as requested. Thanks for working through this!

I think you ran into the same issue I did on the script.

nix/command-output.nix
```diff
@@ -0,0 +1,32 @@
...
+    name = mkOption {
+      default = name;
+      type = types.str;
+      description = "Name of the Command Output.";
+    };
+
+    script = mkOption {
+      default = null;
+      type = types.nullOr types.str;
+      #type = types.nullOr (types.either types.str types.path);
```
AmineChikhaoui: can you please remove the extra comment here

nix/command-output.nix
```diff
@@ -0,0 +1,32 @@
...
+      description = ''
+        Text of a script which will produce a JSON value.
+        <warning>Warning: This uses shell features and is potentially dangerous.</warning>
+        Environment variables: 
+        <envar>$out</envar> is a temp directory available for use.
```
AmineChikhaoui: what's $out ?

tomberek: `out` is available as an environment variable in the script as temp directory.
https://github.com/NixOS/nixops/pull/1048/files#diff-6d7517db2e330d537da7846c6ab7f4c5R69

PR #1047, opened 2018-11-16
nix/eval-machine-info.nix: fix evaluation with nixpkgs master

Mic92:
The evaluation broken with https://github.com/NixOS/nixpkgs/commit/6c68fbd4e1f8beac39cb1f499ff90c78256262d6

I have no idea why this file was imported in the first place:

https://github.com/NixOS/nixops/commit/bac59ad1b04516510e9c67bbd82b3d310084b2ac#diff-6f15e27c3e048b98302b636c2b8f983bR7

I just removed it and it is still working for my `none` backend.
However since this is file is sufficiently complex I cannot
tell what the general consequence of removing is.

PR #1050, opened 2018-11-17
Update release notes.

Mic92:
To fix evaluation with nixpkgs unstable we need a new release of nixops.
Also see https://github.com/NixOS/nixops/pull/1047

cc @rbvermaa @domenkozar

### DIFF
--- a/doc/manual/release-notes.xml
+++ b/doc/manual/release-notes.xml
@@ -5,6 +5,38 @@
 
 <!--==================================================================-->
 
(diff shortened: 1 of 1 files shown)


Mic92:
This should also help: https://github.com/NixOS/nixpkgs/pull/50525

AmineChikhaoui:
Thanks @Mic92 , I have added on top of these with the latest changes. Now applied in https://github.com/NixOS/nixops/commit/4fba1797311d881a7170943b231b922f72b80f67.

PR #1058, opened 2018-11-30
adding check to the key-pair

dorrahadrich:


nixops/resources/ec2_keypair.py
```diff
@@ -127,3 +127,12 @@ def keypair_used():
             self._conn.delete_key_pair(self.keypair_name)
 
         return True
+
+    def check(self):
+        self.connect()
+        try:
+            kp = self._conn.get_key_pair(self.keypair_name)
+        except IndexError as e:
+            kp = None
```
AmineChikhaoui: You could set the state to missing directly right ?
```suggestion
           self.state = self.MISSING
```

dorrahadrich: In fact, we need to check if the kp = None even when the IndexError exception wasn't caught. We can take as an example the use case where the get_key_pair returns None without an exception occurring, then we need to set the state to MISSING.

PR #1060, opened 2018-12-09
EC2: add a physical property for 5th gen instances

amemni:
It seems nice to have a new option under deployment.ec2.physicalProperties to flag 5th generation EC2 instances with Nitro cards, same a it's done for the ebsOptimized option, which instances require specific processing in certain areas. Example: EBS volumes are broadcasted as NVMe block devices named /dev/nvme[0..26]n1 thus cannot be mounted to /dev/xvd[f..z]. Feel free to change the naming, thanks !!

See: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances

I also added a check here for 'tenancy' in product, otherwise the generate-ec2-properties.py would fail in entry HMJJY7J5YC97S5A2 which is a "CPU Credits" not a "Compute Instance" (line 6482 of the [index.json](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json)): https://github.com/amemni/nixops/blob/f7eb86cbe4b9dcc531f35c7d27817965bcf96dd9/nix/generate-ec2-properties.py#L13


yorickvP:
This has the same problem as #971, `ebsOptimized` is gone for some instances.

PR #1067, opened 2018-12-27
Do not suppress rollback exceptions

olorin:
I'm not sure what the original motivation for this try/except block
was, though it seems to have been added several years ago. I think if
an exception occurs during rollback, it needs to be either handled
(though I'm not sure there's anything we could do with it) or surfaced
to the user; silent failures can only cause problems.

(I don't mind how the failure is presented; if you'd prefer to avoid a
top-level exception here I can re-add the try/except block with an
error message.)

### DIFF
--- a/nixops/statefile.py
+++ b/nixops/statefile.py
@@ -37,10 +37,9 @@ class Connection(sqlite3.Connection):
         assert self.nesting >= 0
         if self.nesting == 0:
             if self.must_rollback:
-                try:
-                    self.rollback()
-                except sqlite3.ProgrammingError:
-                    pass
+                # Any exceptions during rollback need to be handled by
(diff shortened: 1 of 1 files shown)


yorickvP:
The relevant commit is https://github.com/NixOS/nixops/commit/60c1be2b2efc3340f11129724feb4547036141b2 , which says " If an error caused the sqlite database to close, don't mask that error with 'Cannot operate on a closed database' from rolling back the db"

This should either be added as a comment or fixed some other way.

olorin:
Yep that's fair - I thought I checked git-blame but must have missed that message. Will update to not suppress the original error.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #1078, opened 2019-01-11
Start multi-user.target on deploy

### LINKED ISSUE #1063: nixops deploy doesn't bring server in desired state because it doesn't start stopped systemd units
If my `nginx` shuts down because of some failure in another systemd unit that it depends on, and I fix the issue and deploy with nixops, my nginx isn't actually started.

In general, when I run `nixops deploy`, the desired state isn't reached if a unit stopped for some reason.

It's only reached with `--force-reboot` or `nixops reboot`.

I think this is because `nixops deploy` doesn't actually `systemctl isolate` any target like `multi-user.target` (which is the default target that we put `requiredBy` on in NixOS service modules if we want it to be started).

I propose that we should probably `systemctl isolate multi-user.target` during `nixops deploy`.

Any opposing views?

nh2:
Fixes #1063.

This makes `nixops deploy` congruent (the declared services
will always be running after deploy, no matter what the state
of the server was before the deploy).

### DIFF
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -747,6 +747,11 @@ class Deployment(object):
                     # failed to start after the reboot.
 
                 if res == 0:
+                    res = m.run_command("systemctl isolate multi-user.target")
+                    if res != 0:
+                        raise Exception("unable to start multi-user.target")
+
+                if res == 0:
                     m.success("activation finished successfully")
 
                 # Record that we switched this machine to the new


grahamc:
Should the NixOS activation script do it, instead?

nh2:
@grahamc I asked that too on https://github.com/NixOS/nixops/issues/1063#issuecomment-453253939, but I can't easily tell if this would be too much of a change for non-NixOps NixOS (desktop) users.

deepfire:
@grahamc, anything can be done to move this forward?

AmineChikhaoui:
@deepfire I'm not sure if it's the job of Nixops to handle this. Nixops calls `switch-to-configuration` and that's where this should be handled imo.

deepfire:
@AmineChikhaoui, @grahamc, do we have a Nixpkgs issue to discuss this change of `switch-to-configuration`?

AmineChikhaoui:
@deepfire Not that I'm aware of.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #1080, opened 2019-01-16
show-physical backup (aws): extract devices encryption keys

PsyanticY:
In general we use `nixops show-physical -d test backup xxxxxxx ` to get the mapping between the devices and the snapshots then we use the output to deploy a new env using these snapshot, but when using encrypted volumes, another manually (and tedious) steps to extract and put the keys in that file is needed.
This pr Aims to include that automatically. 
Not sure if this can be considered bad practice from security perspective but since the keys are already in the state file and we are just extracting and formatting them, i don't think thats an issue.
Tested this with non encrypted and encrypted volumes created as separate/non separate resources.

### DIFF
--- a/nixops/backends/ec2.py
+++ b/nixops/backends/ec2.py
@@ -228,6 +228,11 @@ class EC2State(MachineState, nixops.resources.ec2_common.EC2CommonState):
 
                 if not is_root_device:
                     val[device_real] = { 'disk': Call(RawValue("pkgs.lib.mkOverride 10"), snap)}
+
+                    if self.block_device_mapping[device_stored]['encrypt']:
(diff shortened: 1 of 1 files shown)


PsyanticY:
@AmineChikhaoui  Addressed the concerns you raised. The key is now stored with the snapshot each time a backup operations run. Also added backward compatibility that i think should be removed after some  a release or so.

tewfik-ghariani:
Any updates regarding this?

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

PR #1085, opened 2019-01-28
aws[s3]: Add versioning support

PsyanticY:


nix/s3-bucket.nix
```diff
@@ -40,6 +40,12 @@ with lib;
       description = "The JSON lifecycle management string to apply to the bucket.";
     };
 
+    versioning = mkOption {
+      default = "Suspended";
+      type = types.str;
```
grahamc: Instead of a string, should use an enum: https://github.com/NixOS/nixops/blob/28231a177d751e800af3223a8763ea75b0ef9dd9/nix/cloudwatch-metric-alarm.nix#L86

grahamc:
```suggestion
      type = types.enum [ "Suspended" "Enabled" ];
```

should do it, I think (please test it).

PsyanticY: Thanks, Updated ( and tested)

nixops/resources/s3_bucket.py
```diff
@@ -105,10 +107,21 @@ def create(self, defn, check, allow_reboot, allow_recreate):
             except botocore.exceptions.ClientError as e:
                 if e.response['Error']['Code'] != "BucketAlreadyOwnedByYou": raise
 
+            try:
```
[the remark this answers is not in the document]
PsyanticY: Done

PR #1087, opened 2019-02-01
HashiCorp Vault AppRole support.

PsyanticY:
Hashicorp vault provide a highly available/secure  why to store secrest in the cloud using various back-ends (consul for instance). Authentication to the vault can occur using multiple methods such as LDAP, GitHub and Approle which is the best choice for programmatic users like ec2/gcp (...) deployments.
The idea is to create an AppRole for each deployment which will allow it to have access to secrets based on the policies (an argument). Once created the approle will then be used by the instance to request the keys/secrets from the vault server. ( I can explain the use case more if it is not that clear ). 

What left to to in This PR is:
- Take advantage of the diff engine and create functions that will handle updating the role once some parameters changes.
- Clean up some redundant code.


nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+    def show_type(self):
+        return "{0}".format(self.get_type())
+
+
+class VaultApproleState(nixops.resources.DiffEngineResourceState):
+    """State of a Vault Approle"""
+
+    state = nixops.util.attr_property("state", nixops.resources.ResourceState.MISSING, int)
+    vault_token = nixops.util.attr_property("vaultToken", None)
+    _reserved_keys = [ "vaultAddress", "roleName", "roleId", "bindSecretId"
```
AmineChikhaoui: are these really `_reserved_keys` ? it seems like these are actually the options from nix expression that defines the vault app role so you actually need them to fire up the diff engine.

`_reserved_keys` is meant to be used for ignoring attributes that are stored or internal which we don't want to handle e.g in EC2 the instance id would be reserved as otherwise the diff engine will always notice that there is a diff between the config and the state.

nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+                        "tokenMaxTtl", "tokenTtl", "period", "enableLocalSecretIds"
+                        "tokenType", "secretId", "cidrList"
+                      ]
+
+    @classmethod
+    def get_type(cls):
+        return "vault-approle"
+
+    def __init__(self, depl, name, id):
+        nixops.resources.DiffEngineResourceState.__init__(self, depl, name, id)
```
AmineChikhaoui: I would expect some handlers to be defined here since you're using `DiffEngineResourceState`.
The handler should handle a set of keys e.g `["vaultAddress" "roleName"]` and whenever there is a diff between the config and stored state it will call a function that will handle the realization of the diff. (see https://github.com/NixOS/nixops/blob/master/nixops/resources/vpc_internet_gateway.py#L41 for a simple example)

nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+        self.vault_address = self._state.get('vaultAddress', None)
+        self.secret_id = self._state.get('secretId', None)
+
+    def show_type(self):
+        s = super(VaultApproleState, self).show_type()
+        return s
+
+    @property
+    def resource_id(self):
+        path = '/v1/auth/approle/role/'
```
AmineChikhaoui: this path is being in many places, please defined once in a static variable, maybe in vault_common ?

nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+        return "resources.vaultApprole."
+
+    def get_physical_spec(self):
+        path = '/v1/auth/approle/role/'
+        return {'url': self.vault_address + path + self.role_name } if self.role_name else {}
+
+    def prefix_definition(self, attr):
+        return {('resources', 'vaultApprole'): attr}
+
+    def _connect(self, vault_token):
```
AmineChikhaoui: It doesn't seem that _connect does any actual connection, we typically use a connect method to create an http client and cache it for later uses. And btw any reason for not using an actual Vault python library such as https://github.com/hvac/hvac

nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+            r = requests.get(remote_endpoint, headers=header, timeout=5)
+        except requests.exceptions.RequestException as e:
+            raise e.args[0]
+        if r.status_code != 200:
+            self.log("Vault token is invalid. Request returned {} status".format(status))
+            raise Exception(str(r.status_code) + ' ' + r.reason, r.json())
+        # log warning if token lifetime is less then 30 min ?!s
+        # check for permissions and take sure it can create role_ids
+
+    def create(self, defn, check, allow_reboot, allow_recreate):
```
AmineChikhaoui: If you use handlers you should be able to get rid of the create method and only keep methods for realizing specific attributes changes

nixops/resources/vault_approle.py
```diff
@@ -0,0 +1,218 @@
...
+        self.vault_token = config['vaultToken']
+        self.role_name = config['roleName']
+        self._connect(self.vault_token)
+
+        self.log("Creating Approle {}' at {}".format(self.role_name, self.vault_address))
+        try:
+            header = {"X-Vault-Token": self.vault_token}
+            path = '/v1/auth/approle/role/'
+            remote_endpoint = self.vault_address + path + self.role_name
+            data = {
```
AmineChikhaoui: I would expect the API to support modifying most of the configs which I don't see being handled.
Which brings us back to the diff engine handers :) It should be much cleaner to define a handler for each modifiable attribute so that any change from the nix expression is handled correctly

nixops/resources/vault_approle.py
PsyanticY: @AmineChikhaoui  ok so `_reserved` is used for attributes set by nixops while the options are set using `nixops.util.attr_property` ?

nixops/resources/vault_approle.py
PsyanticY: Connect is basically to test if the token is actually valid and can create role it is not lile other _conect methods in ec2 pr datadog.
we can use hvoc but it is an overkill. Also, I was not able to package it in nixpkgs since it relies on another package where the maintainer is freezing a given other package (lot of packages :p) to a version and fail for to build with the version in nixpkgs.

nixops/resources/vault_approle.py
PsyanticY: As i said in the description i will be taking care of that next.

nixops/resources/vault_approle.py
PsyanticY: Yup.

PR #1102, opened 2019-02-20
Support persistent spot

PsyanticY:
This a followup and some fixes to #1011 
cc @Chakerbh

### DIFF
--- a/nixops/backends/ec2.py
+++ b/nixops/backends/ec2.py
@@ -681,68 +681,6 @@ class EC2State(MachineState, nixops.resources.ec2_common.EC2CommonState):
 
         return groups
 
-    def _get_network_interfaces(self, defn):
-        groups = self.security_groups_to_ids(defn.subnet_id, defn.security_group_ids)
-
-        return boto.ec2.networkinterface.NetworkInterfaceCollection(
-            boto.ec2.networkinterface.NetworkInterfaceSpecification(
-                subnet_id=defn.subnet_id,
-                associate_public_ip_address=defn.associate_public_ip_address,
-                groups=groups
-            )
-        )
-    def _build_run_instance_request(self, defn, zone, devmap, user_data, ebs_optimized):
-        """ Build the instance request definition in Boto3 format """
-
-        # FIXME: Include device mapping.
-
-        placement = dict(
-            AvailabilityZone=zone or ""
-            )
-
-        IamInstanceProfile = {}
-        if defn.instance_profile.startswith("arn:") :
-            IamInstanceProfile["Arn"] = defn.instance_profile
-        else:
-            IamInstanceProfile["Name"] = defn.instance_profile
-
-        args = dict(
-            InstanceType=defn.instance_type,
-            Placement=placement,
-            ImageId=defn.ami,
-            EbsOptimized=ebs_optimized,
-            IamInstanceProfile=IamInstanceProfile,
-            KeyName=defn.key_pair,
-            MaxCount=1, # We always want to deploy one instance.
-            MinCount=1
-        )
-
-        if defn.subnet_id != "":
-            if defn.security_groups != [] and defn.security_groups != ["default"]:
-                raise Exception("‘deployment.ec2.securityGroups’ is incompatible with ‘deployment.ec2.subnetId’")
-
-            args['NetworkInterface'] = dict(
-                AssociatePublicIpAddress=defn.associate_public_ip_address,
-                SubnetId=defn.subnet_id,
-                Groups=self.security_groups_to_ids(defn.subnet_id, defn.security_group_ids)
-            )
-        else:
-            args['SecurityGroups'] = defn.security_groups
-
-        if defn.spot_instance_price:
-            args["InstanceMarketOptions"] = dict(
-                MarketType="spot",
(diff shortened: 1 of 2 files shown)


nh2:
I’ve just tested to deploy my staging environment with this branch and it worked fine for me.

Note I haven’t tried out any of the new features added, only whether what I’m using right now still works (I also use a spot instance, but not yet the new flags you’ve added).

PsyanticY:
@nh2  Thanks for the review. i did get @Chakerbh  branch and added o ntop of it so i may did something wrong there to cause the commit message to appear twice Ill check and fix that.

PsyanticY:
@nh2  So i fixed the latter commits but i was not able to do for the duplicates ones i got a lot of conflicts when rebasing :( .

AmineChikhaoui:
@PsyanticY maybe try rebasing and squashing any extra commit + the 2 first commits

PsyanticY:
@AmineChikhaoui  Done squashing and the commits, also did re-base the pr on current master.  (that was a pain)

AmineChikhaoui:
@PsyanticY thanks !
Looking at the [spot instance requests docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#creating-spot-request-status), looks like we can actually test this manually. As the request will remain active in case of `persistent` unless we cancel it, an `aws ec2 terminates-instances` call should fake a spot instance termination and the instance should restart at some point (at least this is my understanding).
I'm trying with that, would like to double check the behavior before merging.

AmineChikhaoui:
Ok interesting, so here is what happens:

I create a persistent spot instance request `sir-foo`, that request gets fullfilled with instance id `i-first`, then I terminate the instance. When following the output of `describe-spot-instance-requests`, the sir creates another instance `i-second` after a couple of minutes, now nixops doesn't have any idea about that so the deploy would fail as it doesn't check the output of the sir again to pull out the new vm id.
Even a destroy would fail and barf about the sir being fulfilled unexpectedly with the new instance id, the only way to destroy is to run `nixops check` followed with `nixops destroy` as the first will update the state with `vmId = None` which will allow the cancellation of the sir as we don't have a way to compare vm ids anymore.

So, I believe that this PR needs more work to teach nixops to recover cleanly after a termination behavior of stop. The good part is at least it's possible to fake a spot instance termination :)

AmineChikhaoui:
(experimenting with some labels to make filtering PRs easy)

PsyanticY:
I think the way you did test the persistent spot is wrong since persistent spot relies on stopping the instance when we are out of capacity and restart it when we have. since the instance is stopped the instance id is not lost and we can recover cleanly given that. Terminating a persistent spot is not supported since if you even try to deploy a a persistent spot with `spotInstanceInterruptionBehavior` set to `terminate` ec2 will complain about `InvalidParameterCombination`: 
```creating EC2 instance (AMI ‘ami-8b3538f4’, type ‘r4.2xlarge’, region ‘us-east-1’)...
master..> got (possibly transient) EC2 error '{'Message': 'The terminate InstanceInterruptionBehavior is not supported when requestType is set to persistent.', 'Code': 'InvalidParameterCombination'}', retrying...
master..> got (possibly transient) EC2 error '{'Message': 'The terminate InstanceInterruptionBehavior is not supported when requestType is set to persistent.', 'Code': 'InvalidParameterCombination'}', retrying...
master..> got (possibly transient) EC2 error '{'Message': 'The terminate InstanceInterruptionBehavior is not supported when requestType is set to persistent.', 'Code': 'InvalidParameterCombination'}', retrying...
^Cerror: interrupted
```

AmineChikhaoui:
Ok I think you're right [the annoucement](https://aws.amazon.com/about-aws/whats-new/2017/09/amazon-ec2-spot-can-now-stop-and-start-your-spot-instances/) mentioned:
>Upon restart, the EBS root device is restored from its prior state, previously attached data volumes are reattached, and the instance retains its instance ID.

nixops/backends/ec2.py
```diff
@@ -679,97 +681,103 @@ def security_groups_to_ids(self, subnetId, groups):
...
-                         datetime.timedelta(0, defn.spot_instance_timeout)).isoformat()
+            args["InstanceMarketOptions"] = dict(
+                MarketType="spot",
+                SpotOptions=dict(
+                    MaxPrice=str(defn.spot_instance_price/100.0),
+                    SpotInstanceType=defn.spot_instance_request_type,
+                    InstanceInterruptionBehavior=defn.spot_instance_interruption_behavior
+                )
+            )
+            if defn.spot_instance_timeout and defn.spot_instance_request_type != "one-time":
```
AmineChikhaoui: Why do we need `defn.spot_instance_request_type != "one-time"` here?

PsyanticY: The argument `ValidUntil` is not relevant in a one time spot request since we cancel the request once it i fulfilled

AmineChikhaoui: Yeah but I think the whole point was to cancel requests if they exceed X time without getting fulfilled.

PsyanticY: ok then that make sense.

PR #1121, opened 2019-04-02
ec2: Usinng boto3 in operations: stop/start/reboot/console-output

PsyanticY:


PsyanticY:
Closing since it will be taken care of in here NixOS/nixops-aws#1

PR #1127, opened 2019-04-10
EC2: Add physical specs for elastic IP resources

PsyanticY:


nixops/resources/elastic_ip.py
```diff
@@ -51,6 +51,15 @@ def show_type(self):
     def resource_id(self):
         return self.public_ipv4
 
+   def get_physical_spec(self):
```
AmineChikhaoui: indentation off here ^

PR #1128, opened 2019-04-12
EC2[EBS] Allow assigning existing volume to the resource

PsyanticY:


nixops/resources/ebs_volume.py
```diff
@@ -74,6 +75,14 @@ def show_type(self):
     def resource_id(self):
         return self.volume_id
 
+    def get_physical_spec(self):
+        physical = {}
+        if self.volume_id:
+            physical['volumeId'] = self.volume_id
```
AmineChikhaoui: Any idea where this can be used ?

nixops/resources/ebs_volume.py
```diff
@@ -104,20 +113,24 @@ def create(self, defn, check, allow_reboot, allow_recreate):
...
 
-            if defn.size == 0 and defn.snapshot != "":
-                snapshots = self._conn.get_all_snapshots(snapshot_ids=[defn.snapshot])
-                assert len(snapshots) == 1
-                defn.size = snapshots[0].volume_size
-
-            if defn.snapshot:
-                self.log("creating EBS volume of {0} GiB from snapshot ‘{1}’...".format(defn.size, defn.snapshot))
+            if defn.volume_id != "":
+                vol_id = defn.volume_id
```
AmineChikhaoui: what about the other volume attributes ? I think it'd be better to implement a _check method that pulls the volume attribute and persist to the state so in case it's an existent volume we run self._check() first

AmineChikhaoui:
Btw I like the idea in general but sometimes I think it should be implemented at the resource level where we'll have something like `terraform import`

nixops/resources/ebs_volume.py
PsyanticY: I plan to use it to to tweak the show-physical-backup function to get  it to generate a file to override volumes with snapshots/other volumes

nixops/resources/ebs_volume.py
```diff
@@ -74,12 +76,48 @@ def show_type(self):
...
         if self._conn: return self._conn
         self._conn = nixops.ec2_utils.connect(region, self.access_key_id)
         return self._conn
 
+    def connect_boto3(self, region):
+        if self._conn_boto3: return self._conn_boto3
+        self._conn_boto3 = nixops.ec2_utils.connect_ec2_boto3(region, self.access_key_id)
+        return self._conn_boto3
+
+    def _check(self, defn):
```
AmineChikhaoui: So this won't run during `nixops check` as that expects a method _check() that doesn't get any arguments

nixops/resources/ebs_volume.py
```diff
@@ -74,12 +76,48 @@ def show_type(self):
...
+        self.connect_boto3(defn.region)
+        try:
+            _vol = self._conn_boto3.describe_volumes(VolumeIds=[defn.volume_id])['Volumes'][0]
```
AmineChikhaoui: also what if there is no volume_id in the definition like the old behavior ?

PsyanticY: the function should not run when there is no defined volume_id. the function should not run during a `nixops check` so i assume i should change the name if `_check` if  reserved ?

nix/ebs-volume.nix
```diff
@@ -29,6 +29,16 @@ with lib;
       description = "The AWS Access Key ID.";
     };
 
+    volumeId = mkOption {
+      default = "test123";
```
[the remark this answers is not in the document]
PsyanticY: Yup and i missed to remove that since i used it for testing. ty

Issue #1065, opened 2018-12-19
Update for recent azure-mgmt-* libraries


bendlas:
I've recently reverted an update to the `azure-mgmt-*` libraries in nixpkgs, because it broke `nixops`. I've submitted a PR with the updates, but `nixops` still doesn't play with the newer libraries

from https://github.com/NixOS/nixpkgs/pull/52550

```
Traceback (most recent call last):
  File "/nix/store/jsnrx8ggppa9q7iiry88s763y0p2qhb5-nixops-1.6.1pre2706_d5ad09c/bin/..nixops-wrapped-wrapped", line 5, in <module>
    from nixops import deployment
  File "/nix/store/jsnrx8ggppa9q7iiry88s763y0p2qhb5-nixops-1.6.1pre2706_d5ad09c/lib/python2.7/site-packages/nixops/deployment.py", line 1275, in <module>
    _load_modules_from("backends")
  File "/nix/store/jsnrx8ggppa9q7iiry88s763y0p2qhb5-nixops-1.6.1pre2706_d5ad09c/lib/python2.7/site-packages/nixops/deployment.py", line 1273, in _load_modules_from
    importlib.import_module("nixops." + dir + "." + module[:-3])
  File "/nix/store/w30zkhdjf6qp4ppx9xzffsh9dic31aq8-python-2.7.15/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/nix/store/jsnrx8ggppa9q7iiry88s763y0p2qhb5-nixops-1.6.1pre2706_d5ad09c/lib/python2.7/site-packages/nixops/backends/azure_vm.py", line 23, in <module>
    from nixops.azure_common import ResourceDefinition, ResourceState, ResId, normalize_location
  File "/nix/store/jsnrx8ggppa9q7iiry88s763y0p2qhb5-nixops-1.6.1pre2706_d5ad09c/lib/python2.7/site-packages/nixops/azure_common.py", line 19, in <module>
    from azure.mgmt.network import NetworkResourceProviderClient
ImportError: cannot import name NetworkResourceProviderClient
```

teto:
I don't think the pinning done in nixpkgs works when using `./dev-shell` that uses the local release.nix. 
Can't we finally use the same derivation and just run `nix-shell -A nixops` instead ?  
I just tried only to end up with
```
[...]

PR #944, opened 2018-05-05
Add initial support for hetzner cloud

goodraven:
This is based on the digital ocean backend. It also uses nixos-infect. I extended nixos-infect to be generic
for both backends.

Fixes #855

--

I see @nh2 has some work at https://github.com/nh2/nixops/tree/hetznercloud but I'm not sure what the status of that is.

PR #1130, opened 2019-04-16
nixos-infect: update channel to 18.09

asymmetric:
Not sure if 19.03 is a good idea given [the issues with azure-mgmt-*](https://github.com/NixOS/nixops/issues/1065), so sticking to 18.09.

### DIFF
--- a/nixops/data/nixos-infect
+++ b/nixops/data/nixos-infect
@@ -158,7 +158,7 @@ curl https://nixos.org/nix/install | sh
 
 source ~/.nix-profile/etc/profile.d/nix.sh
 
-[ -z "$NIX_CHANNEL"] && NIX_CHANNEL="nixos-17.09"
+[ -z "$NIX_CHANNEL" ] && NIX_CHANNEL="nixos-18.09"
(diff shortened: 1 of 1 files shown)


AmineChikhaoui:
@asymmetric btw I see nixos-infect is only used with Digital Ocean, I don't have an account there. Does it still work and were you planning to update this for DO usage ?

asymmetric:
@AmineChikhaoui I was actually using this for the [Hetzner Cloud backend](https://github.com/NixOS/nixops/pull/944/), I haven't tried Digital Ocean yet, but I can tomorrow.

Considering that [this](https://github.com/NixOS/nixpkgs/pull/59614) and [this](https://github.com/NixOS/nixops/pull/1131) have been merged though, I think we could go with 19.03.

asymmetric:
Upstream [switched](https://github.com/elitak/nixos-infect/commit/20853bf8d8872b4cb648796176e863a1b78c4d3f#diff-0210bc89d9d6d8c678b91aa8a839d65f) to 18.09, so I guess we can move on with this?

I can test 19.03 in another PR.

PR #1131, opened 2019-04-16
disable Azure backend :(

AmineChikhaoui:
It's unfortunate but I've gave the update a shot and concluded that it will need some non-trivial amount of work to revive it (more than just few hours). I'd be happy to remove this if someone creates a PR to make it work but for now I think it would be wise to move forward with this so that we can create a new release that would be usable with NixOS 19.03.

### DIFF
--- a/doc/manual/manual.xml
+++ b/doc/manual/manual.xml
@@ -396,6 +396,12 @@
 
     <section>
       <title>Azure Resources</title>
+      <warning>
+        <para>The Azure backend in Nixops is now disabled. See <link xlink:href="https://github.com/NixOS/nixops/pull/1131">PR#1131</link>
+        </para>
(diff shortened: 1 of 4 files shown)


AmineChikhaoui:
applied in https://github.com/NixOS/nixops/commit/cc74b57920df3daa10a78aa52c7283f4860a81e3

mwilsoninsight:
First step in an attempt to effectively revert this: https://github.com/NixOS/nixpkgs/pull/60435

toraritte:
Got here and to NixOps hoping to make it easy to deploy to Azure... Just in case someone needs it, there is an `azure-new` script in the Nixpkgs repo to do this:
https://github.com/NixOS/nixpkgs/tree/master/nixos/maintainers/scripts/azure-new

PR #1156, opened 2019-05-28
auto-luks: require new `mountPoint` option

andir:
With the upcoming systemd v242 change (https://github.com/NixOS/nixpkgs/pull/61321) in nixpkgs there will be a breaking change to how auto-luks used to work. We are going to remove a commit from the systemd fork (https://github.com/NixOS/systemd/commit/ce79214307f59fdd567271a0cad7be67e05c46ab) that was added for auto-luks to work. The
auto-luks feature will continue to work but another mount option is required to mark those `fileSystems` as not local. All mount points that carry the `_netdev` mount option are not part of the `local-fs.target` and thus not blocking the start of things like the sshd daemon.

Adding this option is (so far) the least intrusive way to ensure that systems will not break. Every user of the feature will be required to add only one more value to it's configuration.

Trying to automatically detect which mount points are mounted from the unlocked block devices is not really feasible in all/most cases. There might be another layer of indirection in between the actual (luks) device and the mount point. Good examples that will make it hard for us to guess are things like RAID, LVM, ZFS, ….

Dropping the systemd patch removes a bunch of issues and race conditions in regards to creating `/run/<service>` and `/var/lib/<service>` directories (see https://github.com/NixOS/nixpkgs/issues/47550, https://github.com/NixOS/nixpkgs/issues/53852, https://github.com/NixOS/nixpkgs/issues/53852). Those paths are usually set up using the `systemd-tempfiles.service` which is (again) part of `sysinit.target`.

This is also part of the ongoing effort to remove large a portion of our `PreStart` scripts that just create directories. In the long run there should be less custom scripting required to get a service with (temporary and persistent) state up and running.

For all the current version of NixOS this should be mostly a No-Op. In a local test against libvirtd there was no observable difference in behavior between 19.03 and the systemd PR.

Ideally we would have a way to warn users that deploy a more recent NixOS that they must use a newer NixOps version. It is still not obvious how to do that in an elegant way. Opinions are very welcome. Not updating systemd or keeping the broken sysinit behaviour isn't really an option in the long run.

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixops/1156)

### DIFF
--- a/nix/auto-luks.nix
+++ b/nix/auto-luks.nix
@@ -67,6 +67,25 @@ with utils;
               luksFormat</command>.
             '';
           };
+
+          mountPoint = mkOption {
+            type = types.either (types.enum ["none"]) types.str;
+            description = ''
+              This option is required so the autoLuks module knows where the
+              device will be mounted at.  With this option we inject the
+              required <literal>_netdev</literal> mount option. Without the
+              mount option the <literal>local-fs.target</literal> would fail.
+
+              Set to the path used in the corresponding
+              <literal>fileSystems.<replaceable>path</replaceable></literal>
+              option or to <literal>null</literal> if you know what you are doing.
+
+              If you do not use the device on a mount point pass <literal>none</literal>.
+
+              WARNING: failing to provide the correct value here might render
+              the system unbootable.
+            '';
+          };
         };
       });
       description = ''
@@ -85,6 +104,12 @@ with utils;
   ###### implementation
 
   config = {
+
+    fileSystems =
+      let
+        mkFileSystemEntry = _: attrs: nameValuePair attrs.mountPoint { options = [ "_netdev" ]; };
+      in mapAttrs' mkFileSystemEntry
+          (filterAttrs (_: attrs: attrs.mountPoint != "none") config.deployment.autoLuks);
 
     systemd.services =
       let


nix/auto-luks.nix
```diff
@@ -67,6 +67,24 @@ with utils;
               luksFormat</command>.
             '';
           };
+
+          mountPoint = mkOption {
+            default = false;
```
ip1981: The type is null or string.

nix/auto-luks.nix
```diff
@@ -67,6 +67,24 @@ with utils;
...
+            type = types.nullOr types.string;
```
ip1981: "string" is deprecated, it is better "str" or likely "path".

nix/auto-luks.nix
```diff
@@ -67,6 +67,24 @@ with utils;
...
+            description = ''
+              This option is required so the autoLuks module knows where the
```
ip1981: If required, why allow "null"?

edolstra:
Maybe we should just remove the auto-LUKS feature. I doubt anybody is using it.

nix/auto-luks.nix
andir: That was meant as an escape hatch for people that know what they are doing. There might be cases where it is being used without using it for a filesystem.

andir: I changed the type signature make it explicit that I want either a `str` or the string "none". While technically not required I think making the `"none"` case explicit in the type serves as a good way to document the existence of it.

nix/auto-luks.nix
andir: Thanks. Changed.

andir:
I addressed the review comments but In general I agree with @edolstra that we might just want to remove the feature.

It will still have to be somewhat synchronized with the NixOS releases so people are not creating unbootable systems by using a newer NixOS version from a very old NixOps.

grahamc:
On a call with Andi I suggested we _break_ autoluks as soon as possible in 19.03 stable, to try and get users to come out of the woodwork. If nobody comes out, we delete it.

AmineChikhaoui:
@edolstra @grahamc we use auto-luks at work for all our encrypted deployments.

andir:
@AmineChikhaoui Lets say we keep auto-luks but merge this PR. Would the additional attributes be okay / feasible for your deployments?

AmineChikhaoui:
@andir I'll double check and get back. I just wanted to reply that there are systems using this first, will read the details later.

AmineChikhaoui:
Just as an update: we discussed this in irc's #nixos-systemd channel (thanks to @flokli and @andir for the explanations ), the PR makes sense to me.
There is currently an infinite recursion issue when testing this PR, @andir mentioned that he will be looking at it.

flokli:
What's the status of this? Does the inifinite recursion still occur?

andir:
As far as I remember I tried to reproduce this and then ran into issues reproducing it. I do not have an AWS account or similar to actually test this properly. On the libvirt backend I wasn't able to reproduce.

flokli:
Well, this PR was meant to improve UX for people migrating from 19.03 to 19.09, alerting them for the missing `_netdev` in the mount options.

With nixpkgs now having branched off to 20.03 (and most people migrated to 19.09 at least), people did run into this, and this PR has probably become less useful.

If nobody can test this, we should probably just drop the idea, instead of just keeping this PR open.

AmineChikhaoui:
I'll try this again this week. Though I don't think it needs an AWS account to reproduce as the infinite recursion is an eval error.

grahamc:
Hello!

Thank you for this PR. 

NixOps:

   Here are some of them:
   
   required during CI.

on.

closing it now.

Graham

andir:
On 12:36 26.03.20, Graham Christensen wrote:
> Closed #1156.

Thank you for putting in the effort on NixOps. Appreciate it. It really
needs that love :)

I'll probably not re-open this (Nit-pick: I would have to create a new
PR for that and not just open it again). This has been done to help those that
use NixOps with the obscure auto-luks feature. I failed to reproduce the
issues and the tests didn't cover the feature… Nobody has really cared
in a while about it.

*Someone* should do the judgement call of retiring that feature if
nobody can fix it. It should be broken since a few releases anyway.
Currently we seem to be deadlocked on someone using it but not requiring
the fix.

Maybe it is worth having it downstream if your deployment depends on it.

PR #1163, opened 2019-06-10
Hashicorp Vault kv2 secret engine + policy resources

PsyanticY:


PsyanticY:
Changes included in Hashicorp vault Plugin here: https://github.com/PsyanticY/nixops-vault

PR #1169, opened 2019-06-27
manual: more options description fixes

AmineChikhaoui:


teto:
this may need a refresh, on current nixos-unstable I have the file in nixos/lib/make-options-doc/options-to-docbook.xsl which could explain the error:
```
while evaluating the attribute 'buildCommand' of the derivation 'options-db.xml' at /home/teto/nixpkgs/pkgs/build-support/trivial-builders.nix:7:14:
getting attributes of path '/home/teto/nixpkgs/nixos/doc/manual/options-to-docbook.xsl': No such file or directory
```

PR #1179, opened 2019-07-09
Nixops Core, Plugin Restructuring PR

johnalotoski:
This PR offers a restructuring of nixops to utilize a plugin architecture.  It follows on the initial work
of @grahamc  found at PR #981.

This restructuring maintains the core functionality of nixops, but only retains internal support for the
"None" backend.  The idea is that all other backends can be moved into their own repositories and act as plugins which can be pulled into nixops core as desired by the end user.  There are a number of reasons to consider this approach, such as those mentioned at issue #1158.

The [Pluggy](https://pluggy.readthedocs.io/en/latest/) plugin system, utilized by pytest, has continued to be utilized in this PR.

The code that was removed from nixops for AWS and Hetzner backend functionality has been placed into their own plugin repos: [nixops-aws](https://github.com/input-output-hk/nixops-aws/) and  [nixops-hetzner](https://github.com/input-output-hk/nixops-hetzner).  The code for each of these new plugin repos is largely the same as it was while a part of nixops with the exception of some pluggy hook implementation code and necessary namespace modifications to accommodate backend plugin code utilizing core nixops code.

In this PR, nixops is implementing three pluggy hooks:
1) load: returns a list of plugin modules to be loaded
2) nixexprs: returns a list of plugin Nix expressions to import
3) parser: used to pass parser and subparser objects to plugins to extend the CLI as needed.

To build nixops with plugins of interest, a similar approach to what Terraform in nixpkgs does was implemented.  In the nixops repo an 'all-plugins.txt' file describes repos where nixops plugins exist.  A script, 'update-all', can be run to check those repos for any plugin updates, detected as the
latest github plugin repo releases, and it records the detected plugins to a 'data.nix' file in a format consumable by fetchFromGitHub.

Nixops can be built with the desired plugins by issuing variants of the following commands where "plugin1" and "plugin2" would be example names of available plugins found in the data.nix file and the command containing callPackage points to a local plugin repo:

`nix-build release.nix -A build.x86_64-linux --arg p "(p: [ p.plugin1 ])"`
`./dev-shell --arg p "(p: [ p.plugin1 p.plugin2 ])"`
`./dev-shell --arg p "(p: [ (p.callPackage ../myplugin/release.nix {}) ])"`

[...]

PR #1186, opened 2019-07-24
[WIP]: Rewritting to use boto3

takeda:
The way how my AWS accounts are setup (assumeRole + MFA) makes it impossible to use old version of nixops. I initially tried to add an option to specify a profile, but turns out that old boto doesn't seem to support this kind of setup (since it was introduced later on) and it mandated porting to boto3.

So this version works fine with assumeRole + MFA and I also added an option to use the same cache that `aws` command uses. That way you don't have to enter MFA every for every single command.

Currently all of the commands that I used appear to be working, these are:
- list
- create
- modify
- delete
- info
- check
- deploy
- send-keys
- destroy
- stop
- start
- reboot
- show-arguments
- show-physical
- ssh
- ssh-for-each
- scp
- mount
- rename
- backup
- backup-status
- clean-backups
- restore (you need to provide backup id, but it looks like that was unrelated to my change?).
- show-option
- list-generations
- show-console-output
- dump-nix-paths
- export
- edit

If you can, please test it (clone my branch and run `nix-env -f release.nix -iA build.<your platform>`) if there are still broken commands let me know otherwise perhaps this can be merged.

### DIFF
--- a/nix/ec2-keypair.nix
+++ b/nix/ec2-keypair.nix
@@ -17,6 +17,11 @@ with lib;
       description = "AWS region.";
     };
 
+    profile = mkOption {
+      type = types.str;
+      description = "Name of the profile.";
+    };
+
     accessKeyId = mkOption {
       default = "";
       type = types.str;
--- a/nix/ec2.nix
+++ b/nix/ec2.nix
@@ -175,6 +175,17 @@ in
 
   options = {
 
+    deployment.ec2.profile = mkOption {
+      default = null;
+      example = "dev-Administrator";
+      type = types.str;
+      description = ''
+        The profile name as defined in <filename>~/.aws/config</filename>
+        If left empty it will be left to boto to decide, which means it will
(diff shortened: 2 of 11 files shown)


disassembler:
I'd like to see https://github.com/NixOS/nixops/pull/1179 finalized and get this PR targeting AWS plugin rather than merging this. Sorry for the extra work here, but in the long run I think you'll get faster response on major PRs like this when they only affect one backend.

takeda:
@disassembler the plugin architecture is orthogonal to this change though, I see that it still uses the unsupported boto package and still doesn't work correctly with assumeRole/MFA. If that's the way it will be restructured, perhaps my changes can be adapted to the aws plugin it seems like it was mostly moving aws components as is to the new repo.

takeda:
@disassembler so I moved my patches to nixops-aws, my PR is here: https://github.com/NixOS/nixops-aws/pull/1

edolstra:
So can we close this one?

takeda:
Yes, closing it.

Issue #931, opened 2018-04-23
permissions problems


teto:
I regularly seem to have ownership problems when using nixops. For instance` journalctl -b` would show

` Cannot execute '/etc/NetworkManager/dispatcher.d/02overridedns': not owned by root.`
or when setting zsh as a shell https://github.com/NixOS/nixpkgs/issues/39189 
Output from the nixops VM:
```
[root@client:~]# ls -lL /etc/NetworkManager/dispatcher.d/
total 12
-r-xr-xr-x 1 nobody nogroup  511 Jan  1  1970 02overridedns
-r-xr-xr-x 1 root   root    1251 Jan  1  1970 03userscript0001
-r-xr-xr-x 1 root   root     288 Jan  1  1970 03userscript0002
```
I configure the libvirtd service with
```
    qemuVerbatimConfig = ''
      namespaces = []
      # # Whether libvirt should dynamically change file ownership
      # # dynamic_ownership = 1
      user="teto"
      group="libvirtd"
    '';

```

sorki:
Try this fix I've got on IRC (looks like bug during VM store init):

```
mount -o remount,rw /nix/store
chown -R root:root /nix/store
```

teto:
Seems like I missed your answer. It worked great but if anyone has a hint on the root cause, it would be nice to fix. I've started to stack up a few patches on top of nixops to fix various things and would enjoy a better out of the box experience.

leshy:
seeing this as well on a fresh nixos 18.09 + nixops + libvirt

within guests:
```
sudo: error in /etc/sudo.conf, line 0 while loading plugin "sudoers_policy"
sudo: /nix/store/kqb1mmkmacysi9whzqjwqlmi3dqr5r5y-sudo-1.8.24/libexec/sudo/sudoers.so must be owned by uid 0
sudo: fatal error, unable to load plugins
```

pjones:
I have this in my `defaults` section as a nasty workaround:
```nix
system.activationScripts.nixops-vm-fix-931 = {
  text = ''
    if ls -l /nix/store | grep sudo | grep -q nogroup; then
      mount -o remount,rw  /nix/store
      chown -R root:nixbld /nix/store
    fi
  '';
  deps = [];
};

```

clefru:
[...]

PR #1187, opened 2019-07-30
Convert libvirtd to use make-disk-image

clefru:
This fixes https://github.com/NixOS/nixops/issues/931 by reusing the
image builder of <nixpkgs/nixos/lib/make-disk-image.nix>. This image
builder produces nixos.qcow2 instead of disk.qcow2. This image is now
"baseless", so it obsoletes "qemu-img rebase". We can also remove our
special-case image builder libvirtd-image.nix as a whole.


teto:
thanks for fixing this. The PR looks sound. I have a few trouble with my nixops setup at the moment but I will try to give it a run next week !

I am concerned abouyt the removal of  nix/libvirtd-image.nix . I have pending changes that make use of qemuGuest in libvirtd-image.nix . If this gets removed, where am I supposed to move the declaration

clefru:
@teto Can you show me your modifications? I couldn't find them in your pull requests or in github.com/teto/nixops.

teto:
for instance this line is new (the filename has changed but it's the same :  https://github.com/teto/nixops/blob/ad4c1866b88075bd6b28d2c9f4e39d74a8ca2027/nix/libvirtd/image.nix#L22).
It would be nice to keep having some libvirtd specific configuration ? even though this might be a bad example as it's useful outside libvirt, as soon as we use qemu in fact.

clefru:
We already have libvirtd specific options even with this image builder. Just stick "services.qemuGuest.enable = true;" to the bottom of the file in libvirtd.nix inside the config set. The image builder will pick that up.

teto:
this looks good, can we merge ?

clefru:
From my point of view, yes :)

teto:
Now that nixops  backends are implemented as plugins, you should reopen the PR against https://github.com/nix-community/nixops-libvirtd (I believe).

clefru:
Meh, sorry. I don't feel like testing this again. I'll drop the PR entirely. and keep nixops as is locally.

PR #1197, opened 2019-09-12
log more commands

teto:
while developing on nixops, I regularly have issues and logging the run command helps a lot in debugging.
I wonder if nixops should have different levels but it would be cool to better log commands. 
(I can remove the NIX_SSHOPTS)


teto:
I have removed the controversial logging and logged an additionnal ssh command. Looking forward to see this merged.

AmineChikhaoui:
I don't think emitting verbose logging by default is a good idea ? this should at least be done only if --debug is set imo.

teto:
sure. Fixed.

teto:
really addressed the comment this time.

PR #1198, opened 2019-09-16
Improves update-all script and adds related docs

johnalotoski:
* Adds shellcheck and related fixes to update-all
* Makes github token usage optional with update-all
* Prints github token availability and reset time
* Provides github token setup info on rate limit error
* Rolls back to original data.nix upon script error
* Trap the exit for temp file clean up
* Adds some additional clarifications to README.md
* Adds docs and examples to all-plugins.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,17 +20,25 @@ To start developing on nixops, you can run:
 ```
 
 Where plugin1 can be any available nixops plugin, and where
-none or more than one can be specified, including local plugins.  An example is:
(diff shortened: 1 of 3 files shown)


edolstra:
IMHO it would be better to add a `flake.nix` to NixOps and plugins than build an ad hoc updater.

johnalotoski:
Hi @edolstra.  Maybe until we get a flake in place to handle this we can consider this as an interim improvement since it's already available?

PR #1201, opened 2019-10-03
Adding nixops delete-resources command

PsyanticY:
This command will allow users to painlessly remove resources from nixops database while preserving the actual resource. This can be helpful in various disater recovery scenarios when we need to attach an EBS volume to prod form shadow or vice versa. then can safely reeuse that deployment since the shadow won't clain the volume transfered to prod as it's own and tires to destory it

### DIFF
--- a/nixops/resources/__init__.py
+++ b/nixops/resources/__init__.py
@@ -189,6 +189,20 @@ class ResourceState(object):
         )
         return False
 
+    def delete_resources(self):
+        """delete this resource state, if possible."""
+        if not self.depl.logger.confirm(
+                "are you sure you want to clear the state of {}? "
+                "this will only remove the resource from the local "
+                "NixOPS state and the resource may still exist outside "
+                "of the NixOPS database.".format(self.name)):
+            return False
+
+        self.logger.warn(
+            "removing resource {} from the local NixOPS database ...".format(self.name)
+        )
+        return True
+
     def next_charge_time(self):
         """Return the time (in Unix epoch) when this resource will next incur
         a financial charge (or None if unknown)."""
--- a/scripts/nixops
+++ b/scripts/nixops
@@ -81,6 +81,11 @@ subparser.add_argument('--exclude', nargs='+', metavar='MACHINE-NAME', help='des
 subparser.add_argument('--wipe', action='store_true', help='securely wipe data on the machines')
 subparser.add_argument('--all', action='store_true', help='destroy all deployments')
 
+subparser = add_subparser(subparsers, 'delete-resources', help='deletes the resource from the local NixOPS state file.')
+subparser.set_defaults(op=op_delete_resources)
+subparser.add_argument('--include', nargs='+', metavar='RESOURCE-NAME', help='delete only the specified resources')
+subparser.add_argument('--exclude', nargs='+', metavar='RESOURCE-NAME', help='delete all resources except the specified resources')
+
 subparser = add_subparser(subparsers, 'stop', help='stop all virtual machines in the network')
 subparser.set_defaults(op=op_stop)
 subparser.add_argument('--include', nargs='+', metavar='MACHINE-NAME', help='stop only the specified machines')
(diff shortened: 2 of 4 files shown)


nixops/resources/__init__.py
```diff
@@ -189,6 +189,17 @@ def destroy(self, wipe=False):
         )
         return False
 
+    def clear_state(self):
+        """clear this resource state, if possible."""
+        if not self.depl.logger.confirm("are you sure you want to clear resources state `{0}`".format(self.name)):
```
AmineChikhaoui: I'd prefer if we emit a more clear warning of what this does, something like:
```suggestion
        if not self.depl.logger.confirm("are you sure you want to clear resources state {}: this will only remove the resource from the local NixOPS state and the resource may still exist outside of the NixOPS database.".format(self.name)):
```

nixops/resources/__init__.py
```diff
@@ -189,6 +189,17 @@ def destroy(self, wipe=False):
...
+            self.depl.logger.get_logger_for(self.name).warn("Not clearing state...")
```
AmineChikhaoui: I don't think this is needed
```suggestion
```

nixops/resources/__init__.py
```diff
@@ -189,6 +189,17 @@ def destroy(self, wipe=False):
...
+            return False
+
+        self.logger.warn(
+            "resource ‘{0}’ will be removed from NixOps state file but not actually".format(self.name)
```
AmineChikhaoui:
```suggestion
            "removing resource {} from the local NixOPS database ...".format(self.name)
```

nixops/resources/__init__.py
```diff
@@ -189,6 +189,16 @@ def destroy(self, wipe=False):
         )
         return False
 
+    def clear_state(self):
+        """clear this resource state, if possible."""
+        if not self.depl.logger.confirm("are you sure you want to clear resources state {}: this will only remove the resource from the local NixOPS state and the resource may still exist outside of the NixOPS database.".format(self.name)):
```
AmineChikhaoui: Can you split into multiple lines, something like:
```python
def clear_state(self):
    """clear this resource state, if possible."""
    if not self.depl.logger.confirm(
            "are you sure you want to clear resources state {}:"
            "this will only remove the resource from the local "
            "NixOPS state and the resource may still exist outside"
            " of the NixOPS database.".format(self.name)):
        return False
```

scripts/nixops
```diff
@@ -81,6 +81,11 @@ subparser.add_argument('--exclude', nargs='+', metavar='MACHINE-NAME', help='des
 subparser.add_argument('--wipe', action='store_true', help='securely wipe data on the machines')
 subparser.add_argument('--all', action='store_true', help='destroy all deployments')
 
+subparser = add_subparser(subparsers, 'clear-state', help='clear resources from state file')
```
AmineChikhaoui:
```suggestion
subparser = add_subparser(subparsers, 'clear-state', help='deletes the resource from the local NixOPS state file.')
```

scripts/nixops
```diff
@@ -81,6 +81,11 @@ subparser.add_argument('--exclude', nargs='+', metavar='MACHINE-NAME', help='des
...
+subparser.set_defaults(op=op_clear_state)
+subparser.add_argument('--include', nargs='+', metavar='MACHINE-NAME', help='clear only the specified machines')
```
AmineChikhaoui:
```suggestion
subparser.add_argument('--include', nargs='+', metavar='RESOURCE-NAME', help='delete only the specified resources')
```

scripts/nixops
```diff
@@ -81,6 +81,11 @@ subparser.add_argument('--exclude', nargs='+', metavar='MACHINE-NAME', help='des
...
+subparser.add_argument('--exclude', nargs='+', metavar='MACHINE-NAME', help='clear all resources except the specified machines')
```
AmineChikhaoui:
```suggestion
subparser.add_argument('--exclude', nargs='+', metavar='RESOURCE-NAME', help='delete all except the specified resources')
```

PR #1203, opened 2019-10-10
Restore commandOutput resources

dsg22:
The customCommand resource is no longer usable since acb2ab15ac85189758aa530a76b7ec3cc7ceb295.

I'm not sure if this is intentional or if it was accidentally removed from eval-machine-info.nix during the plugin restructuring. Assuming it was not intentional, this PR re-enables it.

After applying, tests/functional/single_machine_outputs.nix should be deployable.


AmineChikhaoui:
Thanks. Though the test expression doesn't work out of the box, I think we need to have a default value for `resources.commandOutput.<name>.value` so that the first eval works which evaluates the network info.

PR #1214, opened 2019-12-19
Upgrade to Python3

Taneb:
I haven't tested this executes properly but it passes all the tests on Linux.

There's a note in `nixops/diff.py` suggesting some improvements to the type declarations possible in python3 which I haven't dealt with.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -312,7 +312,7 @@ class MachineState(nixops.resources.ResourceState):
 
     def write_ssh_private_key(self, private_key):
         key_file = "{0}/id_nixops-{1}".format(self.depl.tempdir, self.name)
-        with os.fdopen(os.open(key_file, os.O_CREAT | os.O_WRONLY, 0600), "w") as f:
+        with os.fdopen(os.open(key_file, os.O_CREAT | os.O_WRONLY, 0o600), "w") as f:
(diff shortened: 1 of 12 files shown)


Taneb:
The CI failure seems to be unrelated to my PR, it looks like the nix Travis config now creates `/etc/nix` for us so when we try to do it in the `.travis.yml` pre-step it fails.

lheckemann:
Did you use 2to3 for the syntactic/trivial changes? Why are they split into two commits?

As for the unit test target, I think it would be better to change `assert_merge` to compare `set(functools.reduce(…))` and `set(expect)` instead, to better encode that we don't guarantee an order.

Taneb:
@lheckemann I did not use 2to3, the changes were manual. I agree with the solution to the test issue.

grahamc:
I'm seeing a lot of failures around `.iteritems()` which should be `.items()` instead. Why not use 2to3?.

grahamc:
It looks like the by-hand / partial conversion of deployment.py breaks 2to3.

grahamc:
I did some more work on it here https://github.com/NixOS/nixops/compare/master...grahamc:python3 and make nixops info work, and nixops ssh-for-each work, but nixops deploy hangs forever with defunct ssh processes.

PR #1228, opened 2020-02-21
travis: check files are formatted correctly with `black`

grahamc:


grahamc:
If this merges, I'll point my 2to3 branch to this: https://github.com/grahamc/nixops/compare/format...grahamc:python-2to3-formatted

@samueldr pointed out it makes sense to do these at the same time, in this order, since the 2->3 transition is so mechanical already.

grahamc:
I got a +1 from @AmineChikhaoui to merge this.

PR #1233, opened 2020-02-29
Start with types

TheWizardTower:
A few commits. The first batch are cleanup-things of lint flags that my editor told me about; the last one is the one that changes business logic.

I added type annotations to nixops/nixops/known_hosts.py.

I haven't tested this at all, as I'm brand new to the repo and to nixos in general. I'm just fairly familiar with working with static type systems. :)

Looking forward to finding what I've broken.

### DIFF
--- a/nixops/known_hosts.py
+++ b/nixops/known_hosts.py
@@ -2,12 +2,13 @@ import os
 import threading
 import fcntl
 
+from typing import Optional
 
 # Allow only one thread to rewrite known_hosts at a time.
 LOCK = threading.Lock()
 
 
-def _rewrite(ip_address, add, public_host_key):
+def _rewrite(ip_address: str, add_ip: bool, public_host_key: str) -> None:
     with LOCK:
         path = os.path.expanduser("~/.ssh/known_hosts")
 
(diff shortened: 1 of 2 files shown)


grahamc:
Thank you a lot!

nixops/known_hosts.py
```diff
@@ -53,20 +53,18 @@ def rewrite(l):
...
-def update(prev_address, new_address, public_host_key):
-    assert public_host_key is not None
+def update(prev_address: str, new_address: str, public_host_key: str) -> None:
     # FIXME: this rewrites known_hosts twice.
-    if prev_address is not None and prev_address != new_address:
+    if prev_address != new_address:
         remove(prev_address, public_host_key)
-    if new_address is not None:
-        add(new_address, public_host_key)
+    add(new_address, public_host_key)
```
bhipple: As a heads up, this is a backwards incompatible API change to the function.  I think it's a good change, but previously plugins like [nixops-aws were calling "update" with None values to mean "remove" or "add"](https://github.com/NixOS/nixops-aws/pull/38).  I've fixed nixops-aws by having it call `add` or `remove` directly, but it may be worth scanning other plugins to see how they're calling `update`.

PR #1234, opened 2020-02-29
Start with types

TheWizardTower:
Me, again. More stuff I want some more eyeballs on.

Most of this is in nixops/diff.py, which already had some typing work done, but I think I found a few improvements.

Namely:

* Removed some unused imports.
* Re-shuffled the class definitions. I can't get Handler to show up as a type Python knows about inside Handler, but if Handler comes first, then Diff methods can reference it.
* ~~Solved a todo that was pending python3.~~ JKLOL nope.
* Changed the type annotations from comments to explicit in the function defs.
* Changed a few if conditionals to use 'not expr' or 'if exper is not None' rather than !=
* Removed an unneeded else.

Not sure if this is at all useful, I'm just following mypy and flycheck.

Looking forward to your critical feedback. :)

### DIFF
--- a/nixops/known_hosts.py
+++ b/nixops/known_hosts.py
@@ -2,12 +2,13 @@ import os
 import threading
 import fcntl
 
+from typing import Optional
 
 # Allow only one thread to rewrite known_hosts at a time.
 LOCK = threading.Lock()
 
 
-def _rewrite(ip_address, add, public_host_key):
+def _rewrite(ip_address: str, add_ip: bool, public_host_key: str) -> None:
     with LOCK:
         path = os.path.expanduser("~/.ssh/known_hosts")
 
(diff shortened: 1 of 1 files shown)


nixops/diff.py
```diff
@@ -31,12 +30,10 @@ def _default_handle(self):
         """
         raise NotImplementedError
 
-    def get_deps(self):
-        # type: () -> List[Handler]
+    def get_deps(self) -> List[Handler]:
         return self._dependencies
 
     def get_keys(self, *_: AnyStr) -> List[str]:
```
grahamc: I don't even know what this code is doing :o

TheWizardTower: To be honest, me neither. :)

nixops/diff.py
```diff
@@ -30,7 +30,9 @@ def _default_handle(self):
         """
         raise NotImplementedError
 
-    def get_deps(self) -> List[Handler]:
+    # This is List[Handler], but that name doesn't exist inside the
+    # class definition.
+    def get_deps(self) -> List[Any]:
```
grahamc: I think this can be `List[Handler]` with `from __future__ import annotations`, give that a try?

nixops/diff.py
AmineChikhaoui: A handler is basically a handler of few attributes e.g `handler(['region', 'instanceType'])`, which means it realizes the needed changes when those attributes are changed,
I think get_keys is just convenience to return the keys, don't remember why I added it to be honest :D

AmineChikhaoui: or are you guys asking about `*_: AnyStr` which I don't what it does either :-)

PR #1235, opened 2020-02-29
Start with types rebase

TheWizardTower:
Let's try this again.

I think I vaporized my other branch, which nullified my existing PR. But. This is the same code.

### DIFF
--- a/nixops/diff.py
+++ b/nixops/diff.py
@@ -12,7 +12,6 @@ class Handler:
         after: Optional[List] = None,
         handle: Optional[Callable] = None,
     ) -> None:
-        # type: (List[str], Optional[List], Optional[Callable]) -> None
         if after is None:
             after = []
         if handle is None:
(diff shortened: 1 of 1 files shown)


grahamc:
I wonder if that __future__ patch makes the nixops.deployment.Deployment type hint work, too. I've wanted to do that elsewhere but got bit by the circular dependency.

TheWizardTower:
It looks like it passes mypy on my machine? No harm in giving it a go here.

grahamc:
Blah. Dang. Burned by reality. Let's undo that and merge :)

TheWizardTower:
```
ImportError: cannot import name 'Deployment' from 'nixops.deployment' (/build/nixops-1.8pre0_abcdef/nixops/deployment.py)
```

I think this is toast. :(

will un-merge.

grahamc:
Thank you, this is amazing!

PR #1236, opened 2020-03-01
deployment.py: more mypy type checking

AmineChikhaoui:


nixops/deployment.py
```diff
@@ -1272,31 +1323,34 @@ def worker(r):
                     )
 
                     if is_machine(r):
+                        # NOTE: unfortunate mypy doesn't check that
+                        # is_machine calls an isinstance() function
+                        m = cast(nixops.backends.MachineState, r)
```
grahamc: In other places, I replaced is_machine() calls with an instanceof of my own. Maybe ugly ... I wonder if there is a way we could fix this without cast?

nixops/deployment.py
```diff
@@ -1374,16 +1428,18 @@ def cleanup_worker(r):
...
         pass
 
-    def notify_failed(self, action, e):
+    def notify_failed(
+        self, action: str, e: Union[KeyboardInterrupt, Exception]
+    ) -> None:
         pass
 
-    def run_with_notify(self, action, f):
+    def run_with_notify(self, action: str, f: Callable) -> None:
```
grahamc:
```suggestion
    def run_with_notify(self, action: str, f: Callable[[], None]) -> None:
```

PR #1238, opened 2020-03-02
Even more mypy type checking

doshitan:
It started out a little more disciplined then became a big ball of changes, apologies.

Also some of this might be overkill.

Included the stricter mypy config that this passes for reference, but it's maybe we don't want to require it yet. 

Running mypy (with stricter config):
image

The mypy report part errors out early, but what it does get to:
```
+--------------------------+-------------------+----------+
| Module                   | Imprecision       | Lines    |
+--------------------------+-------------------+----------+
| nixops                   |   0.00% imprecise |    0 LOC |
| nixops.ansi              |   0.00% imprecise |   18 LOC |
| nixops.diff              |  15.17% imprecise |  211 LOC |
| nixops.known_hosts       |  11.43% imprecise |   70 LOC |
| nixops.logger            |   0.00% imprecise |  158 LOC |
| nixops.nix_expr          |  27.53% imprecise |  385 LOC |
| nixops.parallel          |   9.48% imprecise |  116 LOC |
| nixops.plugins           |  58.33% imprecise |   12 LOC |
| nixops.plugins.hookspecs |  15.38% imprecise |   26 LOC |
| nixops.resources         |  15.31% imprecise |  320 LOC |
| nixops.script_defs       |  28.99% imprecise | 1128 LOC |
| nixops.ssh_util          |   5.62% imprecise |  409 LOC |
| nixops.state             |  18.07% imprecise |   83 LOC |
| nixops.util              |  10.83% imprecise |  600 LOC |
+--------------------------+-------------------+----------+
| Total                    |  18.30% imprecise | 3536 LOC |
+--------------------------+-------------------+----------+
```

Tests:
image

@grahamc saw your tweet about working on this and having some experience hitting python over the head with mypy, figured I'd make a pass (^_^)

### DIFF
--- a/nixops/logger.py
+++ b/nixops/logger.py
@@ -1,30 +1,30 @@
 # -*- coding: utf-8 -*-
 import sys
 import threading
-from typing import TextIO, List
+from typing import List, Optional, TextIO
 
-from nixops.util import ansi_warn, ansi_error, ansi_success
+from nixops.ansi import ansi_warn, ansi_error, ansi_success
 
 __all__ = ["Logger"]
 
 
 class Logger(object):
     def __init__(self, log_file: TextIO) -> None:
-        self._last_log_prefix = None  # XXX!
+        self._last_log_prefix: Optional[str] = None  # XXX!
         self._log_lock = threading.Lock()
         self._log_file = log_file
-        self._auto_response = None
+        self._auto_response: Optional[str] = None
         self.machine_loggers: List[MachineLogger] = []
 
     @property
-    def log_file(self):
+    def log_file(self) -> TextIO:
         # XXX: Remove me soon!
         return self._log_file
 
-    def isatty(self):
+    def isatty(self) -> bool:
         return self._log_file.isatty()
 
-    def log(self, msg):
+    def log(self, msg: str) -> None:
         with self._log_lock:
             if self._last_log_prefix is not None:
                 self._log_file.write("\n")
@@ -32,7 +32,7 @@ class Logger(object):
             self._log_file.write(msg + "\n")
             self._log_file.flush()
 
-    def log_start(self, prefix, msg):
+    def log_start(self, prefix: str, msg: str) -> None:
         with self._log_lock:
             if self._last_log_prefix != prefix:
                 if self._last_log_prefix is not None:
@@ -42,7 +42,7 @@ class Logger(object):
             self._last_log_prefix = prefix
             self._log_file.flush()
 
-    def log_end(self, prefix, msg):
+    def log_end(self, prefix: str, msg: str) -> None:
         with self._log_lock:
             last = self._last_log_prefix
             self._last_log_prefix = None
@@ -55,7 +55,7 @@ class Logger(object):
             self._log_file.write(msg + "\n")
             self._log_file.flush()
 
-    def get_logger_for(self, machine_name):
+    def get_logger_for(self, machine_name: str) -> "MachineLogger":
         """
         Returns a logger instance for a specific machine name.
         """
@@ -64,7 +64,7 @@ class Logger(object):
         self.update_log_prefixes()
         return machine_logger
 
-    def set_autoresponse(self, response):
+    def set_autoresponse(self, response: str) -> None:
         """
         Automatically respond to all confirmations with the response given by
         'response'.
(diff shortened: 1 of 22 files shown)


grahamc:
This is amazing! There are some module import errors which I can help track down and fix. Thank you!

nixops/parallel.py
```diff
@@ -40,9 +40,9 @@ def print_all_backtraces(self):
 
 def run_tasks(
     nr_workers: int, tasks: Iterable[Task], worker_fun: Callable[[Task], Result]
-) -> FinalResult:
+) -> FinalResult[Result]:
     task_queue: queue.Queue[Task] = queue.Queue()
-    result_queue: queue.Queue[WorkerResult] = queue.Queue()
+    result_queue: queue.Queue[WorkerResult[Result]] = queue.Queue()
```
grahamc: I'm a bit surprised -- is FinalResult / WorkerResult a generic type?

nixops/logger.py
```diff
@@ -55,7 +55,7 @@ def log_end(self, prefix, msg):
             self._log_file.write(msg + "\n")
             self._log_file.flush()
 
-    def get_logger_for(self, machine_name):
+    def get_logger_for(self, machine_name: str) -> "MachineLogger":
```
grahamc: add:

```
from __future__ import annotations
```

to the  first first line of code (after the encoding line) and this can just be `MachineLogger`.

nixops/nix_expr.py
```diff
@@ -1,55 +1,64 @@
...
-from typing import Optional
+from typing import Any, Dict, List, Optional, Sequence, Tuple, TypeVar, Union, overload
 from textwrap import dedent
 
+import sys
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
TheWizardTower: I think this is the source of the build issue. I'm not yet sure how to fix it, though.

nixops/parallel.py
TheWizardTower: WorkerResult is defined a little further up, it's basically an Either from Haskell, or std::variant from C++.

nixops/nix_expr.py
```diff
@@ -156,13 +177,29 @@ def py2nix(value, initial_indentation=0, maxwidth=80, inline=False):
...
+    ) -> Union[RawValue, Container]:
+        ...
+
+    @overload
+    def _enc_str(node: str) -> Union[RawValue, Container]:
+        ...
+
+    def _enc_str(
+        node: str, for_attribute: bool = False
+    ) -> Union[str, RawValue, Container]:
```
TheWizardTower: I'm curious to try this overall PR without the `Literal` import, given that it's only used in two places. If that works, we can merge the rest and make this it's own PR and figure out if/how python does dependency management.

nixops/nix_expr.py
```diff
@@ -1,55 +1,64 @@
...
-    def is_inlineable(self):
+    def is_inlineable(self) -> bool:
         return True
 
-    def indent(self, level=0, inline=False, maxwidth=80):
+    def indent(self, level: int = 0, inline: bool = False, maxwidth: int = 80) -> str:
         return "  " * level + self.value
 
-    def __repr__(self):
+    def __repr__(self) -> str:
```
grahamc: I'm a bit unsure about this -- we say we're returning a str, but above we accept an Any in the constructor. What do you think?

doshitan: My understanding of Python is that `__repr__` should always return a string (and typeshed seems to agree). But yea, strange that it just returns the raw value and I don't know if the behavior is relied on somewhere :/

nixops/nix_expr.py
doshitan: Yea, I'd kinda prefer separate functions with it clear in the name what they do/return, rather than relying on a special parameter, but the pattern was present in a variety of places and I don't know if there are calls to this stuff (particularly the `logged_exec` calls/wrappers) outside of this code base, so was hesitant to change function names.

If we can refactor and do away with all the `Literal` junk without also adding a bunch of `cast`s, I think that'd be a win.

doshitan:
Welp, down to just one coverage test error (due to more import cycle shenanigans).

grahamc:
I'm remiss to close this, but it is a bit out of date from master and the remaining commits are tricky to merge. Maybe the work here could be broken up in to a few smaller and related chunks?

doshitan:
Yea, I've not had a chance to get back to this lately, if you've already lifted what useful bits you could then feel free to close! I may have time this weekend to rebase/pick out any standalone chunks, but can't promise anything.

grahamc:
If you rebase, go ahead and reopen? Thank you!

PR #1244, opened 2020-03-06
script_defs: Remove None assertion

adisbladis:
`r` being `None` is peferctly valid when the resource does not yet exist.

Without this fix a resource not yet deployed fails with AssertionError.

### DIFF
--- a/nixops/script_defs.py
+++ b/nixops/script_defs.py
@@ -241,9 +241,7 @@ def op_info(args):
 
             resource_state: str = "Missing"
             if isinstance(r, nixops.backends.MachineState):
-                resource_state = "{0} / {1}".format(
-                    r.show_state() if r else "Missing", state(depl, d, r)
-                )
+                resource_state = "{0} / {1}".format(r.show_state(), state(depl, d, r))
             elif r:
                 resource_state = r.show_state()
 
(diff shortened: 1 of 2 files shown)


adisbladis:
@grahamc Updated PR with more typing.

nixops/script_defs.py
```diff
@@ -236,20 +238,20 @@ def name_to_key(name: str) -> Tuple[str, str, List[object]]:
         for name in names:
             d = definitions.get(name)
             r = depl.resources.get(name)
-            assert r is not None
-            if deployment.is_machine(r):
+
+            resource_state: str = "Missing"
+            if r is not None and deployment.is_machine(r):
```
grahamc: The logic here and in `state()` was freaking me out, so I traced it out a bit and tried to simplify it some: https://github.com/adisbladis/nixops/pull/1 I'm not ... 100% ... confident ... that it is the same thing. Take a look?

adisbladis: The logic in `state` is not exactly the same, instead of implicitly returning `None` i raise an error.
The old behaviour in that function would result in a string containing `None` (as a string).

adisbladis: This block behaves the same as before adding the assertion.

grahamc:
resources: 45.58% imprecise ->41.50% imprecise
script_defs: 58.85% imprecise -> 57.95% imprecise

PR #1247, opened 2020-03-10
Make it possible to use Nixops without automatic SSH key provisioning

adisbladis:


nix/options.nix
```diff
@@ -39,6 +39,15 @@ in
       '';
     };
 
+    deployment.provisionSSHKey = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        This option specifices whether to let Nixops provision SSH deployment keys.
```
grahamc:
```suggestion
        This option specifices whether to let NixOps provision SSH deployment keys.
```

though maybe should describe more about the history, and default behavior, etc.

nixops/backends/none.py
```diff
@@ -31,6 +35,7 @@ class NoneState(MachineState):
     def get_type(cls):
         return "none"
 
+    provison_ssh_key = nixops.util.attr_property("provisionSSHKey", None, bool)
```
grahamc:
```suggestion
    provision_ssh_key: bool = nixops.util.attr_property("provisionSSHKey", None, bool)
```

(and the fall-out of this rename)

nixops/backends/none.py
```diff
@@ -66,13 +71,19 @@ def create(self, defn, check, allow_reboot, allow_recreate):
...
 
         if not self.vm_id:
-            self.log_start("generating new SSH keypair... ")
-            key_name = "NixOps client key for {0}".format(self.name)
-            self._ssh_private_key, self._ssh_public_key = create_key_pair(
-                key_name=key_name
-            )
+            self._ssh_private_key = None
+            self._ssh_public_key = None
+
```
grahamc:
```suggestion
            self._ssh_private_key: Optional[str] = None
            self._ssh_public_key: Optional[str] = None

```

adisbladis:
> You and I were talking about this a few days ago, and this is what I wrote up then. I wonder if there is a way to author this PR in a way way to build out support for the other use cases?

I actually think we end up with _more_ flexibility by not supporting each use case explicitly.
I think the most reasonable and flexible approach is to leave the specifics of authentication up to the user.

By doing this we automatically support most use cases that supports SSH agent. 
PKCS11 is supported natively by OpenSSH too so that's supported with this approach already too.

I don't know about using Vault, that may require explicit support from Nixops.

tomberek:
There is some overlap between this and #1048 #1054. A cohesive solution would be great.

adisbladis:
Rebased on latest master

chreekat:
> I actually think we end up with _more_ flexibility by not supporting each use case explicitly.
> I think the most reasonable and flexible approach is to leave the specifics of authentication up to the user.

While working on a similar problem recently, I realized that accepting and/or generating an entire openssh config file is a really flexible option! In my utility, I was able to clean up all my ssh invocations this way (simply `ssh -F $config_file`) .

adisbladis:
> While working on a similar problem recently, I realized that accepting and/or generating an entire openssh config file is a really flexible option!

This has been somewhat addressed in https://github.com/NixOS/nixops/pull/1270 with the addition of `sshOptions`.

nix/options.nix
```diff
@@ -49,6 +49,21 @@ in
...
     };
 
+    deployment.provisionSSHKey = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        This option specifies whether to let NixOps provision SSH deployment keys.
+
+        NixOps will by default generate an SSH key, store it in its state file
+        and deploy it to the remote host.
```
grahamc, line 59:
```suggestion
        NixOps will by default generate an SSH key, store the private key in its state file,
        and add the public key to the remote host.
```

nixops/backends/none.py
```diff
@@ -6,12 +6,15 @@
...
 
+from typing import Optional
+
 
 class NoneDefinition(MachineDefinition):
     """Definition of a trivial machine."""
 
     _target_host: str
     _public_ipv4: Optional[str]
+    _provision_ssh_key: bool
```
grahamc, line 17: Let's move this to MachineDefinition's base implementation, and from there plugins need to handle it (and failing to handle it can be considered a bug.)

nixops/backends/none.py
```diff
@@ -32,10 +34,11 @@ class NoneState(MachineState):
     def get_type(cls):
         return "none"
 
+    provison_ssh_key: bool = nixops.util.attr_property("provisionSSHKey", None, bool)
```
grahamc, line 37:
```suggestion
    provison_ssh_key: bool = nixops.util.attr_property("provisionSSHKey", true, bool)
```

and I wonder about making the attr_property return type be validated against its default

nixops/backends/none.py
```diff
@@ -90,18 +93,21 @@ def get_ssh_name(self):
     def get_ssh_private_key_file(self):
```
grahamc, line 93:
```suggestion
    def get_ssh_private_key_file(self) -> Optional[Path]
```

could we type this and `write_ssh_private_key` a bit?

PR #1240, opened 2020-03-02
Convert scripts/nixops to setuptools entry point

adisbladis:
This is more idiomatic & makes it possible to run `python -m nixops`.

PR #1248, opened 2020-03-11
WIP: Add deploy without creating a state file

adisbladis:
Create an in-memory SQLite db if `--no-state` is passed to deploy.
This is probably the easiest way possible for us to add support for deployments without using a state file.

If `--no-state` is passed a create is implicitly done in the same process.

While this works perfectly fine with the `none` backend it's marked WIP because it needs docs.

Based on top of https://github.com/NixOS/nixops/pull/1240 since they touch a lot of the same code & meant to be used together with https://github.com/NixOS/nixops/pull/1247.

### DIFF
--- a/nixops/__main__.py
+++ b/nixops/__main__.py
@@ -164,6 +164,7 @@ subparser.add_argument(
     help="activate unchanged configurations as well",
 )
 add_common_deployment_options(subparser)
+add_common_modify_options(subparser)
 
 subparser = add_subparser(subparsers, "send-keys", help="send encryption keys")
 subparser.set_defaults(op=op_send_keys)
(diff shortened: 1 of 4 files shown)


cleverca22:
main thing i can see being lost with this change, is that the `stateVersion` is normally queried on the first deploy and held in the state file, to deploy compatible nixos's in the future

but the user must now manage `stateVersion` themselves, in the deployment files

grahamc:
I just force-pushed a rebase on to master.

adisbladis:
https://github.com/NixOS/nixops/pull/1264 which builds on this is a better and more comprehensive solution. Closing.

PR #1251, opened 2020-03-12
ci: init

lovesegfault:
This adds a `ci.nix` which is able to generate a `ci.yml` for use with
github actions.
Usage: `nix-build ci.nix --no-out-link | bash`

The main idea behind generating the yaml with nix, as opposed to just
writing yaml, is that Nix is easier to write and maintain. It would've
taken a lot more copy-paste to achieve the same done here in raw gh
actions yaml.


ci.nix
```diff
@@ -0,0 +1,58 @@
...
+        name = "Black";
+        run = "nix-shell --run \"black . --check --diff\"";
+      }];
+      coverage = mkJob [{
+        name = "Coverage";
+        run = "nix-shell --exclude tarball --run \"./coverage-tests.py -a '!libvirtd,!gce,!ec2,!azure' -v\"";
+      }];
+      build = mkJob [{
+        name = "Build";
+        run = "nix-build --quiet release.nix -A build.x86_64-linux -I nixpkgs=channel:19.09";
```
bjornfor: Since release.nix already has a default nixpkgs, isn't adding -I nixpkgs=... asking for trouble?

ci.nix
```diff
@@ -0,0 +1,58 @@
...
+      }];
+      ciCheck = mkJob [{
+        name = "Check CI";
+        run = ''
+          cp ${ciPath} /tmp/ci.reference.yml
+          nix-build ci.nix --no-out-link | bash
+          diff ${ciPath} /tmp/ci.reference.yml || exit 1
```
bjornfor: The generated yml/json file is one long line. Perhaps prettifying it with jq makes it more readable and results in easier to read diffs?

ci.nix
lovesegfault: This was copied over from `.travis.yml`

ci.nix
lovesegfault: Deferring to @grahamc, I'd be fine with that

edolstra:
Not sure this is a good idea. The generated file may be more verbose, but it's easier to understand. E.g. if I now need to change something to the GitHub actions, I have to figure out exactly what `ci.nix` is doing. (Lisp curse and all that.)

PR #1253, opened 2020-03-12
More mypy, fix a couple of return types

aaronchall:
Feel free to edit further before merging if you like.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -9,6 +9,7 @@ import nixops.resources
 import nixops.ssh_util
 import xml.etree.ElementTree as ET
 
+
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
 
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -197,7 +197,7 @@ class MachineState(nixops.resources.ResourceState):
                         continue
                     res.failed_units.append(match.group(1))
 
-    def restore(self, defn, backup_id: str, devices: List[str]=[]):
+    def restore(self, defn, backup_id: Optional[str], devices: List[str]=[]):
         """Restore persistent disks to a given backup, if possible."""
         self.warn(
             "don't know how to restore disks from backup for machine ‘{0}’".format(
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -35,7 +35,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
             == "true"
         )
 
-        def _extract_key_options(x) -> Dict[str, Any]:
+        def _extract_key_options(x) -> Dict[str, str]:
             opts = {}
             for (key, xmlType) in (
                 ("text", "string"),
@@ -98,7 +98,7 @@ class MachineState(nixops.resources.ResourceState):
         self._ssh_private_key_file: Optional[str] = None
         self.new_toplevel: Optional[str] = None
 
-    def prefix_definition(self, attr) -> None:
+    def prefix_definition(self, attr):
         return attr
 
     @property
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -7,7 +7,7 @@ from typing import Dict, Any, List, Optional, Union, Set
 import nixops.util
 import nixops.resources
 import nixops.ssh_util
-
+import xml.etree.ElementTree as ET
 
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
@@ -35,7 +35,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
             == "true"
         )
 
-        def _extract_key_options(x) -> Dict[str, str]:
+        def _extract_key_options(x: ET.Element) -> Dict[str, str]:
             opts = {}
             for (key, xmlType) in (
                 ("text", "string"),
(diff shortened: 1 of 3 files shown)


nixops/util.py
```diff
@@ -270,7 +270,7 @@ def _maybe_abspath(s: str) -> str:
     return os.path.abspath(s)
 
 
-def abs_nix_path(x):
+def abs_nix_path(x) -> str:
```
grahamc:
```suggestion
def abs_nix_path(x: str) -> str:
```

nixops/resources/__init__.py
```diff
@@ -89,19 +89,19 @@ def _set_attrs(self, attrs):
...
+    def _del_attr(self, name) -> None:
         """Delete a machine attribute from the state file."""
         with self.depl._db:
             self.depl._db.execute(
                 "delete from ResourceAttrs where machine = ? and name = ?",
                 (self.id, name),
             )
 
-    def _get_attr(self, name, default=nixops.util.undefined):
+    def _get_attr(self, name, default=nixops.util.undefined) -> Any:
```
grahamc:
```suggestion
    def _get_attr(self, name: str, default=nixops.util.undefined) -> Any:
```

nixops/resources/__init__.py
```diff
@@ -89,19 +89,19 @@ def _set_attrs(self, attrs):
                         (self.id, n, v),
                     )
 
-    def _set_attr(self, name, value):
+    def _set_attr(self, name, value) -> None:
         """Update one machine attribute in the state file."""
         self._set_attrs({name: value})
 
-    def _del_attr(self, name):
+    def _del_attr(self, name) -> None:
```
grahamc:
```suggestion
    def _del_attr(self, name: str) -> None:
```

nixops/resources/__init__.py
```diff
@@ -89,19 +89,19 @@ def _set_attrs(self, attrs):
                         (self.id, n, v),
                     )
 
-    def _set_attr(self, name, value):
+    def _set_attr(self, name, value) -> None:
```
grahamc: I think this?

```suggestion
    def _set_attr(self, name: str, value: Any) -> None:
```

nixops/resources/__init__.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, depl, name: str, id):
         self.logger = depl.logger.get_logger_for(name)
         self.logger.register_index(self.index)
 
-    def _set_attrs(self, attrs):
+    def _set_attrs(self, attrs) -> None:
```
grahamc: maybe this? Feeling less sure about these Any values, but the str is for sure right.

```suggestion
    def _set_attrs(self, attrs: Dict[str, Any]) -> None:
```

nixops/backends/__init__.py
```diff
@@ -199,7 +197,7 @@ def _check(self, res):
                         continue
                     res.failed_units.append(match.group(1))
 
-    def restore(self, defn, backup_id, devices=[]):
+    def restore(self, defn, backup_id, devices: List=[]):
```
grahamc:
```suggestion
    def restore(self, defn, backup_id: str, devices: List[str]=[]):
```

nixops/backends/__init__.py
```diff
@@ -100,31 +98,31 @@ def __init__(self, depl, name: str, id: int):
         self._ssh_private_key_file: Optional[str] = None
         self.new_toplevel: Optional[str] = None
 
-    def prefix_definition(self, attr):
+    def prefix_definition(self, attr) -> None:
```
grahamc: Not sure about this None? :)

nixops/backends/__init__.py
```diff
@@ -37,7 +35,7 @@ def __init__(self, xml, config={}):
             == "true"
         )
 
-        def _extract_key_options(x):
+        def _extract_key_options(x) -> Dict[str, Any]:
```
grahamc: `x` is going to be an

```
 import xml.etree.ElementTree as ET


ET.Element
```

grahamc: and it looks like we have a guarantee this returns `Dict[str, str]`, no?

nixops/backends/__init__.py
```diff
@@ -12,14 +12,12 @@
...
-    def __init__(self, xml, config={}):
+    def __init__(self, xml, config={}) -> None:
         nixops.resources.ResourceDefinition.__init__(self, xml, config)
-        self.encrypted_links_to = set(
-            [
-                e.get("value")
-                for e in xml.findall("attrs/attr[@name='encryptedLinksTo']/list/string")
-            ]
-        )
+        self.encrypted_links_to = {
```
grahamc: Hmm not sure about changing this to a dict from a set?

aaronchall: Old way was list comprehension passed to set (which unnecessarily builds a list and then throws it away), edited way is set comprehension (syntax new in Python 2.7, see end of sets section in the official tutorial: https://docs.python.org/3/tutorial/datastructures.html#sets).

nixops/backends/__init__.py
grahamc: I pushed a fix implementing this part

nixops/backends/__init__.py
grahamc: I see this was removed, thanks.

grahamc:
Thank you a lot!

grahamc:
Do my commits look good to you?

aaronchall:
looks great! I like the set annotation!

grahamc:
By the way, this is how this PR did:

```
Imprecision went down:
nixops.backends:		50.29 -> 39.66
nixops.deployment:		18.94 -> 18.22
nixops.resources:		41.5 -> 39.53
nixops.util:		21.89 -> 18.0
Total:		30.96 -> 29.6
```

PR #1259, opened 2020-03-20
Update overview.xml

JosephLucas:
Two small suggestions.

1) I haven't seen any mention of `nixops ssh` before.
2) IIUC `argtest` is used afterwards to refer to the same thing


grahamc:
Thank you!

PR #1262, opened 2020-03-22
Use context manager for file opens

bhipple:
Aside from being safer and more idiomatic, it's also fewer lines of code.

No functional change, just cleanup.


grahamc:
Ratchet failed, but this change isn't making any type changes, nor do the lines involved really need types imo.

PR #1263, opened 2020-03-24
script_defs: open state files and deployments with a context manager

grahamc:
Use context managers to open and manage statefiles and deployments. This lets us do some fancier things with those state files. For example, storing the state other places.

This change is almost entirely whitespace. Check it out without whitespace in the diff: https://github.com/NixOS/nixops/pull/1263/files?w=1

Also consider looking commit-by-commit, which makes it easier, I think, to review.

Here is a diff adding state backends using this: legacy, in-memory, and s3: https://github.com/grahamc/nixops/compare/context-manager...grahamc:context-manager-states-rebase?expand=1

### DIFF
--- a/nixops/script_defs.py
+++ b/nixops/script_defs.py
@@ -607,13 +607,13 @@ def op_send_keys(args):
 
 
 def op_set_args(args):
-    depl = open_deployment(args)
-    for [n, v] in args.args or []:
-        depl.set_arg(n, v)
-    for [n, v] in args.argstrs or []:
-        depl.set_argstr(n, v)
-    for [n] in args.unset or []:
-        depl.unset_arg(n)
+    with deployment(args) as depl:
+        for [n, v] in args.args or []:
+            depl.set_arg(n, v)
+        for [n, v] in args.argstrs or []:
+            depl.set_argstr(n, v)
+        for [n] in args.unset or []:
+            depl.unset_arg(n)
 
 
 def op_destroy(args):
--- a/nixops/script_defs.py
+++ b/nixops/script_defs.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 
-from nixops import deployment
 from nixops.nix_expr import py2nix
 from nixops.parallel import MultipleExceptions, run_tasks
 import pluggy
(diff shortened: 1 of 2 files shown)


nixops/script_defs.py
```diff
@@ -782,157 +818,166 @@ def scp_loc(user, ssh_name, remote, loc):
...
 
 
 def op_list_generations(args):
-    depl = check_rollback_enabled(args)
-    if (
-        subprocess.call(["nix-env", "-p", depl.get_profile(), "--list-generations"])
-        != 0
-    ):
-        raise Exception("nix-env --list-generations failed")
+    with with_rollback_enabled(args) as depl:
```
adisbladis: Maybe this is nitpicking but I feel that `with with_` is a bit strange.

grahamc: Oh yeah, totally. I don't really like the names of these context managers. Maybe you have a different suggestion?

adisbladis: Maybe invert the naming and call it `deployment_with_rollback`? It also makes things more discoverable.

grahamc: Fixed!

nixops/script_defs.py
```diff
@@ -37,6 +37,22 @@
...
+        depl = open_deployment(sf, args)
+        yield depl
+
+
+@contextlib.contextmanager
+def network_state(args: Namespace) -> Generator[nixops.statefile.StateFile, None, None]:
+    state = nixops.statefile.StateFile(args.state_file)
+    try:
+        yield state
+    finally:
```
grahamc: This used to catch the exception and no longer tries to.

nixops/script_defs.py
```diff
@@ -782,157 +815,166 @@ def scp_loc(user, ssh_name, remote, loc):
...
+        env["NIX_SSHOPTS"] = " ".join(m.get_ssh_flags())
+        res = nixops.util.logged_exec(
+            [
+                "nix",
+                "copy",
+                "--to",
+                "ssh://{}".format(m.get_ssh_name()),
+                args.storepath,
+            ],
+            m.logger,
```
domenkozar: :+1:

PR #1264, opened 2020-03-24
Example NixOps State Backends

grahamc:
Store in S3:

```
{
  network = {
    storage.s3 = {
      profile = "rootlol";
      region = "us-east-1";
      bucket = "grahamc-nixops-state";
      key = "personal.nixops";
      kms_keyid = "166c5cbe-b827-4105-bdf4-a2db9b52efb4";
    };
  };

  flexo = {
    deployment = {
      targetHost = "147.75.105.137";
    };
    ...
  };
}
```

Store in Memory:

```
{
  network = {
    storage.memory = {};
  };

  flexo = {
    deployment = {
      targetHost = "147.75.105.137";
    };
    ...
  };
}
```

Legacy, exactly as master works now:
```
{
  network = {
    storage.legacy = {};
  };

  flexo = {
    deployment = {
      targetHost = "147.75.105.137";
    };
    ...
  };
}
```


nixops/storage/legacy.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, args: StorageArgValues) -> None:
+        pass
+
+    # fetchToFile: acquire a lock and download the state file to
+    # the local disk. Note: no arguments will be passed over kwargs.
+    # Making it part of the type definition allows adding new
+    # arguments later.
+    def fetchToFile(self, path: str, **kwargs) -> None:
+        os.symlink(os.path.abspath(self.state_location()), path)
```
grahamc: This at one point did not abspath, which was causing  problems with legacy paths to state files.

nixops/storage/s3.py
```diff
@@ -0,0 +1,82 @@
...
+            return {}
+
+    def lock(self, path) -> None:
+        r = self.aws.client("dynamodb").put_item(
+            TableName=self.dynamodb_table,
+            Item={"LockID": {"S": f"{self.bucket}/{self.key}"},},
+            ConditionExpression="attribute_not_exists(LockID)",
+        )
+
+    def unlock(self, path: str) -> None:
```
domenkozar: Would be cool to use contextmanager here as well to ensure a bit of safety

domenkozar: This way it can be moved into Storage abstraction and then backend can implement it or not

grahamc:
Right now if you don't define `network`, you get a bad error:

```
RuntimeError: error: attribute 'network' in selection path 'network' not found
```

grahamc:
If you define `network.storage = {}` you get a bad error:

```
    key = list(storage.keys()).pop()
IndexError: pop from empty list
```

nixops/storage/s3.py
adisbladis: I'm not sure we should completely conflate looking & storage. 
S3, Backblaze B2 & friends don't have any looking mechanisms themselves, so you may want to utilise a dynamodb table for locking.

It's fine if a storage backend implements a default locking mechanism though, as long as it can be overriden by the user.

ip1981:
A PR can be (and converted to) to a draft, so you wouldn't need a "do not merge" warning.

nixops/storage/legacy.py
kyrias: It seems to me like it would be a bit safer to make an actual copy rather than a symlink.

That way we know that if something goes wrong in the middle of writing the state file to disk, the "proper"  statefile won't be corrupted.

grahamc: If we were to copy, we would need to implement our own implementation of locking on top. The legacy backend already supports its own locking. My thinking is it would be better to find a way to "escape" the legacy backend, or implement a "backup" feature in the Legacy backend, than implement copying. What do you think?

grahamc:
todo:

- [ ] split the locking from the storage
- [ ] treat these as context managers
- [ ] remove the S3 one and relocate it to nixops-aws
- [ ] error messages / data validation is terrible, reimplement on top of #1275
- [ ] tests, somehow
- [ ] docs. lots, lol

this PR removes the idea of "global " networks and the "-d ..." argument, because of the dependence on `./network.nix`. So, some thinking of the implications of this are needed.

grahamc:
Using HEAD I see this oddity:

```
$ nixops check
The network requires the 'legacy' state provider, but no plugin provides it:

```
(Pdb) PluginManager.storage_backends()
{}

(Pdb) pprint([x for x in nixops.plugins.manager.get_plugins()])
[<nixopsvbox.plugin.NixopsVboxPlugin object at 0x7fd7a1c06690>]
```

**Comment by kisik21** [comment]
Can the s3 backend use minio?

**Comment by adisbladis** [comment]
I have just pushed the current state of this PR to https://github.com/NixOS/nixops/tree/context-manager-states.

I will be cleaning this up from things belonging in external plugins (like the s3 backend) and implement some interface changes to this PR.

If you'd like to revive any of this functionality or use it as a basis for a plugin state backend I'm going to leave this code around.

**Comment by deepfire** [comment]
So, this is the last component of https://github.com/NixOS/nixops/issues/1242?

This is really exciting!

**Review by lastthyme — APPROVED** [review]

**Comment by adisbladis** [comment]
We're never going to move forward with this without getting this change out to more people for review.

Therefore I'm merging this, with the caveat that it might be somewhat changed before a final release.

File: nixops/evaluation.py [hunk 3]
```diff
@@ -0,0 +1,204 @@
...
+            """
+WARNING: NixOps 1.0 -> 2.0 conversion step required
+
+NixOps 2.0 added support for multiple storage backends.
+
+Upgrade steps:
+1. Open %s
+2. Add:
+    network.storage.legacy = {
+      databasefile = "~/.nixops/deployments.nixops"
```
rehno-lindeque, line 156: There's a reference to `databasefile` here that doesn't seem to exist in the code

rehno-lindeque:
> - [X] remove the S3 one and relocate it to nixops-aws

Looks like the relocation part of this one still needs doing if I'm not mistaken? I couldn't find any reference to `StorageBackend` in nixops-aws.

roberth:
Whew, I finally got around to playing with this. Good call merging this though!

My observations so far:

The interface is quite coarse grained, dealing with the whole db only and having no knowledge of which deployment is used. This coarseness means that state file history becomes noisy and locking more problematic as the number of deployments increases. Users can choose distinct backend configurations to compensate for this, but that's effectively duplicating the `-d` feature that already exists. I'd be ok removing `-d` altogether, but that's perhaps too radical. Making the interface fine grained seems like the way to go.

State is always written, even for read-only operations like `nixops info`. I could compensate for this in a backend, but that's not the right solution. The same problem applies to locking, where it can't be worked around in the backend. So if NixOps needs to know what it's about to do anyway, it may as well use that knowledge in order not to write state redundantly.

Sqlite isn't really an interchange format. If something goes wrong, it's hard to read. I don't know about its format/version strategy (maybe it's good though?) and it's not easy to diff. Perhaps the backends could be based on `nixops import`/`export` instead?

Which brings us to: `nixops import --include-keys`-like functionality doesn't seem to be included. I haven't tested it and I may have read the code wrong, but it also seems like a good idea to have a single code path for this stuff anyway.

Mic92:
Is the memory backend functional at all? We tried to use both legacy backend and memory backend in https://github.com/nix-community/infra/pull/123 but they both seem not to work. nixops stable is also broken, which means there is no functional nixops left.

PR #1266, opened 2020-03-26
send keys in parallel

### LINKED ISSUE #671: Upload all deployment keys at once
### Motivation

When there are multiple `deployment.keys`, a separate `scp` upload is done for every single key file. This takes a lot of time and could probably be faster.

### Idea
Maybe the keys could be uploaded all at once -- either using recursive `scp -r` on the whole keys directory or by some other means.

domenkozar:
Fixes #671 

not great not terrible but vewy fast

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -275,7 +275,8 @@ class MachineState(nixops.resources.ResourceState):
             return
         if self.store_keys_on_machine:
             return
-
+        # preflight the connection
+        self.run_command("true")
         def worker(task: nixops.parallel.Task) -> None:
             k = task.key
             opts = task.opts
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -4,7 +4,9 @@ import os
 import re
 import subprocess
 from typing import Dict, Any, List, Optional, Union, Set
+import types
 import nixops.util
+import nixops.parallel
 import nixops.resources
 import nixops.ssh_util
 import xml.etree.ElementTree as ET
@@ -273,12 +275,18 @@ class MachineState(nixops.resources.ResourceState):
             return
         if self.store_keys_on_machine:
             return
-        for k, opts in self.get_keys().items():
+
+        def worker(task: nixops.parallel.Task) -> None:
+            k = task.key
+            opts = task.opts
             self.log("uploading key ‘{0}’...".format(k))
-            tmp = self.depl.tempdir + "/key-" + self.name
+
             if "destDir" not in opts:
                 raise Exception("Key '{}' has no 'destDir' specified.".format(k))
 
+            tmp = "{dir}/key-{name}-{key}".format(
+                dir=self.depl.tempdir, name=self.name, key=k,
+            )
             destDir = opts["destDir"].rstrip("/")
             self.run_command(
                 (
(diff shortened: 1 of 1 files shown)


nixops/backends/__init__.py
```diff
@@ -329,6 +337,20 @@ def send_keys(self) -> None:
             )
             self.run_command("mv " + tmp_outfile_esc + " " + outfile_esc)
             os.remove(tmp)
+
+        def task(key: str, opts: Dict[str, Any]) -> nixops.parallel.Task:
+            task = types.SimpleNamespace()
```
grahamc: is there any concern with having two things named `task` here?

domenkozar: fixed

nixops/backends/__init__.py
```diff
@@ -273,12 +275,18 @@ def send_keys(self) -> None:
             return
         if self.store_keys_on_machine:
             return
-        for k, opts in self.get_keys().items():
+
```
grahamc: When I tried this, all the threads tried to open the control connection at once:

```
flexo.......> uploading key ‘wireguard.nix’...
flexo.......> uploading key ‘wireguard1’...
flexo.......> uploading key ‘wireguard2’...
flexo.......> uploading key ‘wireguard3’...
flexo.......> uploading key ‘wireguard4’...
flexo.......> uploading key ‘wireguard5’...
flexo.......> uploading key ‘wireguard6’...
flexo.......> sign_and_send_pubkey: signing failed for RSA "/home/grahamc/.ssh/id_rsa" from agent: agent refused operation
flexo.......> mux_client_request_session: read from master failed: Broken pipe
flexo.......> mux_client_request_session: read from master failed: Broken pipe
flexo.......> mux_client_request_session: read from master failed: Broken pipe
flexo.......> mux_client_request_session: read from master failed: Broken pipe
flexo.......> uploading key ‘wireguard7’...
flexo.......> uploading key ‘wireguard8’...
flexo.......> uploading key ‘wireguard9’...
```

Preflighting the connection lets the control socket open, and then proceed:

```suggestion
        # preflight the connection
        self.run_command("true")
```

nixops/backends/__init__.py
```diff
@@ -273,12 +275,19 @@ def send_keys(self) -> None:
             return
         if self.store_keys_on_machine:
             return
-        for k, opts in self.get_keys().items():
+        # preflight the connection
+        self.run_command("true")
```
domenkozar:
```suggestion
        self.run_command("true")
        
```

nixops/backends/__init__.py
```diff
@@ -329,6 +339,20 @@ def send_keys(self) -> None:
...
+        def make_task(key: str, opts: Dict[str, Any]) -> nixops.parallel.Task:
+            task = types.SimpleNamespace()
+            task.name = self.name
+            task.key = key
+            task.opts = opts
+            return task
+
+        nixops.parallel.run_tasks(
+            # by default ssh allows 10 sessions per connection via MaxSessions setting
+            nr_workers=8,
```
grahamc: Some day, somebody will send a PR making the 8 configurable for some weird edge case. Today is not that day :P

domenkozar: :D

dhess:
This change appears to be effectively forkbombing my Mac. My network isn't very large, something like 10 machines, but I do have probably 5-10 keys per host. 

When I run a `nixops deploy`, it brings the Mac's networking stack to a standstill. :(

grahamc:
Nice. I'll revert for now, we can / should add a global limit on the number of keys being sent at a time.

domenkozar:
Huh, 10 connections with 100 sessions kills networking stack? What kind of Mac is this?

dhess:
iMac Pro, 64GiB, 10 cores, so no slouch.

I'm as surprised as you are. But there were so many SSH processes running, my `nixops deploy` hung and I couldn't even `ps aux|wc -l` to count them all from another terminal.

grahamc:
I've seen similar problems, actually. As far as I could tell it had to do with the thundering herd of connections. I wonder if we could pre-calculate how many keys need to be sent, and allocate a total max or something? I dunno, that gets complicated.

dhess:
Just tested `head`, that's much better. Thanks for the remarkably fast turnaround!

Another possibility is this: I'm using `gpg-agent` to manage my SSH keys, and my SSH private key is stored on a YubiKey. I was under the impression that NixOps always uses the SSH key it generates for `root@host` at `nixops create` time for each `host`, but every once in awhile, NixOps will prompt me for my card's PIN anyway, for reasons I don't understand. So I wonder if some of the SSH sessions are hitting the YubiKey and it, or `gpg-agent`, just can't keep up?

dhess:
After speaking with @grahamc on IRC about this, it sounds like which SSH key NixOps uses is not necessarily deterministic, and that it may use a personal SSH key? If so, anyone who's using a hardware device to manage their SSH private keys is going to have problems with this implementation, assuming I'm right about what was going on here.

PR #1268, opened 2020-03-28
util: Don't check if process.stdin is None unless stdin_string is passed

adisbladis:
When calling subprocess.Popen passing anything except subprocess.PIPE as stdin the results stdin attribute is None.

If a file descriptor was passed as stdin we don't want to run this check.

### DIFF
--- a/nixops/util.py
+++ b/nixops/util.py
@@ -119,10 +119,6 @@ def logged_exec(
             fds = [process.stdout]
         log_fd_opt = process.stdout
 
-    if process.stdin is None:
-        raise ValueError("process.stdin was None")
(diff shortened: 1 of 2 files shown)


grahamc:
I wonder how tricky it would be to write a test for this, what do you think?

adisbladis:
I've added a small test :)

adisbladis:
This test https://github.com/NixOS/nixops/pull/1268/files#diff-f622ee648d48b94128d5bbef35efceceR23 would have caught the issue.

grahamc:
test coverage :chart_with_upwards_trend: 
type coverage :chart_with_upwards_trend:

PR #1269, opened 2020-03-28
Revert "send keys in parallel"

grahamc:
Reverts NixOS/nixops#1266

We effectively forkbomb'd @dhess, and we should probably add a global limit to the number of sends at a time.


grahamc:
Not surprised the ratchet failed, every revert will cause the ratchet to fail :)

Issue #730, opened 2017-09-23
Allow nixops to use another user than root for SSHing


lejonet:
It would seem like nixops is hardcoded to use root as the user account it uses for SSH connections to the machines.
Regardless if deployment.targetHost references a ssh-config host or is in the form of "example@1.2.3.4" nixops prepends "root@" to it and tries to connect.

It would be nice to have the option of specifying both what user it should connect as, and what mechanism it should use to gain the elevated privileges, when it needs them, on the target machine (sudo, su, doas, future implementations).

lejonet:
The code in use by the SSH and SSHMaster classes seems to already support supplying an user other than root, both the run_command()and _get_target() methods have an argument to supply an user other than root but it defaults to User=None which triggers _get_target() to supply root as user. Even the __init__ method of SSHMaster has an user argument.

However, this isn't used at all in the 3 invocations of run_command() I can find in nixops/deployments.py (line 672, 682 and 927) but the SSH class supplies the argument when instantiating SSHMaster in self._ssh_master at line 195 in nixops/ssh_util.py.

So all-in-all it seems like the "only" thing that needs to be done is make it possible to set a config value for the user, and make it trickle down to the SSH/SSHMaster objects and the same with the elevation mechanism, tho there some code to amend the command to be executed when it is set needs to be coded.

jbgi:
As Azure no more allows to provisionning of vm with `root` as admin user, this a blocking point for us to try out nixops on azure.

teto:
I can't login directly as root on some servers here for security reasons so I concur on this.

mvgijssel:
+1 on this, would love not to expose the root user over SSH

devhell:
+1 from me too. Not only that, but the [documentation](https://nixos.org/nixops/manual/#sec-deploying-to-physical-nixos) seems to be a bit misleading, as it gives the impression that this is the most straight-forward way of managing a fleet of already existing NixOS machines, which is clearly not the case if one has not enabled root SSH access on those machines (e.g., because one has specified a different username for maintenance tasks).

adisbladis:
See https://github.com/NixOS/nixops/pull/1270. Ideally the privilege escalation mechanism should be pluggable, but I'm OK with just hard-coding sudo for now.

PR #1270, opened 2020-03-29
Add support for non-root deployments

adisbladis:
This adds a new `deployment` configuration attribute (`targetUser`).
To inherit the username from the local user issuing the deployment
set:
``` nix
deployment.targetUser = null;
```
Setting this to a string will deploy as that user. This option
defaults to "root".

 ### We assume the following for non-root deploys:

- Passwordless sudo (no TTY allocation possible)

I'm using the following NixOS configuration
``` nix
security.pam.services.sudo.sshAgentAuth = true;
security.pam.enableSSHAgentAuth = true;
```

For this use-case I've introduced:
``` nix
deployment.sshOptions = [ "-A" ];
```

- The deployment user is trusted by the Nix daemon
``` nix
nix.trustedUsers = [ "adisbladis" ];
```
This is required because of nix-copy-closure.

Closes https://github.com/NixOS/nixops/issues/730

### DIFF
--- a/doc/guides/deploy-without-root.rst
+++ b/doc/guides/deploy-without-root.rst
@@ -1,0 +1,95 @@
+Deploying without Root
+====
+
+Requirements
+----
+
+NixOps 2.0 allows for deploying as users other than root, as long
+as the deploying user meets two requirements:
+
+1. The user can become root without needing to type a password.
+2. Nix considers the user to be a `"trusted user" <https://nixos.org/nix/manual/#conf-trusted-users>`_.
+
+In this guide, we will use passwordless sudo.
+
+We assume:
+
+1. The deploying user's name is "deployer".
+2. The target machine's name is "hermes".
+3. The target machine is already managed by NixOps.
+
+Steps
+----
+
+1. Configure the target machine according to the listed requirements.
+2. Update the NixOps network to use our alternative user.
+3. Deploy as the new user.
+
+
+Configuring the Target Machine
+****
+
+First, mark the deploying user as trusted:
+
+.. code-block:: nix
+
+  {
+    nix.trustedUsers = [ "deployer" ];
+  }
+
+This will let the user copy Nix store paths to the target.
+
+Let the deploying user use sudo:
+
+.. code-block:: nix
+
+  {
+    users.users.deployer.extraGroups = [ "wheel" ];
+  }
+
+Then, we configure the machine to have passwordless sudo:
+
+.. code-block:: nix
+
+  {
+    security.sudo.wheelNeedsPassword = false;
+  }
+
+Now use NixOps to deploy these changes to the server before taking
+the next step.
+
+Configuring the NixOps Network
+****
+
+Edit your network.nix to specify the machine's
+``deployment.targetUser``:
+
+.. code-block:: nix
+
+  {
+    network.description = "Non-root deployment";
+
+    hermes =
+      { resources, ... }:
+      {
+        deployment.targetUser = "deployer";
+      };
+  }
+
+
+Testing our Changes
+****
+
+Then, run ``nixops deploy`` to update the NixOps database. This deploy
+will use your "deployer" user instead of root.
+
+Try running ``nixops ssh``, and see that you are logged in as
+"deployer".
+
+Notes
+----
+
+* NixOps caches the target user and related variables in its state
+  file, and commands like ``nixops send-keys`` and ``ssh`` use the
+  cached data. After changing these values, run ``nixops deploy`` to
+  update the cache.
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -118,7 +118,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
       foldr (a: b: a // b)
-        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser sshOptions alwaysActivate owners keys hasFastConnection;
+        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser sshOptions privilegeEscalationCommand alwaysActivate owners keys hasFastConnection;
           nixosRelease = v.config.system.nixos.release or v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
           publicIPv4 = v.config.networking.publicIPv4;
         }
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -118,7 +118,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
       foldr (a: b: a // b)
-        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser alwaysActivate owners keys hasFastConnection;
+        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser sshOptions alwaysActivate owners keys hasFastConnection;
           nixosRelease = v.config.system.nixos.release or v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
           publicIPv4 = v.config.networking.publicIPv4;
         }
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -118,7 +118,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
       foldr (a: b: a // b)
-        { inherit (v.config.deployment) targetEnv targetPort targetHost alwaysActivate owners keys hasFastConnection;
+        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser alwaysActivate owners keys hasFastConnection;
           nixosRelease = v.config.system.nixos.release or v.config.system.nixosRelease or (removeSuffix v.config.system.nixosVersionSuffix v.config.system.nixosVersion);
           publicIPv4 = v.config.networking.publicIPv4;
         }
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -64,6 +64,17 @@ in
       '';
     };
 
+    deployment.privilegeEscalationCommand = mkOption {
+      type = types.listOf types.str;
+      default = [ "sudo" "-H" ];
+      description = ''
+        A command to escalate to root privileges when using SSH as a non-root user.
+        This option is ignored if the <literal>targetUser</literal> option is set to <literal>root</literal>.
+
+        The program and its options are executed verbatim without shell.
+      '';
+    };
+
     deployment.alwaysActivate = mkOption {
       type = types.bool;
       default = true;
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -34,9 +34,12 @@ in
       type = types.nullOr types.str;
       default = "root";
       description = ''
-        This option specifies the username to be used by
-        NixOps on the remote system to execute deployment
-        operations.
+        The username to be used by NixOps by SSH when connecting to the
+        remote system.
+
+        If <literal>targetUser<literal> is set to <literal>null<literal>
+        the username is set to the username of the user invoking
+        <literal>nixops<literal>.
       '';
     };
 
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -69,12 +69,15 @@ in
 
     deployment.privilegeEscalationCommand = mkOption {
       type = types.listOf types.str;
-      default = [ "sudo" "-H" ];
+      default = [ "sudo" "-H" "--" ];
       description = ''
         A command to escalate to root privileges when using SSH as a non-root user.
         This option is ignored if the <literal>targetUser</literal> option is set to <literal>root</literal>.
 
         The program and its options are executed verbatim without shell.
+
+        It's good practice to end with "--" to indicate that the privilege escalation command
+        should stop processing command line arguments.
       '';
     };
 
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -30,6 +30,16 @@ in
       '';
     };
 
+    deployment.targetUser = mkOption {
+      type = types.nullOr types.str;
+      default = "root";
+      description = ''
+        This option specifies the username to be used by
+        NixOps on the remote system to execute deployment
+        operations.
+      '';
+    };
+
     deployment.targetHost = mkOption {
       type = types.str;
       description = ''
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -31,16 +31,16 @@ in
     };
 
     deployment.targetUser = mkOption {
-      type = types.nullOr types.str;
+      # type = types.nullOr types.str;
+      type = types.str;
       default = "root";
       description = ''
         The username to be used by NixOps by SSH when connecting to the
         remote system.
-
-        If <literal>targetUser<literal> is set to <literal>null<literal>
-        the username is set to the username of the user invoking
-        <literal>nixops<literal>.
       '';
+      # If <literal>targetUser</literal> is set to <literal>null</literal>
+      # the username is set to the username of the user invoking
+      # </literal>nixops</literal>.
     };
 
     deployment.targetHost = mkOption {
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -56,6 +56,14 @@ in
       '';
     };
 
+    deployment.sshOptions = mkOption {
+      type = types.listOf types.str;
+      default = [];
+      description = ''
+        Extra options passed to the OpenSSH client verbatim, and are not executed by a shell.
+      '';
+    };
+
     deployment.alwaysActivate = mkOption {
       type = types.bool;
       default = true;
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -54,10 +54,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
         self.keys = {k: KeyOptions(**v) for k, v in config["keys"].items()}
         self.ssh_options = config["sshOptions"]
 
-        self.ssh_user = getpass.getuser()
-        ssh_user = config["targetUser"]
-        if ssh_user is not None:
-            self.ssh_user = ssh_user
+        self.ssh_user = config["targetUser"]
 
         self.privilege_escalation_command = config["privilegeEscalationCommand"]
 
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -43,6 +43,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
     keys: Mapping[str, KeyOptions]
     ssh_user: str
     ssh_options: List[str]
+    privilege_escalation_command: List[str]
 
     def __init__(self, name: str, config: nixops.resources.ResourceEval):
         super().__init__(name, config)
@@ -57,6 +58,8 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
         ssh_user = config["targetUser"]
         if ssh_user is not None and ssh_user:
             self.ssh_user = ssh_user
+
+        self.privilege_escalation_command = config["privilegeEscalationCommand"]
 
 
 class MachineState(nixops.resources.ResourceState):
@@ -116,6 +119,8 @@ class MachineState(nixops.resources.ResourceState):
         self.has_fast_connection = defn.has_fast_connection
         if not self.has_fast_connection:
             self.ssh.enable_compression()
+
+        self.ssh.privilege_escalation_command = list(defn.privilege_escalation_command)
 
     def stop(self) -> None:
         """Stop this machine, if possible."""
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -56,7 +56,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
 
         self.ssh_user = getpass.getuser()
         ssh_user = config["targetUser"]
-        if ssh_user is not None and ssh_user:
+        if ssh_user is not None:
             self.ssh_user = ssh_user
 
         self.privilege_escalation_command = config["privilegeEscalationCommand"]
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -42,6 +42,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
     has_fast_connection: bool
     keys: Mapping[str, KeyOptions]
     ssh_user: str
+    ssh_options: List[str]
 
     def __init__(self, name: str, config: nixops.resources.ResourceEval):
         super().__init__(name, config)
@@ -50,6 +51,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
         self.owners = config["owners"]
         self.has_fast_connection = config["hasFastConnection"]
         self.keys = {k: KeyOptions(**v) for k, v in config["keys"].items()}
+        self.ssh_options = config["sshOptions"]
 
         self.ssh_user = getpass.getuser()
         ssh_user = config["targetUser"]
@@ -67,6 +69,7 @@ class MachineState(nixops.resources.ResourceState):
     ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
     ssh_port: int = nixops.util.attr_property("targetPort", 22, int)
     ssh_user: str = nixops.util.attr_property("targetUser", "root", str)
+    ssh_options: List[str] = nixops.util.attr_property("sshOptions", [], "json")
     public_vpn_key: Optional[str] = nixops.util.attr_property("publicVpnKey", None)
     keys: Mapping[str, str] = nixops.util.attr_property("keys", {}, "json")
     owners: List[str] = nixops.util.attr_property("owners", [], "json")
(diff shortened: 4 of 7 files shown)


nixops/ssh_util.py
```diff
@@ -169,6 +183,9 @@ def register_host_fun(self, host_fun: Callable[[], str]) -> None:
         self._host_fun = host_fun
 
     def _get_target(self, user: Optional[str] = None) -> str:
+        if user is None:
```
adisbladis: Note: This assertion is a temporary addition. Ideally `user` should not be `Optional`.

nixops/backends/__init__.py
```diff
@@ -455,15 +455,15 @@ def get_scp_name(self):
         return ssh_name
 
     def upload_file(self, source, target, recursive=False):
-        master = self.ssh.get_master()
+        master = self.ssh.get_master(user=self.ssh_user)
```
grahamc: Let's refactor the `get_master` stuff so call-sites don't need to remember to pass `user=...`. From what we were discussing, it sounds like `get_master` should be a private function anyway, internal to the SSH util.

nixops/backends/__init__.py
```diff
@@ -439,7 +449,8 @@ def copy_closure_to(self, path):
         # Any remaining paths are copied from the local machine.
         env = dict(os.environ)
         env["NIX_SSHOPTS"] = " ".join(
-            ssh._get_flags() + ssh.get_master(user=self.ssh_user).opts
+            ssh._get_flags()
+            + ssh.get_master(user=self.ssh_user, forward_agent=False).opts
```
grahamc, line 461: I'm a bit curious about the passing of `forward_agent` too to get_master, etc. We should be careful that plugins have an easy interface to do the thing they want, instead of needing to focus on the details of nixops internals I think?

nix/options.nix
```diff
@@ -55,6 +55,14 @@ in
       '';
     };
 
+    deployment.forwardSSHAgent = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        This option whether to forward the local SSH agent socket to the remote.
```
grahamc, line 62: Let's explain why this is maybe a bad idea.

nixops/ssh_util.py
```diff
@@ -285,37 +298,44 @@ def split_openssh_args(self, args: Iterable[str]) -> Tuple[List[str], Command]:
...
-                " ".join(
-                    ["'{0}'".format(arg.replace("'", r"'\''")) for arg in command]
-                ),
-            ]
+            cmd.append(
+                " ".join(["'{0}'".format(arg.replace("'", r"'\''")) for arg in command])
+            )
+
+        if user and user != "root":
+            cmd.insert(0, "sudo")
```
grahamc, line 326:
```suggestion
            cmd.insert(0, "sudo")
```

We may want to pass some more options:

```
  -n, --non-interactive         non-interactive mode, no prompts are used
  -i, --login                   run login shell as the target user; a command may also be specified
  -H, --set-home                set HOME variable to target user's home dir
```

though I'm not sure. I'm sort of thinking this implies we should make this configurable right out of the gate?

grahamc:
I'm thinking we should make two things more configurable, and less "baked-in":

1. make the escalation command (currently always `sudo`) configurable as a list of arguments to put before the command being run
2. make it possible to specify additional options to SSH, so we don't need to explicitly add support for forwarded agents, or other things like GSSAPI.

nixops/ssh_util.py
adisbladis: `-n` breaks my use case (SSH agent forwarding + PAM) despite not actually requiring a prompt, so I'd rather not set that as a default
`-i` is a bad idea because it's not the default when using OpenSSH & specifying a command, so it wouldn't be equivalent to a SSH session as root.
`-H` is a good idea. It's already default on my systems but depends on the remote host configuration so it's a good default.

adisbladis:
> make the escalation command (currently always sudo) configurable as a list of arguments to put before the command being run

Done! The option is in `deployment.privilegeEscalationCommand` (which defaults to using `sudo -H`).

> make it possible to specify additional options to SSH, so we don't need to explicitly add support for forwarded agents, or other things like GSSAPI.

This is now called `deployment.sshOptions`.

nixops/backends/__init__.py
adisbladis: I would like to do that in a follow-up PR. This PR is already big enough and it might require some significant refactoring.

nix/options.nix
adisbladis: This is now removed and `sshOptions` should be used instead. Now it's up to the user to know why a certain option is a bad idea.

nixops/backends/__init__.py
adisbladis: Let's sort that out in a future refactor of the `SSH`/`SSHMaster` classes.

nix/options.nix
```diff
@@ -55,6 +55,14 @@ in
       '';
     };
 
+    deployment.sshOptions = mkOption {
+      type = types.listOf types.str;
+      default = [];
+      description = ''
+        This option specifies extra options passed to OpenSSH.
```
grahamc, line 62:
```suggestion
        This option specifies extra options passed to the OpenSSH client.
```

grahamc: Is this list interpreted by a shell, or passed via `argv`? Let's clarify that in the option description.

grahamc:
```suggestion
        Extra options passed to the OpenSSH client verbatim, and are not executed by a shell.
```

nix/options.nix
```diff
@@ -63,6 +63,15 @@ in
       '';
     };
 
+    deployment.privilegeEscalationCommand = mkOption {
+      type = types.listOf types.str;
+      default = [ "sudo" "-H" ];
+      description = ''
+        This option specifies the command to escalate to root privileges
+        when using SSH as a non-root user.
```
grahamc, line 71:
```suggestion
        A command to escalate to root privileges when using SSH as a non-root user.
        This option is ignored if the <literal>targetUser</literal> option is set to <literal>root</literal>
```

nixops/backends/__init__.py
```diff
@@ -137,6 +145,8 @@ def set_common_state(self, defn) -> None:
         if not self.has_fast_connection:
             self.ssh.enable_compression()
 
+        self.ssh.privilege_escalation_command_set(defn.privilege_escalation_command)
```
grahamc, line 148: Should we put `set_` first? What is more pythonic?

adisbladis: I don't know what's more pythonic. Personally I tend to like suffixing the operation so editor completion will suggest related operations.

nix/eval-machine-info.nix
```diff
@@ -118,7 +118,7 @@ rec {
     machines =
       flip mapAttrs nodes (n: v': let v = scrubOptionValue v'; in
       foldr (a: b: a // b)
-        { inherit (v.config.deployment) targetEnv targetPort targetHost storeKeysOnMachine alwaysActivate owners keys hasFastConnection;
+        { inherit (v.config.deployment) targetEnv targetPort targetHost targetUser sshOptions storeKeysOnMachine alwaysActivate owners keys hasFastConnection privilegeEscalationCommand;
```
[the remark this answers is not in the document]
adisbladis: I want to put every single `inherit` on it's own line to reduce future merge conflicts around this.
Not right now though as I have other PRs touching the exact same line and that would make those harder to rebase than neccessary.

nixops/backends/__init__.py
adisbladis: The most Pythonic way would actually be _not_ to use setters/getters at all but just mutate the attribute.

adisbladis: Let's leave it like this for now and reconsider in the future `SSH`/`SSHMaster` refactor you and me discussed.

grahamc:
I gave this a go on a regular ol' server and it worked well. I then tried it on `kif` which sends some keys:

```
$ nixops deploy --deployment personal --include kif
kif.........> copying closure...
personal> closures copied successfully
kif.........> uploading key ‘vault-login’...
kif.........> scp: /run/keys/.vault-login.tmp: Permission denied
kif.........> error: Traceback (most recent call last):
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/deployment.py", line 920, in worker
    m.send_keys()
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/backends/__init__.py", line 330, in send_keys
    self.upload_file(tmp, tmp_outfile)
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/backends/__init__.py", line 478, in upload_file
    return self._logged_exec(cmdline)
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/backends/__init__.py", line 413, in _logged_exec
    return nixops.util.logged_exec(command, self.logger, **kwargs)
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/util.py", line 207, in logged_exec
    raise CommandFailed(err, res)
nixops.util.CommandFailed: command ‘['scp', '-P', '22', '-o', 'StrictHostKeyChecking=accept-new', '-i', '/run/user/1000/nixops-tmp04t91o0s/id_nixops-kif', '-oControlPath=/run/user/1000/nixops-ssh-tmp7jl7pc7v/master-socket', '/run/user/1000/nixops-tmp04t91o0s/key-kif', 'grahamc@10.5.3.16:/run/keys/.vault-login.tmp']’ failed on machine ‘kif’ (exit code 1)
```

I think we're going to need to go back to the drawing board a bit and plan this a bit more. In particular:

* callers of `run_command` have the assumption they're root
* `upload_file` does too

Leaving me with a couple questions:
* Does this feature make uploading riskier by uploading files as the user and then moving them in to place via sudo?
* What is the API guarantee we want to retain around `run_command`?

I'm sorry to drag this PR out.

adisbladis:
Marking this as Draft while https://github.com/NixOS/nixops/pull/1323 is still in-flight.

nix/options.nix
```diff
@@ -30,6 +30,16 @@ in
       '';
     };
 
+    deployment.targetUser = mkOption {
+      type = types.nullOr types.str;
+      default = "root";
+      description = ''
+        This option specifies the username to be used by
+        NixOps on the remote system to execute deployment
+        operations.
```
grahamc, line 39: "This option specifies..." is redundant in context. Also, let's specify a bit more of the details.

```suggestion
        The username to be used by NixOps by SSH when connecting to the
        remote system.
        
        If this user is not <literal>root</literal>, it must be able to become root using the command in deployment.privilegeEscalationCommand .
        When null, NixOps uses the current user's name. 
```

nixops/backends/__init__.py
```diff
@@ -45,6 +52,14 @@ def __init__(self, name: str, config: nixops.resources.ResourceEval):
         self.owners = config["owners"]
         self.has_fast_connection = config["hasFastConnection"]
         self.keys = {k: KeyOptions(**v) for k, v in config["keys"].items()}
+        self.ssh_options = config["sshOptions"]
+
+        self.ssh_user = getpass.getuser()
+        ssh_user = config["targetUser"]
+        if ssh_user is not None and ssh_user:
```
grahamc: I don't understand this, do you mean to check `self.ssh_user`? or ... not sure.

nixops/backends/__init__.py
```diff
@@ -56,6 +71,8 @@ class MachineState(nixops.resources.ResourceState):
     )
     ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
     ssh_port: int = nixops.util.attr_property("targetPort", 22, int)
+    ssh_user: str = nixops.util.attr_property("targetUser", "root", str)
+    ssh_options: List[str] = nixops.util.attr_property("sshOptions", [], "json")
     public_vpn_key: Optional[str] = nixops.util.attr_property("publicVpnKey", None)
```
grahamc, line 76: We may not want to store ssh_user in the state file, especially since it can be dynamically discovered from the current user.

nixops/backends/__init__.py
```diff
@@ -322,11 +343,11 @@ def get_keys(self):
     def get_ssh_name(self):
         assert False
 
-    def get_ssh_flags(self, scp=False):
+    def get_ssh_flags(self, scp=False) -> List[str]:
```
grahamc, line 346: Nice.

grahamc: I looked at making this _get_ssh_flags, but other plugins (aws at least) defines the function. Maybe it shouldn't, though.

nixops/backends/__init__.py
grahamc: Actually, send_keys and other commands don't evaluate, but do need to be able to SSH successfully. This might be marking the death of eval-less send-keys?

nixops/backends/__init__.py
grahamc: re #1307

nixops/backends/__init__.py
```diff
@@ -427,10 +452,21 @@ def _get_scp_name(self) -> str:
...
 
         ssh_cmdline: List[str] = ["ssh"] + self.get_ssh_flags() + master.opts
         cmdline = ["rsync", "-e", nixops.util.shlex_join(ssh_cmdline)]
+
+        if self.ssh_user != "root":
+            cmdline.extend(
+                [
+                    "--rsync-path",
+                    nixops.util.shlex_join(
+                        self.ssh.privilege_escalation_command + ["rsync"]
```
grahamc, line 465: I wonder if we should make  a note in the docs for priv esc cmd about ending their command with `--`? We can't universally add one here, but they should probably take care to tell their tool to "okay, stop parsing  your args now".

nixops/script_defs.py
```diff
@@ -777,15 +777,14 @@ def op_ssh(args):
         if not m:
             raise Exception("unknown machine ‘{0}’".format(machine))
         flags, command = m.ssh.split_openssh_args(args.args)
-        user = None if username == "root" else username
```
grahamc, line 780: I don't quite follow what is going on here

nixops/ssh_util.py
```diff
@@ -168,7 +169,7 @@ def register_host_fun(self, host_fun: Callable[[], str]) -> None:
         """
         self._host_fun = host_fun
 
-    def _get_target(self, user: Optional[str] = None) -> str:
+    def _get_target(self, user: str) -> str:
         if self._host_fun is None:
             raise AssertionError("don't know which SSH host to connect to")
         return "{0}@{1}".format("root" if user is None else user, self._host_fun())
```
grahamc, line 175:
```suggestion
        return "{0}@{1}".format(user, self._host_fun())
```

nixops/script_defs.py
adisbladis: This is argument handling for `nixops ssh ...`

nix/options.nix
```diff
@@ -30,6 +30,20 @@ in
       '';
     };
 
+    deployment.targetUser = mkOption {
+      type = types.nullOr types.str;
```
grahamc, line 34:
```suggestion
      type = types.str;
```

nixops/backends/__init__.py
```diff
@@ -45,6 +52,14 @@ def __init__(self, name: str, config: nixops.resources.ResourceEval):
         self.owners = config["owners"]
         self.has_fast_connection = config["hasFastConnection"]
         self.keys = {k: KeyOptions(**v) for k, v in config["keys"].items()}
+        self.ssh_options = config["sshOptions"]
+
+        self.ssh_user = getpass.getuser()
+        ssh_user = config["targetUser"]
+        if ssh_user is not None:
+            self.ssh_user = ssh_user
```
grahamc, line 60: make this not support None user, and a specified user in the expr

nixops/backends/__init__.py
```diff
@@ -56,6 +71,8 @@ class MachineState(nixops.resources.ResourceState):
     )
     ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
     ssh_port: int = nixops.util.attr_property("targetPort", 22, int)
+    ssh_user: str = nixops.util.attr_property("targetUser", "root", str)
+    ssh_options: List[str] = nixops.util.attr_property("sshOptions", [], "json")
     public_vpn_key: Optional[str] = nixops.util.attr_property("publicVpnKey", None)
     keys: Mapping[str, str] = nixops.util.attr_property("keys", {}, "json")
     owners: List[str] = nixops.util.attr_property("owners", [], "json")
```
grahamc, line 78: we would store the ssh_user in the statefile, and it would always be the same since it is a hard-coded string in the expr

grahamc:
I tested this PR and it works pretty well: `nixops deploy` works nicely as long as your user is `trusted` by Nix, and that your user is able to escalate privileges.

An open question is `nixops ssh` logs you in as root unless you specify the user with which you're logging in as. I expected to be logged in as `targetUser`, but then have become `root` via the priv escalation command automatically -- but this isn't what happened. I think we should ask some users about what they expect the behavior to be here. (Interestingly, TIL that you can do `nixops ssh username@machine`)

grahamc:
I tried this using commands that didn't exist, commands that were broken, an empty escalation command, a user which didn't exist, among other things and the errors were all fairly well handled: saying the command didn't exist.

As an untrusted user, it fails in a not great way -- but I think the problem is the Nix error message isn't good:

```
kif.........> copying path '/nix/store/kimzp0ncpmqza54shqjgczwrm24sqbvj-nix.conf' to 'ssh://grahamc@10.5.3.16'...
kif.........> error: cannot add path '/nix/store/kimzp0ncpmqza54shqjgczwrm24sqbvj-nix.conf' because it lacks a valid signature
kif.........> error: unexpected end-of-file
```

It would be good to write up a bit of documentation, like a little page, about deploying as not-root and the requirements and debugging problems like this one.

chreekat:
> An open question is `nixops ssh` logs you in as root unless you specify the user with which you're logging in as. I expected to be logged in as `targetUser`, but then have become `root` via the priv escalation command automatically -- but this isn't what happened. I think we should ask some users about what they expect the behavior to be here. (Interestingly, TIL that you can do `nixops ssh username@machine`)

I would also expect to be logged in as `targetUser` by default.

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,92 @@
...
+
+NixOps 2.0 allows for deploying as users other than root, as long
+as the deploying user meets two requirements:
+
+1. The user can become root without needing to type a password.
+2. Nix considers the user to be a `"trusted user" <https://nixos.org/nix/manual/#conf-trusted-users>`_.
+
+In this guide, we will use passwordless sudo.
+
+This guide assumes:
```
cole-h, line 15: "In this guide... This guide assumes" is repetitive. Maybe something like:

```suggestion

Our assumptions include:
```

or

```suggestion

We assume:
```

(newline because all the other lists do the same)

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,92 @@
...
+1. The deploying user's name is "deployer"
+2. The target machine's name is "hermes"
+3. the target machine is already managed by NixOps
```
cole-h, line 18:
```suggestion
3. The target machine is already managed by NixOps
```

cole-h, line 18:
```suggestion
1. The deploying user's name is "deployer".
2. The target machine's name is "hermes".
3. the target machine is already managed by NixOps.
```

All the other lists end with a period.

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,92 @@
...
+Edit your network.nix to specify the machine's
+``deployment.targetUser``:
+
+.. code-block:: nix
+
+  {
+    network.description = "Non-root deployment";
+
+    hermes =
+      { resources, ... }:
```
cole-h, line 69: Is `resources` necessary here? Or is it because of the contents prior to trimming it?

grahamc: I think it is useful to provide context in to where this goes, like to give a contextual anchor.

cole-h: Sure, makes sense.

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+***
+
+Then, run ``nixops deploy`` to update the NixOps database. This deploy
+will use your "deployer" user instead of root.
+
+Try running ``nixops ssh``, and see that you are logged in as
+deployer.
+
+Notes
+---
```
cole-h, line 88:
```suggestion
----
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+      { resources, ... }:
+      {
+        deployment.targetUser = "deployer";
+      };
+    }
+  }
+
+
+Testing our Changes
+***
```
cole-h, line 79:
```suggestion
****
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+.. code-block:: nix
+  {
+    security.sudo.wheelNeedsPassword = false;
+  }
+
+Now use NixOps to deploy these changes to the server before taking
+the next step.
+
+Configuring the NixOps Network
+***
```
cole-h, line 59:
```suggestion
****
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+Steps
+---
+
+1. Configure the target machine according to the listed requirements.
+2. Update the NixOps network to use our alternative user.
+3. Deploy as the new user.
+
+
+Configuring the Target Machine
+***
```
cole-h, line 30:
```suggestion
****
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+In this guide, we will use passwordless sudo.
+
+
+We assume:
+1. The deploying user's name is "deployer".
+2. The target machine's name is "hermes".
+3. The target machine is already managed by NixOps.
+
+Steps
+---
```
cole-h, line 22:
```suggestion
----
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
+Deploying without Root
+===
+
+Requirements
+---
```
cole-h, line 5:
```suggestion
----
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
+Deploying without Root
+===
```
cole-h, line 2:
```suggestion
====
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Try running ``nixops ssh``, and see that you are logged in as
+deployer.
+
+Notes
+---
+
+* NixOps caches the target user and related variables in its state
+  file, and commansd like ``nixops send-keys`` and ``ssh`` use the
+  cached data. After changing thes evalues, run ``nixops deploy`` to
```
cole-h, line 92:
```suggestion
  file, and commands like ``nixops send-keys`` and ``ssh`` use the
  cached data. After changing these values, run ``nixops deploy`` to
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+
+
+Testing our Changes
+***
+
+Then, run ``nixops deploy`` to update the NixOps database. This deploy
+will use your "deployer" user instead of root.
+
+Try running ``nixops ssh``, and see that you are logged in as
+deployer.
```
cole-h, line 85:
```suggestion
"deployer".
```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+NixOps 2.0 allows for deploying as users other than root, as long
+as the deploying user meets two requirements:
+
+1. The user can become root without needing to type a password.
+2. Nix considers the user to be a `"trusted user" <https://nixos.org/nix/manual/#conf-trusted-users>`_.
+
+In this guide, we will use passwordless sudo.
+
+
+We assume:
```
cole-h, line 16:
```suggestion
We assume:

```

Whoops, I did this backwards :D

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+2. Update the NixOps network to use our alternative user.
+3. Deploy as the new user.
+
+
+Configuring the Target Machine
+****
+
+First, mark the deploying user as trusted:
+
+.. code-block:: nix
```
cole-h, line 34:
```suggestion
.. code-block:: nix

```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+  {
+    nix.trustedUsers = [ "deployer" ];
+  }
+
+This will let the user copy Nix store paths to the target.
+
+Let the deploying user use sudo:
+
+.. code-block:: nix
```
cole-h, line 43:
```suggestion
.. code-block:: nix

```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,93 @@
...
+  {
+    users.users.deployer.extraGroups = [ "wheel" ];
+  }
+
+Then, we configure the machine to have passwordless sudo:
+
+.. code-block:: nix
```
cole-h, line 50:
```suggestion
.. code-block:: nix

```

doc/guides/deploy-without-root.rst
```diff
@@ -0,0 +1,96 @@
...
+
+  {
+    network.description = "Non-root deployment";
+
+    hermes =
+      { resources, ... }:
+      {
+        deployment.targetUser = "deployer";
+      };
+    }
```
cole-h, line 77:
```suggestion
    };
```

adisbladis:
:tada: :rocket:

PR #1271, opened 2020-03-31
Remove support for encrypted links

adisbladis:
If someone still wants to use this implementation it can be split into
a plugin.

I'd be more inclined to use something based on Wireguard nowadays.


AmineChikhaoui:
hm ok that will make our switch harder at work, we use this.

grahamc:
Nice. I think we could add the hooks needed for a plugin to implement it, and those same hooks could be used to make wireguard integration automatic in the same way. What do you think @AmineChikhaoui ?

PR #1275, opened 2020-04-02
Move from xml intermediate Nix representation to JSON

adisbladis:
This change is intended to make life easier for plugin authors.
We have removed the XML parameter to ResourceDefinition and you are
now only provided with a name & a dict containing the evaluated
values.

Note that this **_will_** break all current plugins.

An example of this in use in a plugin can be found here: https://github.com/adisbladis/nixops-terraform.

### DIFF
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -113,6 +113,34 @@ Important Notes
              os.path.dirname(os.path.abspath(__file__)) + "/nix"
          ]
 
+5. Resource subclasses must now work with Python objects instead of XML
+
+   This old-style ResourceDefinition subclass:
+
+   .. code-block:: python
+
+     class MachineDefinition(nixops.resources.ResourceDefinition):
+
+         def __init__(self, xml):
+             super().__init__(xml)
+             self.store_keys_on_machine = (
+                 xml.find("attrs/attr[@name='storeKeysOnMachine']/bool").get("value")
+                 == "true"
+             )
+
+   Should now look like:
+
+   .. code-block:: python
+
+     class MachineDefinition(nixops.resources.ResourceDefinition):
+
+         store_keys_on_machine: bool
+
+         def __init__(self, name: str, config: nixops.resources.ResourceOptions):
+             super().__init__(name, config)
+             self.store_keys_on_machine = config["storeKeysOnMachine"]
+
+
 On with Poetry
 ----
 
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -136,7 +136,7 @@ Important Notes
 
          store_keys_on_machine: bool
 
-         def __init__(self, name: str, config: nixops.resources.ResourceOptions):
+         def __init__(self, name: str, config: nixops.resources.ResourceEval):
              super().__init__(name, config)
              self.store_keys_on_machine = config["storeKeysOnMachine"]
 
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -132,13 +132,25 @@ Important Notes
 
    .. code-block:: python
 
+     class MachineOptions(nixops.resources.ResourceOptions):
+         storeKeysOnMachine: bool
+
      class MachineDefinition(nixops.resources.ResourceDefinition):
+
+         config: MachineOptions
 
          store_keys_on_machine: bool
 
          def __init__(self, name: str, config: nixops.resources.ResourceEval):
              super().__init__(name, config)
-             self.store_keys_on_machine = config["storeKeysOnMachine"]
+             self.store_keys_on_machine = config.storeKeysOnMachine
+
+   ``ResourceEval`` is an immutable ``typing.Mapping`` implementation.
+   Also note that ``ResourceEval`` has turned Nix lists into Python tuples, dictionaries into ResourceEval objects and so on.
+   ``typing.Tuple`` cannot be used as it's fixed-size, use ``typing.Sequence`` instead.
+
+   ``ResourceOptions`` is an immutable object that provides type validation both with ``mypy`` _and_ at runtime.
+   Any attributes which are not explicitly typed are passed through as-is.
 
 
 On with Poetry
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -119,7 +119,7 @@ Important Notes
 
    .. code-block:: python
 
-     class MachineDefinition(nixops.resources.ResourceDefinition):
+     class NeatCloudMachineDefinition(nixops.resources.ResourceDefinition):
 
          def __init__(self, xml):
              super().__init__(xml)
@@ -132,10 +132,10 @@ Important Notes
 
    .. code-block:: python
 
-     class MachineOptions(nixops.resources.ResourceOptions):
+     class NeatCloudMachineOptions(nixops.resources.ResourceOptions):
          storeKeysOnMachine: bool
 
-     class MachineDefinition(nixops.resources.ResourceDefinition):
+     class NeatCloudMachineDefinition(nixops.resources.ResourceDefinition):
 
          config: MachineOptions
 
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -7,48 +7,24 @@ from typing import Dict, Any, List, Optional, Union, Set
 import nixops.util
 import nixops.resources
 import nixops.ssh_util
-import xml.etree.ElementTree as ET
 
 
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
 
-    def __init__(self, xml, config={}) -> None:
-        nixops.resources.ResourceDefinition.__init__(self, xml, config)
-        self.ssh_port = int(xml.find("attrs/attr[@name='targetPort']/int").get("value"))
-        self.always_activate = (
-            xml.find("attrs/attr[@name='alwaysActivate']/bool").get("value") == "true"
-        )
-        self.owners = [
-            e.get("value")
-            for e in xml.findall("attrs/attr[@name='owners']/list/string")
-        ]
-        self.has_fast_connection = (
-            xml.find("attrs/attr[@name='hasFastConnection']/bool").get("value")
-            == "true"
-        )
+    ssh_port: int
+    always_activate: bool
+    owners: List[str]
+    has_fast_connection: bool
+    keys: Dict[str, Dict]
 
-        def _extract_key_options(x: ET.Element) -> Dict[str, str]:
-            opts = {}
-            for (key, xmlType) in (
-                ("text", "string"),
-                ("keyFile", "path"),
-                ("destDir", "string"),
-                ("user", "string"),
-                ("group", "string"),
-                ("permissions", "string"),
-            ):
-                elem = x.find("attrs/attr[@name='{0}']/{1}".format(key, xmlType))
-                if elem is not None:
-                    value = elem.get("value")
-                    if value is not None:
-                        opts[key] = value
-            return opts
-
-        self.keys = {
-            k.get("name"): _extract_key_options(k)
-            for k in xml.findall("attrs/attr[@name='keys']/attrs/attr")
-        }
+    def __init__(self, name: str, config: nixops.resources.ResourceOptions):
+        super().__init__(name, config)
+        self.ssh_port = config["targetPort"]
+        self.always_activate = config["alwaysActivate"]
+        self.owners = config["owners"]
+        self.has_fast_connection = config["hasFastConnection"]
+        self.keys = config["keys"]
 
 
 class MachineState(nixops.resources.ResourceState):
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -3,22 +3,33 @@
 import os
 import re
 import subprocess
-from typing import Dict, Any, List, Optional, Union, Set
+from typing import Mapping, Any, List, Optional, Union, Set, Sequence
 import nixops.util
 import nixops.resources
 import nixops.ssh_util
 
 
+class MachineOptions(nixops.resources.ResourceOptions):
+    targetPort: int
+    alwaysActivate: bool
+    owners: Sequence[str]
+    hasFastConnection: bool
+    keys: Mapping[str, Mapping]
+    nixosRelease: str
+
+
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
+
+    config: MachineOptions
 
     ssh_port: int
     always_activate: bool
     owners: List[str]
     has_fast_connection: bool
-    keys: Dict[str, Dict]
+    keys: Mapping[str, Mapping]
 
-    def __init__(self, name: str, config: nixops.resources.ResourceOptions):
+    def __init__(self, name: str, config: nixops.resources.ResourceEval):
         super().__init__(name, config)
         self.ssh_port = config["targetPort"]
         self.always_activate = config["alwaysActivate"]
@@ -37,7 +48,7 @@ class MachineState(nixops.resources.ResourceState):
     ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
     ssh_port: int = nixops.util.attr_property("targetPort", 22, int)
     public_vpn_key: Optional[str] = nixops.util.attr_property("publicVpnKey", None)
-    keys: Dict[str, str] = nixops.util.attr_property("keys", {}, "json")
+    keys: Mapping[str, str] = nixops.util.attr_property("keys", {}, "json")
     owners: List[str] = nixops.util.attr_property("owners", [], "json")
 
     # Nix store path of the last global configuration deployed to this
@@ -178,7 +189,7 @@ class MachineState(nixops.resources.ResourceState):
             "don't know how to remove a backup for machine ‘{0}’".format(self.name)
         )
 
-    def get_backups(self) -> Dict[str, Dict[str, Any]]:
+    def get_backups(self) -> Mapping[str, Mapping[str, Any]]:
         self.warn("don't know how to list backups for ‘{0}’".format(self.name))
         return {}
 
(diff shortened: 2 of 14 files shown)


nixops/deployment.py
```diff
@@ -440,16 +439,15 @@ def evaluate_args(self) -> Any:
...
-            # FIXME: use --json
-            xml = subprocess.check_output(
+            _json = subprocess.check_output(
                 ["nix-instantiate"]
                 + self.extra_nix_eval_flags
                 + self._eval_flags(self.nix_exprs)
                 + [
                     "--eval-only",
-                    "--xml",
+                    "--json",
```
grahamc, line 450: Unfortunately this is not quite equivalent:

```
[grahamc@Petunia:~]$ nix-instantiate  --readonly-mode --eval-only --xml --strict ./test.nix
<?xml version='1.0' encoding='utf-8'?>
<expr>
  <attrs>
    <attr column="3" line="2" name="foo" path="/home/grahamc/test.nix">
      <path value="/home/grahamc/test.nix" />
    </attr>
  </attrs>
</expr>

[grahamc@Petunia:~]$ nix-instantiate  --readonly-mode --eval-only --json --strict ./test.nix
{"foo":"/nix/store/ml1vsmamxm0ijbc67g7za455n3g75j1l-test.nix"}%                                                                                                                                                     
```

I wonder if there is a way to get the XML behavior under the JSON output mode.

grahamc: or changing the `keyFile`'s module behavior to `apply` `tostring` to the file paths

nixops/backends/__init__.py
```diff
@@ -7,52 +7,19 @@
...
-                    value = elem.get("value")
-                    if value is not None:
-                        opts[key] = value
-            return opts
-
-        self.keys = {
-            k.get("name"): _extract_key_options(k)
-            for k in xml.findall("attrs/attr[@name='keys']/attrs/attr")
-        }
+    def __init__(self, name: str, config: Dict):
```
grahamc, line 15: I'm inclined to make a new type other than `Dict` here, even if it was just a wrapper to Dict. That way we could add new superpowers to it later, or new rules later. What do you think?

nixops/backends/__init__.py
```diff
@@ -7,52 +7,19 @@
...
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
 
-    def __init__(self, xml, config={}) -> None:
-        nixops.resources.ResourceDefinition.__init__(self, xml, config)
-        self.store_keys_on_machine = (
-            xml.find("attrs/attr[@name='storeKeysOnMachine']/bool").get("value")
-            == "true"
-        )
-        self.ssh_port = int(xml.find("attrs/attr[@name='targetPort']/int").get("value"))
```
grahamc, line 22: I couldn't remember if this would seek down in to the tree or if it was a "rooted" expression:

I wrote this XML:

```xml

<foo>
  <attrs> 
    <attr name="targetPort"> 
      <int>1</int> 
    </attr> 
  </attrs>
</foo>
```

and this xpath expression: `attrs/attr[@name='targetPort']/int` and it didn't find it, but did when I removed the wrapping `foo`. In other words, this code didn't have the edge case I was worried about.

I guess that means the Backends are initialized with the `deployment` attrset as the config?

nixops/backends/__init__.py
adisbladis: I'm up for that. I was considering implementing an immutable dict-like type we could use instead of a plain dict.

nixops/backends/__init__.py
adisbladis: > I guess that means the Backends are initialized with the deployment attrset as the config?

Yes, that's correct.

nixops/backends/__init__.py
```diff
@@ -7,52 +7,26 @@
...
-            for k in xml.findall("attrs/attr[@name='keys']/attrs/attr")
-        }
+    store_keys_on_machine: bool
+    ssh_port: int
+    always_activate: bool
+    owners: List[str]
+    has_fast_connection: bool
+    keys: Dict[str, Dict]
+
+    def __init__(self, name: str, config: nixops.resources.ConfigDict):
```
grahamc, line 22: _after some brainstorming on the call_ let's call these ResourceOptions.

grahamc: Adam says these would be immutable, ideally, and I think that is a good idea.

nixops/util.py
```diff
@@ -59,6 +72,38 @@ def __str__(self) -> str:
         return "{0} (exit code {1})".format(self.message, self.exitcode)
 
 
+class ImmutableMapping(Mapping):
```
grahamc, line 75: looks good but let's get a test on this. I don't quite understand it all, and I think some tests will help us make sure we understand / don't break what is going on.

grahamc:
lgtm :)

grahamc:
This is great! I think the only thing it needs from here is a bit of docs for a migration guide

nixops/backends/__init__.py
```diff
@@ -7,52 +7,26 @@
 import nixops.util
 import nixops.resources
 import nixops.ssh_util
-import xml.etree.ElementTree as ET
 
 
 class MachineDefinition(nixops.resources.ResourceDefinition):
     """Base class for NixOps machine definitions."""
 
-    def __init__(self, xml, config={}) -> None:
```
grahamc, line 16: What is the difference here between `config` and `xml`? A literal translation might be `s/xml/json/` and keeping `config`,  what is in `config` which we're giving up?

adisbladis: `xml` is an instance of `xml.etree.ElementTree.Element` 

`config` was a previous attempt at making Python objects from the eval rather than working with the XML tree in every resource using this method https://github.com/NixOS/nixops/blob/ba0c3f6/nixops/util.py#L449.

They both contain the same data, just in different formats. It seems to me like the `config` parameter never really got any usage anywhere.

nixops/resources/__init__.py
```diff
@@ -3,12 +3,17 @@
...
 from threading import Event
-from typing import List, Optional, Dict, Any
+from typing import List, Optional, Dict, Any, Iterator
 from nixops.state import StateDict
 from nixops.diff import Diff, Handler
+from nixops.util import ImmutableMapping
 
 
-class ResourceDefinition(object):
+class ResourceOptions(ImmutableMapping[Any, Any]):
```
grahamc, line 12: I wonder if we can get some good juju from https://www.python.org/dev/peps/pep-0589/ here, somehow

adisbladis: `TypedDict` certainly looks interesting but looking into it I'm not sure it's the correct choice (also not sure that it's _not_ the correct choice mind you).
`TypedDict` would give us type safety both at the _instantiation site_ and at the point of _usage_ using the familiar `__getitem__` interface.

If we want to add types to `ImmutableMapping` I think we may want to remove the `__getitem__` overload and only keep the `__getattr__` one and type it like any other object?
``` python
class MyMapping(ImmutableMapping):
    x: int
    y: str
```

An advantage of this approach could be composability:
``` python
class MyMappingBase(ImmutableMapping):
    x: int


class MyMapping(MyMappingBase):
    y: str


class SomeMapping(ImmutableMapping)
    aoeu: str
    submap: MyMapping


m = SomeMapping({
     "aoeu": "something",
     "submap": {  # submap is converted to a `MyMapping` in SomeMapping.__init__
       "x": 1,
       "y": "somethingelse",
     },
})


print(m.submap.x.lower())  # Mypy fails because `x` is an int
```
This approach would not give us type safety at _instantiation time_ (because we're passing a `dict`) but we could get some type safety out of this approach where the "mapping" is used.
My feeling is that we're very close to an immutable `dataclass` at this point (without the disadvantages that comes with subclassing a `dataclass`).

In this specific instance I don't think having type safety at instantiation gives us much since we're passing a dict from a Nix eval that we can't really know ahead of time what it contains.
That could be "solved" with some runtime reflection and throwing errors at runtime if we want to. Keep in mind that runtime checking of types does not work with generics so we'd lose that.

doc/plugins/authoring.rst
```diff
@@ -113,6 +113,46 @@ Important Notes
              os.path.dirname(os.path.abspath(__file__)) + "/nix"
          ]
 
+5. Resource subclasses must now work with Python objects instead of XML
+
+   This old-style ResourceDefinition subclass:
+
+   .. code-block:: python
+
+     class MachineDefinition(nixops.resources.ResourceDefinition):
```
grahamc, line 122: We might want to make this more specific. Like, `NeatCloudMachineOptions`, to hint that it is specific to their module. (I imagine instructions like this will get a lot of copy-paste.)

grahamc, line 122: We might want to make this more specific. Like, `NeatCloudMachineOptions`, to hint that it is specific to their module. (I imagine instructions like this will get a lot of copy-paste.)

doc/plugins/authoring.rst
```diff
@@ -113,6 +113,46 @@ Important Notes
...
+
+         store_keys_on_machine: bool
+
+         def __init__(self, name: str, config: nixops.resources.ResourceEval):
+             super().__init__(name, config)
+             self.store_keys_on_machine = config.storeKeysOnMachine
+
+   ``ResourceEval`` is an immutable ``typing.Mapping`` implementation.
+   Also note that ``ResourceEval`` has turned Nix lists into Python tuples, dictionaries into ResourceEval objects and so on.
+   Please try to use generic types such as ``typing.Mapping`` instead of ``typing.Dict`` and ``typing.Sequence`` instead of ``typing.List``.
```
grahamc, line 150: > Please try to use generic types such

Why?

doc/plugins/authoring.rst
```diff
@@ -113,6 +113,46 @@ Important Notes
...
+
+   ``ResourceOptions`` is an immutable object that provides type validation both with ``mypy`` _and_ at runtime.
+   Any attributes which are not explicitly typed are passed through as-is.
```
grahamc, line 153: what about something like:

```suggestion
   Any attributes which are not explicitly typed default to `Any` and are passed through as-is.
```

nixops/backends/__init__.py
```diff
@@ -3,56 +3,42 @@
...
-import xml.etree.ElementTree as ET
+
+
+class MachineOptions(nixops.resources.ResourceOptions):
+    storeKeysOnMachine: bool
+    targetPort: int
+    alwaysActivate: bool
+    owners: Sequence[str]
+    hasFastConnection: bool
+    keys: Mapping[str, Mapping]
```
grahamc, line 18: Does this make the interior Mapping `Any, Any`? Just curious. We don't need to fix the world in this PR.

adisbladis: Yes, that's the default.

grahamc:
Nice, when causing a developer-error in the type of a module option:

```
TypeError: type of hasFastConnection must be bool; got str instead
```

grahamc:
Not sure if this is from thi sbranch or on master already, but:

```
kif.........> uploading key ‘vault-login’...
kif.........> error: Traceback (most recent call last):
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/deployment.py", line 910, in worker
    m.send_keys()
  File "/home/grahamc/projects/github.com/NixOS/nixops/nixops/backends/__init__.py", line 275, in send_keys
    f.write(opts["text"])
TypeError: write() argument must be str, not None
```

Will look!

doc/plugins/authoring.rst
adisbladis: It doesn't default to `Any` though, anything untyped is considered a non-existing attribute by `mypy`.

grahamc: TIL :)

grahamc:
I'm not getting that write error on master. It seems to come down to this:

https://github.com/NixOS/nixops/blob/6f4f4da9537e8b7fe082e5c35fd5a7fab997d266/nixops/backends/__init__.py#L287-L296


I'm seeing this json come through with a null text field:

```json
{
  "machines": {
    "kif": {
      "keys": {
        "xxx": {
          "_module": {
            "args": {
              "name": "xxx"
            },
            "check": true
          },
          "destDir": "/run/keys",
          "group": "root",
          "keyFile": "/home/grahamc/projects/github.com/xxx",
          "path": "/run/keys/xxx",
          "permissions": "0600",
          "text": null,
          "user": "root"
        },
```

the XML output:

```xml
<?xml version='1.0' encoding='utf-8'?>
<expr>
  <attrs>
    <attr column="5" line="118" name="machines" path="/home/grahamc/projects/github.com/NixOS/nixops/nix/eval-machine-info.nix">
      <attrs>
        <attr name="kif">
          <attrs>
            <attr column="40" line="121" name="keys" path="/home/grahamc/projects/github.com/NixOS/nixops/nix/eval-machine-info.nix">
              <attrs>
                <attr name="xxx">
                  <attrs>
                    <attr name="_module">
                      <attrs>
                        <attr name="args">
                          <attrs>
                            <attr name="name">
                              <string value="xxx" />
                            </attr>
                          </attrs>
                        </attr>
                        <attr name="check">
                          <bool value="true" />
                        </attr>
                      </attrs>
                    </attr>
                    <attr name="destDir">
                      <string value="/run/keys" />
                    </attr>
                    <attr name="group">
                      <string value="root" />
                    </attr>
                    <attr name="keyFile">
                      <path value="/home/grahamc/projects/github.com/xxx" />
                    </attr>
                    <attr name="path">
                      <string value="/run/keys/xxx" />
                    </attr>
                    <attr name="permissions">
                      <string value="0600" />
                    </attr>
                    <attr name="text">
                      <null />
                    </attr>
                    <attr name="user">
                      <string value="root" />
                    </attr>
                  </attrs>
                </attr>
```

so these seem to be the same, so I'm a bit confused.

grahamc:
This is why:

https://github.com/NixOS/nixops/pull/1275/files#diff-6ea77fd00d5f092f8150c2398eaf53d8L48-L49

By default, `None`s would not get assigned in the first place.

grahamc:
I tried adding a type to the keys:

```diff
diff --git a/nixops/backends/__init__.py b/nixops/backends/__init__.py
index 4428de77..127107b1 100644
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -8,6 +8,14 @@ import nixops.util
 import nixops.resources
 import nixops.ssh_util

+class KeyOptions(nixops.resources.ResourceOptions):
+    text: Optional[str]
+    keyFile: Optional[str]
+    destDir: str
+    user: str
+    group: str
+    permissions: str
+    fizz: str

 class MachineOptions(nixops.resources.ResourceOptions):
     storeKeysOnMachine: bool
@@ -15,7 +23,7 @@ class MachineOptions(nixops.resources.ResourceOptions):
     alwaysActivate: bool
     owners: Sequence[str]
     hasFastConnection: bool
-    keys: Mapping[str, Mapping]
+    keys: Mapping[str, KeyOptions]
     nixosRelease: str


@@ -29,7 +37,7 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
     always_activate: bool
     owners: List[str]
     has_fast_connection: bool
-    keys: Mapping[str, Mapping]
+    keys: Mapping[str, KeyOptions]

     def __init__(self, name: str, config: nixops.resources.ResourceEval):
         super().__init__(name, config)
@@ -38,7 +46,8 @@ class MachineDefinition(nixops.resources.ResourceDefinition):
         self.always_activate = config["alwaysActivate"]
         self.owners = config["owners"]
         self.has_fast_connection = config["hasFastConnection"]
-        self.keys = config["keys"]
+        self.keys = {k: KeyOptions(**v) for k,v in config["keys"].items()}
+


 class MachineState(nixops.resources.ResourceState):
@@ -255,13 +264,14 @@ class MachineState(nixops.resources.ResourceState):
             return
         if self.store_keys_on_machine:
             return
+        opts: KeyOptions
         for k, opts in self.get_keys().items():
+            from pprint import pprint
+            pprint(opts)
             self.log("uploading key ‘{0}’...".format(k))
             tmp = self.depl.tempdir + "/key-" + self.name
-            if "destDir" not in opts:
-                raise Exception("Key '{}' has no 'destDir' specified.".format(k))

-            destDir = opts["destDir"].rstrip("/")
+            destDir = opts.destDir.rstrip("/")
             self.run_command(
                 (
                     "test -d '{0}' || ("
@@ -270,11 +280,11 @@ class MachineState(nixops.resources.ResourceState):
                 ).format(destDir)
             )

-            if "text" in opts:
+            if opts.text is not None:
                 with open(tmp, "w+") as f:
-                    f.write(opts["text"])
-            elif "keyFile" in opts:
-                self._logged_exec(["cp", opts["keyFile"], tmp])
+                    f.write(opts.text)
+            elif opts.keyFile is not None:
+                self._logged_exec(["cp", opts.keyFile, tmp])
             else:
                 raise Exception(
                     "Neither 'text' or 'keyFile' options were set for key '{0}'.".format(
@@ -306,7 +316,7 @@ class MachineState(nixops.resources.ResourceState):
                         "chmod '{3}' {0}",
                     ]
                 ).format(
-                    tmp_outfile_esc, opts["user"], opts["group"], opts["permissions"]
+                    tmp_outfile_esc, opts.user, opts.group, opts.permissions
                 )
             )
             self.run_command("mv " + tmp_outfile_esc + " " + outfile_esc)
```

I was surprised that mypy was totally happy with this, but it doesn't actually work or validate the data. I think this is what you were saying w.r.t. checking generics. Now I'm curious about if this is solved elsewhere, like by pydantic? I'd really like to be able to have typed mappings.

grahamc:
woot:

```
Imprecision went up:
nixops.util:		17.62 -> 18.2
Imprecision went down:
nixops.backends:		38.92 -> 37.24
nixops.backends.none:		54.7 -> 50.0
nixops.deployment:		17.55 -> 16.32
nixops.resources:		39.53 -> 36.42
nixops.resources.commandOutput:		19.47 -> 17.5
nixops.resources.ssh_keypair:		48.33 -> 44.26
Total:		25.3 -> 24.6
```

tewfik-ghariani:
Can we create a branch 'pre-xml-to-json' to keep the plugins alive before working on the switch?

PR #1277, opened 2020-04-07
Docs: Update authoring.rst after a few more implementations.

grahamc:


cole-h:
Might be worth it to also highlight the other Nix blocks, like in the [On with Poetry](https://github.com/NixOS/nixops/blob/05c04b7269894ad249e8fc94594aebe0ce70f503/doc/plugins/authoring.rst#on-with-poetry) section.

doc/plugins/authoring.rst
```diff
@@ -30,8 +30,10 @@ configuration.
 
 First, create a ``pyproject.toml`` (see `PEP-0517
 <https://www.python.org/dev/peps/pep-0517/>`_ to describe your
-project. This is intsead of a ``setup.py``, and using both may cause
-confusing build errors. Only use a ``pyproject.toml``::
+project. This is instead of a ``setup.py``, and using both may cause
+confusing build errors. Only use a ``pyproject.toml``:
+
```
worldofpeace, line 35: "This is instead of a  just sounds weird to me.

We do not use `setup.py`, and trying to use both a `setup.py` and `pyproject.toml` may cause confusing build errors.

grahamc: fixup'd!

PR #1280, opened 2020-04-14
Added a keyCmd option for fetching a key from local command

kisik21:
The option `deployment.keys.*.keyCmd` executes a command on the local machine and sends its output as a key to the remote machine.

The use-case is storing NixOps secrets in encrypted form using, e.g. password-store.

The patch should also apply to the `flake-support` branch - since I developed this feature with flakes in mind, should I file another pull-request to cherry-pick it there?

### DIFF
--- a/nix/keys.nix
+++ b/nix/keys.nix
@@ -15,11 +15,33 @@ let
         <filename><replaceable>destDir</replaceable>/<replaceable>password</replaceable></filename>
         will be <literal>foobar</literal>.
 
-        NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
-        to be set.
+        NOTE: Either <literal>text</literal>, <literal>keyCmd</literal> or
+        <literal>keyFile</literal> have to be set.
       '';
     };
 
+    options.keyCmd = mkOption {
+      default = null;
+      example = "pass show secrettoken";
+      type = types.nullOr types.str;
+      description = ''
+        When non-null, output of this command run on local machine will be
+        deployed to the specified key on the target machine.  If the key name
+        is
+        <replaceable>password</replaceable> and <literal>echo secrettoken</literal>
+        is set here, the contents of the file
+        <filename><replaceable>destDir</replaceable>/<replaceable>password</replaceable></filename>
+        deployed will equal the output of the command <literal>echo secrettoken</literal>.
+
+        This option is especially useful when you don't want to store the secrets
+        inside of your NixOps deployment but rather in a well-guarded place such as an
+        encrypted file. Consider using nixpkgs.password-store as storage for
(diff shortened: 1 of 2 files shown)


grahamc:
I'm excited for this PR. I'd like to merge it after #1275. Can you try to address the failing checks? Thank you!

nix/keys.nix
```diff
@@ -15,11 +15,33 @@ let
...
-        to be set.
+        NOTE: Either <literal>text</literal>, <literal>keyCmd</literal> or
+        <literal>keyFile</literal> have to be set.
       '';
     };
 
+    options.keyCmd = mkOption {
+      default = null;
+      example = "pass show secrettoken";
+      type = types.nullOr types.str;
```
grahamc, line 26: What if this was a listof str's, which are executed without a shell?

kisik21: @grahamc well I could do that thing but I'm not sure how to process it on Python side. I mostly copy-pasted python code for this from other options :3 need to study the internal APIs more I guess?

nix/keys.nix
```diff
@@ -15,11 +15,33 @@ let
...
         will be <literal>foobar</literal>.
 
-        NOTE: Either <literal>text</literal> or <literal>keyFile</literal> have
-        to be set.
+        NOTE: Either <literal>text</literal>, <literal>keyCmd</literal> or
+        <literal>keyFile</literal> have to be set.
       '';
     };
 
+    options.keyCmd = mkOption {
```
grahamc, line 23:
```suggestion
    options.keyCommand = mkOption {
```

grahamc:
:partying_face:

kisik21:
OH NO! I hope you fixed the bug I may or may not have introduced earlier!!!

kisik21:
Whew, you did fix that bug. I was **so worried** about it, I almost had a heart attack

grahamc:
What was the bug you might have introduced? :D :) <3

grahamc:
btw the test deployment we ran used `nix-shell -p fortune --run fortune` to generate our secret. Great success:

```
[grahamc@kif:~]$ sudo cat /run/keys/my-cool-keycommand
If everything is coming your way then you're in the wrong lane.
```

kisik21:
Sometimes it wasn't waiting for the command to complete, and the secret was empty. Should be fine now, but just to check: make a script that waits around 15 seconds and see if Nixops blocks on it. If it doesn't, the bug is there.

The fix should've been in one of my branches I use for flakes, so if it isn't fixed, I'll quickly PR it here

grahamc:
It definitely waits :) Thanks for the PR, Vika!

kisik21:
:heart:

Issue #1150, opened 2019-05-16
Bastion/Jump host support


Nekroze:
Hi, when trying to deploy best practices infrastructure to the cloud, using a small jump/bastion machine being the only machine exposed to the internet is a common pattern. However tooling needs to know how to reach machines behind the jump host, SSH supports this natively these days (`ssh -J`) and so should NixOps.

Personally I want to use this on aws resources however it should be an option for all deployments over SSH.

aij:
Does it still work if you configure it in `~/.ssh/config`? That worked when I needed it last, but things may have changed since then.

(And I realize that's probably not very convenient in general.)

Nekroze:
I'll have to give that a go, it is this week I will be needing to use this so i will have some time to test various work around for suitability throughout the week. However, that does not really help if nixops is the one creating the jumpbox so there is no ahead of time information (without creating resources I would rather not like a route53 just for this) to configure with.

If there is not a suitable work around I will look at how the python code handles the ssh connection and see if I can make PR for this feature my self... time permitting.

Issue #784, opened 2017-11-28
Incorrect logic for checking for ssh port availability on nixops deploy


kuznero:
`nixops deploy` does not take `~/.ssh/config` or `/etc/ssh/ssh_config` into account. But `nixops ssh` does.

*UPDATE* Incorrect logic for checking for ssh port availability on nixops deploy (has nothing to do with adhering to ssh configs, please read the thread)

edolstra:
Please provide more detail.

kuznero:
Generally I am trying to learn how to work with `nixops` behind a corporate proxy. `ssh` can be treated with `~/.ssh/config` file like this:

```
Host *
  ProxyCommand nc -X connect -x 127.0.0.1:3128 %h %p
```

This helps me `ssh` through the proxy (currently I have `cntlm` installed locally that passes through requests to real corporate proxy). This is equivalent to running raw `ssh user@host -o "ProxyCommand nc -X connect -x 127.0.0.1:3128 %h %p"`.

So, `nixops deploy` as one of last steps is trying to `ssh` into provisioned servers (I am testing it with AWS EC2 at the moment). And it does not work. But when I try `nixops ssh server` it works. I also tried playing with adding and remove `~/.ssh/config` file and it seems that it is needed, but it is only `nixops ssh ...` that takes it into account, but not `nixops deploy`.

kuznero:
And even when my machine is configured system-wide with:

```nix
  programs.ssh.extraConfig = ''
Host *
  ProxyCommand nc -X connect -x 127.0.0.1:3128 %h %p
  '';
```

`nixops deploy` still does not take it into account, but `nixops ssh server` does. That is currently a showstopper for using `nixops` behind corporate proxy. At least this is one of the scenarios.

kuznero:
Running `SSH_ARGS="ProxyCommand=nc -X connect -x 127.0.0.1:3128 %h %p" nixops deploy -d box` does not work either.

kuznero:
`waiting for SSH...` is what hangs. So, I guess it does not necessarily indicates that `nixops deploy` does not take ssh configs into account, it might just be waiting for the port openings. But that still hangs for some reason.

kuznero:
I just managed to comment out parts of the `wait_for_ssh` function in `nixops/backends/__init__.py` such that it does not try to wait for ssh port opening without using ssh as a tool:

```python
    def wait_for_ssh(self, check=False):
        """Wait until the SSH port is open on this machine."""
[...]

PR #1283, opened 2020-04-15
Add SSH jump host support

adisbladis:
This work is based on https://github.com/NixOS/nixops/pull/1270 because they touch a lot of the same code.

Additionally as a requisite for this work I have changed the "TCP ping" to actually log in to the remote host and run a command instead of just checking the TCP port availability.

Closes https://github.com/NixOS/nixops/issues/1150
Closes https://github.com/NixOS/nixops/issues/784

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -2,13 +2,14 @@
 from __future__ import annotations
 import os
 import re
-from typing import Mapping, Any, List, Optional, Union, Sequence, TypeVar
+from typing import Mapping, Any, List, Optional, Union, Sequence, TypeVar, Callable
 from nixops.monkey import Protocol, runtime_checkable
 import nixops.util
 import nixops.resources
 import nixops.ssh_util
 from nixops.state import RecordId
 import subprocess
+import threading
 
 
 class KeyOptions(nixops.resources.ResourceOptions):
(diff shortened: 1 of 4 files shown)


adisbladis:
Rebased on master since #1270 was merged.

nixops/backends/__init__.py
```diff
@@ -247,16 +247,29 @@ def reboot_sync(self, hard: bool = False) -> None:
         """Reboot this machine and wait until it's up again."""
         self.reboot(hard=hard)
         self.log_start("waiting for the machine to finish rebooting...")
-        nixops.util.wait_for_tcp_port(
```
grahamc, line 250: Should we explicitly add a deprecation message to this, or even delete the function?

nixops/backends/__init__.py
```diff
@@ -247,16 +247,29 @@ def reboot_sync(self, hard: bool = False) -> None:
...
-            callback=lambda: self.log_continue("."),
+
+        nixops.util.wait_for_fail(
+            fn=lambda: self.ssh.run_command(
+                ["true"],
+                user=self.ssh_user,
+                forward_agent=False,
+                logged=False,
+                connection_tries=1,
+            )
```
grahamc: Let's wrap this in a convenience method of like `is the machine actually reachable?`

nixops/backends/__init__.py
```diff
@@ -247,16 +247,29 @@ def reboot_sync(self, hard: bool = False) -> None:
         """Reboot this machine and wait until it's up again."""
         self.reboot(hard=hard)
         self.log_start("waiting for the machine to finish rebooting...")
-        nixops.util.wait_for_tcp_port(
-            self.get_ssh_name(),
-            self.ssh_port,
-            open=False,
-            callback=lambda: self.log_continue("."),
```
grahamc, line 254: Can we get the `.`s back? I think they are pretty important from a "is nixops even still trying, or is it stuck?" perspective.

nixops/backends/none.py
```diff
@@ -109,6 +109,7 @@ def _check(self, res):
             res.exists = False
             return
         res.exists = True
+        # TODO: Use a better method that works with bastion hosts
         res.is_up = nixops.util.ping_tcp_port(self.target_host, self.ssh_port)
```
grahamc, line 113: I think we're guaranteed to have a working SSH, so let's just use the same "ssh and run `true`" strategy as in other places. @aanderse mentioned some backends might not need to use SSH, but that is a big enough thing to be a whole feature request and "epic".

nixops/util.py
```diff
@@ -398,26 +398,50 @@ def ping_tcp_port(
...
 
 
-def wait_for_tcp_port(
-    ip: str,
-    port: int,
-    timeout: int = -1,
-    open: bool = True,
-    callback: Optional[Callable[[], Any]] = None,
+def wait_for_success(
+    fn: Callable, timeout: int = -1, callback: Optional[Callable[[], Any]] = None,
```
grahamc, line 402:
```suggestion
    fn: Callable, timeout: Optional[int] = None, callback: Optional[Callable[[], Any]] = None,
```

Let's use our type system to our advantage here, and in the next function too.

grahamc:
Since we're opting to go for more of a user-supplied "use sshOptions to do it", we'll need to bolster this use case by writing a user's guide on how to deploy with a jump host.

nixops/backends/__init__.py
lovesegfault: Can there be a real spinner as opposed to just dots? A quick google points me to https://github.com/manrajgrover/py-spinners and https://github.com/verigak/progress/

adisbladis: For now I just re-added the dots. I'd love to make the output more pretty, but that's a separate PR.

PR #1286, opened 2020-04-17
doc: clarify enableRollback

mmilata:
This was unclear to me so let me know if my understanding (and/or wording) is wrong and I'll be happy to update the PR.


domenkozar:
That's a bit confusing to me, they are created remotely or locally?

mmilata:
I haven't dug into the sources, so maybe consider this PR as "question to developers", but my understanding is that:

* network.enableRollback = true -> locally & remotely
* network.enableRollback = false -> remotely

What about:

*Note that enabling the option causes the profiles to be created on the local system from which nixops is executed. Even with network.enableRollback set to false new profile generations are created on the remote machines.*

Thank you for bearing with me while trying to use English.

PR #1288, opened 2020-04-20
Remove deployment.storeKeysOnMachine

adisbladis:
This is a dangerous options and persistent secrets can already be
achieved by setting a destDir to a persistent mutable location.

### DIFF
--- a/nix/auto-luks.nix
+++ b/nix/auto-luks.nix
@@ -48,12 +48,7 @@ with utils;
               the volume.  If left empty, a passphrase is generated
               automatically; this passphrase is lost when you destroy the
               machine or underlying device, unless you copy it from
-              NixOps's state file.  Note that unless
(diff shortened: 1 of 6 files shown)


grahamc:
```
13:32 <gchristensen> is anyone using nixops' storeKeysOnMachine?
13:32 <globin> gchristensen: we were thinking about starting to use it, why?
13:32 <gchristensen> I want to delete it
13:32 <globin> why?
13:32 <gchristensen> would your use case be supported by setting keys.name.destDir  to somewhere persistent instead?
13:33 <gchristensen> globin: it puts keys in the store
13:33 <globin> Ah then we wanted to use the destDir :)
13:33 <globin> ignore me
13:33 <gchristensen> we'll delete it then :P
13:33 <globin> +1
```

Issue #1304, opened 2020-04-20
The README needs some improvements


gilligan:
I think the current state of the README is not ideal. Here are some thoughts on that:

- Whom is the README for? If you land on the nixops page as someone who hasn't already used nixops it doesn't provide a lot of useful information to understand what nixops even is.
- The README links to external resources but it would be great to have some more actual content in the README itself. In fact I think it would be great if the documentation was fully self-contained in this repository
- I don't think the _Issue Tracker_ is helping anyone. Anyone who has used GitHub before is able to locate the _Issues_ tab
- The mailing list link is not valid anymore.
- Building from source has broken markdown syntax (missing closing `)
- The snippet about adding nixops as buildInput seems completely arbitrary and out of place in the README

grahamc:
Ithink this is solved by #1308.

PR #1308, opened 2020-04-20
Redo README.md

gilligan:
This is an attempt at providing a more helpful `README.md` which provides:

- CI status info
- key features
- where to find more info
- how to install
- how to develop
- how to contribute
- LICENSE

This should address most of #1304

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ $ nix-env -if https://github.com/NixOs/nixops/tarball/master
 
 #### Building The Nix Package
 
-You can build the nix package by simply invoking `nix-build` on the project root:
+You can build the Nix package by simply invoking `nix-build` on the project root:
 
 ```
 $ nix-build
--- a/README.md
+++ b/README.md
@@ -1,38 +1,67 @@
 # NixOps
 
-NixOps (formerly known as Charon) is a tool for deploying NixOS
-machines in a network or cloud.
+[![Test](https://github.com/NixOS/nixops/workflows/CI/badge.svg)](https://github.com/NixOS/nixops/actions)
 
-* [Manual](https://hydra.nixos.org/job/nixops/master/tarball/latest/download-by-type/doc/manual)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,38 +1,67 @@
...
+- **Extensible**: _NixOps_ is extensible through a plugin infrastructure which can be used to provide additional backends.
 
-To start developing on nixops, you can run:
+For more information please refer to the [NixOps manual](https://nixos.org/nixos/manual/).
 
-```bash
-  $ ./dev-shell
+### Installing
+
+_NixOps_ is included in nixpkgs and can be installed from your respective channel:
```
grahamc, line 17: We might should warn that this repo contains NixOps 2.0 and the channels don't package it yet?

README.md
```diff
@@ -1,38 +1,67 @@
...
+
+```
+$ nix-env -i nixops
```
grahamc, line 20:
```suggestion
$ nix-shell -p nixops
```

README.md
```diff
@@ -1,38 +1,67 @@
...
+
+```
+$ nix-env -if https://github.com/NixOs/nixops/tarball/master
+```
+
+### Building And Developing
+
+#### Building The Nix Package
+
+You can build the nix package by simply invoking `nix-build` on the project root:
```
grahamc, line 33:
```suggestion
You can build the Nix package by simply invoking `nix-build` on the project root:
```

README.md
gilligan: Oh, yes indeed I should add something to that effect indeed.

README.md
gilligan: I'm totally fine with changing this to `nix-shell` but then the description shouldn't be saying "installing" right? Then it should say something about "trying it in a shell" right?

README.md
gilligan: So I added a Note below, let me know if that's OK for you.

README.md
grahamc: "Use it in a shell" maybe? I've never installed nixops :)

grahamc: Yeah, let's stick to `nix-shell` for here.

gilligan: Alrighty, i changed the text above..

grahamc:
Thanks!

PR #1310, opened 2020-04-20
Extract ci actions to ./ci/ scripts

gilligan:
**what?**
This extracts the actions performed by the GitHub CI into scripts in ./ci/check-*.sh to allow for easier local execution.

**why?**
While it's possible to find out what the github actions do by looking at .github/workflows/ci.yml I think it is nice and convenient when checks are easy to execute locally.


grahamc:
Please rebase?

Issue #1242, opened 2020-03-03
Release 1.8


grahamc:
- [ ] Get mypy types exposed for plugin developers
- [ ] Update AWS plugin for py3
- [ ] Update vbox plugin for py3
- [ ] Update hetzner plugin for py3
- [ ] Solidify the plugin dev / build experience

domenkozar:
If we get first half of that list done for 2.0 I'd say that's enough :)

Nekroze:
> If we get first half of that list done for 2.0 I'd say that's enough :)

Down to the state back ends, I think would be fantastic! Nice list.

arianvp:
>  explore terraform integration

So I looked at this before; and one interesting note is that terraform providers are just  binaries that speak gRPC (using the following definition https://github.com/hashicorp/terraform/blob/master/docs/plugin-protocol/tfplugin5.2.proto) 

The binaries can be run to introspect the schema (and generate nixos modules from them automatically),  take a value that adheres this schema (in the form of "current state") and apply that state


They don't do state management themselves; but at least you can take a declarative `prior_state` + `planned_state` and then apply it to get the `new_state` .  (Hey this is just a very complicated `foldr` type signature ;))

I think having a small shim around terraform provider binaries would be extremely helpful. Turning each terraform provider into a nixops plugin transparently.

This is all a bit undocumented at Terraform's side; (They want you to really really use their Go API) but I think it's doable and relatively stable. It's been an idea that has been on my list for a very long time.

Our core expertise is not describing alll cloud platforms, that's something a lot of people have blood sweat and tears into in the form of `providers`. Our expertise is superious config management for infrastructure descriptions and also provisioning. I think we can have best of both worlds.

adisbladis:
This is a change I'd _really_ like to get in NixOps 2.0: https://github.com/NixOS/nixops/pull/1275.

It makes plugin development so much nicer to only work with structured data.

PsyanticY:
For backups removal, for AWS backend at least, we can use DLM instead: https://github.com/NixOS/nixops-aws/pull/17

We already been using it in prod for a while now

cc @grahamc

PsyanticY:
For auto-luks as well we can use the AWS CMK instead, https://github.com/NixOS/nixops-aws/pull/20
The conflict in the 2 PR is pretty strait-forward to fix if we are interested in merging these PR

jonringer:
[...]

PR #1312, opened 2020-04-21
Remove some niche features

adisbladis:
See https://github.com/NixOS/nixops/issues/1242

Please note that this is _not_ about just removing these features. They will be reintroduced as plugins.


grahamc:
Maybe we should make a repo for useful modules which also don't belong in nixpkgs, and refer to it in a changelog entry?

grahamc:
really nice that they're not actually using any code in nixops itself, and the modules can just be dropped in

grahamc:
LGTM but let's make that repo and move these expressions over. I asked in #nix-community on IRC about a good repo name for it. Once we do and move the exprs, let's make a link to it from the readme.

grahamc:
I've split the history of auto-luks and auto-raid0 in to https://github.com/nix-community/nixos-modules-contrib. We probably should have been doing this from the start, but I'll add a changelog entry for this.

adisbladis:
@grahamc I think we really need a story for how to get https://github.com/nix-community/nixos-modules-contrib into peoples NixOps deployments.
Imo we should make https://github.com/nix-community/nixos-modules-contrib a NixOps plugin.

grahamc:
That sounds good to me. We might want to get the nixos-infect file a better home, too, while we're at it?

adisbladis:
> That sounds good to me. We might want to get the nixos-infect file a better home, too, while we're at it?

Actually thinking about this a bit more I realised it doesn't make much sense to make the external repo a NixOps plugin as it's not directly tied to NixOps, it's just external NixOps modules.

So my recommendation to users in this instance would be to use [Niv](https://github.com/nmattia/niv), or at a later date Nix flakes.

adisbladis:
`nixos-infect` already has another home at https://github.com/elitak/nixos-infect.
We may want to talk to @elitak and get this moved over to https://github.com/nix-community?

grahamc:
I agree about not doing the plugin, and leaving that for other solutions like niv, flakes, or another mechanism. That said, if someone wanted to make the plugin, that would be great.

grahamc:
Just did a rebase and trivial fix-up of including the 2.0 notes in the toc, and fixing header's underbars.

PR #1313, opened 2020-04-21
Add a withPlugins function to the NixOps derivation

adisbladis:
It's a variation of the expression from https://github.com/NixOS/nixpkgs/pull/83548.

With this change you can use the following snippet to build NixOps
with a local plugin:
``` nix
{ pkgs ? import <nixpkgs> {} }:

let
  nixops = import ./default.nix { inherit pkgs; };
in nixops.withPlugins(ps: [
  (ps.callPackage ./path/to/plugin {})
])
```

Plugins can have their own dependencies/requirements which may or may not be fulfilled by the package set `ps`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -17,7 +17,7 @@ let
 
     # The complete buildEnv drv
     buildEnvDrv = interpreter.buildEnv.override {
-      extraLibs = plugins;
+      extraLibs = builtins.map (p: interpreter.pkgs.toPythonModule p) plugins;
     };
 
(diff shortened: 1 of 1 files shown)


adisbladis:
I'm considering whether this can be abstracted in a way that's generic enough to go into `poetry2nix` so that _any_ Python package ecosystem can be "pluginised".

adisbladis:
While this _is_ a blocker for 2.0 I don't like this implementation at all.

PR #1314, opened 2020-04-22
README: tweaks

cole-h:
* point to the NixOps manual instead of the NixOS manual
* add some commas
* adjust wording

---

Hopefully most of these changes are uncontroversial.


grahamc:
Thanks!

PR #1318, opened 2020-04-23
Add plugin hooks required for the encrypted links plugin

adisbladis:
This is being used by https://github.com/adisbladis/nixops-encrypted-links

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -597,3 +597,8 @@ class CheckResult(object):
 
 
 GenericMachineState = MachineState[MachineDefinition]
+
+
+class MachinePlugin(Protocol):
+    def post_wait(self, m: MachineState) -> None:
+        pass
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -1312,6 +1312,9 @@ class Deployment:
 
                         m.wait_for_ssh(check=check)
 
+                        for p in get_plugin_manager().hook.machine_hook():
+                            p.post_wait(m)
+
                 except Exception:
                     r._errored = True
                     raise
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -649,7 +649,7 @@ class Deployment:
             do_machine(m)
 
         def emit_resource(r: nixops.resources.GenericResourceState) -> Any:
-            config = []
+            config: List[Dict[Tuple[str, ...], Any]] = []
             config.extend(attrs_per_resource[r.name])
             if is_machine(r):
                 # Sort the hosts by its canonical host names.
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -19,6 +19,7 @@ import itertools
 import platform
 import time
 import importlib
+from typing_extensions import Protocol
 
 from functools import reduce
 from typing import (
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -1761,5 +1761,12 @@ def _load_modules_from(dir: str) -> None:
         importlib.import_module("nixops." + dir + "." + module[:-3])
 
 
+class DeploymentPlugin(Protocol):
+    def physical_spec(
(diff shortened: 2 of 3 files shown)


nixops/backends/__init__.py
```diff
@@ -50,14 +51,20 @@ def __init__(self, name: str, config: nixops.resources.ResourceEval):
 class MachineState(nixops.resources.ResourceState):
     """Base class for NixOps machine state objects."""
 
+    defn: Optional[MachineDefinition]
```
grahamc, line 54: Should `defn` and its setting, be part of `ResourceState`'s default implementation?

nixops/plugins/enums.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import annotations
```
adisbladis, line 1: Note to self: Delete this file.

nixops/deployment.py
```diff
@@ -139,7 +142,7 @@ def machines(self) -> Dict[str, nixops.backends.MachineState]:
         }
 
     @property
-    def active(
+    def active_machines(
         self,
     ) -> Dict[str, nixops.backends.MachineState]:  # FIXME: rename to "active_machines"
```
grahamc, line 147: Let's mark that FIXME as a FIXEDYOU, and move this rename and its fallout in to a separate commit near the beginning of the series?

nixops/deployment.py
```diff
@@ -551,21 +554,16 @@ def get_arguments(self) -> Any:
...
         }
-        authorized_keys: Dict[str, List[str]] = {
-            m.name: [] for m in active_machines.values()
-        }
-        kernel_modules: Dict[str, Set[str]] = {
-            m.name: set() for m in active_machines.values()
-        }
-        trusted_interfaces: Dict[str, Set[str]] = {
-            m.name: set() for m in active_machines.values()
-        }
```
grahamc, line 568: Let's make these cleanup bits a separate commit, saying they're already unused -- that'll make the addition of plugin hooks easier to parse out.

nixops/deployment.py
```diff
@@ -1699,5 +1679,12 @@ def _load_modules_from(dir: str) -> None:
         importlib.import_module("nixops." + dir + "." + module[:-3])
 
 
+class DeploymentPlugin(Protocol):
+    def physical_spec(
+        self, d: Deployment
+    ) -> Dict[str, List[Dict[Tuple[str, ...], Any]]]:
```
grahamc, line 1685: I think we should probably give this type a name, like:

```
MachineSpecs = Dict[str, List[Dict[Tuple[str, ...], Any]]]
```

or something?

nixops/plugins/enums.py
grahamc, line 1: you asked me to say this file should be deleted :)

nixops/util.py
```diff
@@ -125,8 +126,21 @@ class ImmutableValidatedObject:
 
     _frozen: bool
 
-    def __init__(self, **kwargs):
-        anno: Dict = self.__annotations__
+    def __init__(self, *args: ImmutableValidatedObject, **kwargs):
```
grahamc, line 129: Let's get some tests in for this change, and maybe another commit?

nixops/backends/__init__.py
adisbladis: Yes, I think so.

nixops/deployment.py
adisbladis: Yes, this type is incredibly awkward, it's so awkward in fact that I don't expect plugin authors to directly type these structures.

Two things I have considered instead:
- Simple dict:
``` python
{
  "networking": {"hostName": "foobar"}
}
```

- Abstract away in a class:
``` python
spec = Spec()
spec.add_option("networking.hostName", "foobar")
```

adisbladis:
This PR has been rolled up in #1383

PR #1319, opened 2020-04-24
add nixops eval subcommand

manveru:
This lets you pass a file to eval to `nixops eval`.


grahamc:
Before I review this, a few things: can you fixup the failures with `black`, and take a look to see if you can fix the ratchet's complaints about reduced type coverage? 

And also if you could write a bit about what this is useful for, like what use cases you have for it -- etc. that'd be helpful!

manveru:
Yeah... working on it :)

nixops/__main__.py
```diff
@@ -527,6 +527,16 @@
     help="include the physical specification in the evaluation",
 )
 
+subparser = add_subparser(
+    subparsers, "eval", help="eval the given file is nix code in the network expression"
```
grahamc, line 531:
```suggestion
    subparsers, "eval", help="evaluate a Nix expression with the NixOps network as arguments"
```

my proposed wording isn't quite right, but maybe something like this?

nixops/__main__.py
```diff
@@ -527,6 +527,16 @@
...
+)
+subparser.set_defaults(op=op_eval)
+subparser.add_argument("code", metavar="CODE", help="code")
```
grahamc, line 534:
```suggestion
subparser.add_argument("file", metavar="FILE", help="File containing a Nix expression")
```

nixops/__main__.py
```diff
@@ -527,6 +527,16 @@
...
+subparser.add_argument(
+    "--json", action="store_true", help="print the option value in JSON format"
```
grahamc, line 536:
```suggestion
    "--json", action="store_true", help="print the resulting value as an JSON representation of the abstract syntax tree"
```

what does it print without `--json`?

nixops/__main__.py
```diff
@@ -527,6 +527,16 @@
...
+)
+subparser.add_argument("--strict", action="store_true", help="enable strict evaluation")
```
grahamc, line 538: I don't have a specific suggestion, but we might want to crib what the nix-instantiate manual says:


> recursively evaluate list elements and attributes. Normally, such sub-expressions are left unevaluated
> (since the Nix expression language is lazy).
>    Warning
>    This option can cause non-termination, because lazy data structures can be infinitely large.

I wonder why someone wouldn't use `--strict --json`? Should `--json` imply `--strict`?

nixops/script_defs.py
```diff
@@ -893,6 +893,14 @@ def op_show_option(args):
         )
 
 
+def op_eval(args):
+    with deployment(args) as depl:
+        depl.evaluate()
```
grahamc, line 898: Not necessarily to be fixed here, but it would be nice to think of a way for the deployment API to not require being careful about calling `evaluate`.

nixops/__main__.py
manveru: Yes, `--json` will only work with `--strict` if it's more than one level deep.

The use-case here was if you wanted to evaluate something but finely control the level of strictness and output Nix instead of JSON. For that you'd want lazy evaluation of functions, which is impossible if `--strict` is hardcoded.

Given that this is supposed to be used by other tooling (that's why I don't pass a plain string but a file instead), it shouldn't be too confusing or hard to configure it for the specific usage.

PR #1323, opened 2020-04-24
Use rsync instead of scp for file copying

### LINKED ISSUE #1322: Use rsync instead of scp
This will allow us to pass options to openssh and escalate privileges (see https://github.com/NixOS/nixops/pull/1270) more easily.

adisbladis:
This will allow us to pass options to openssh and escalate privileges (see #1270) more easily.

Closes #1322

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -18,6 +18,7 @@ in pkgs.poetry2nix.mkPoetryApplication {
 
   propagatedBuildInputs = [
     pkgs.openssh
+    pkgs.rsync
   ];
 
   nativeBuildInputs = [
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -419,27 +419,35 @@ class MachineState(nixops.resources.ResourceState):
             env=env,
         )
 
-    def get_scp_name(self):
+    def _get_scp_name(self) -> str:
         ssh_name = self.get_ssh_name()
         # ipv6 addresses have to be wrapped in brackets for scp
         if ":" in ssh_name:
             return "[%s]" % (ssh_name)
         return ssh_name
 
-    def upload_file(self, source, target, recursive=False):
+    def _fmt_rsync_command(self, *args: str, recursive: bool = False) -> List[str]:
         master = self.ssh.get_master()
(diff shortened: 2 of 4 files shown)


lordcirth:
I cannot find a reference now, but I distinctly recall the openssh devs mentioning that scp was supported for legacy reasons, and not really recommended. It was during the discussion of CVE-2018-20685.

default.nix
```diff
@@ -18,6 +18,7 @@ in pkgs.poetry2nix.mkPoetryApplication {
 
   propagatedBuildInputs = [
     pkgs.openssh
+    pkgs.rsync
```
LnL7, line 21: It's probably better to use `propagatedUserEnvPkgs` for these, that should ensure nix-env and environment.systemPackages include the binaries along with nixops.

adisbladis: `propagatedBuildInputs` are wrapped by the Python builder, `rsync` shouldn't necessarily end up in `$PATH` but `nixops` needs to know where to find it.

nixops/backends/__init__.py
```diff
@@ -419,27 +419,35 @@ def copy_closure_to(self, path):
             env=env,
         )
 
-    def get_scp_name(self):
+    def _get_scp_name(self) -> str:
```
grahamc, line 422: double-checked that this wasn't used in plugins. I probably wasn't conclusive, but I checked at least GCE, AWS, and nix-community's plugins.

grahamc:
Just looking at the semantics of the trailing slash in `rsync` vs. `scp`. 

rsync's manpage:

> You use rsync in the same way you use rcp. You must specify a source and a destination, one of which may be remote.

scp's manpage:

> scp is based on the rcp program in Bx source code from the Regents of the University of California.

I think we're good on this front.

grahamc:
I gave this a try on my network and it worked well. I added new keys and did a send-keys, which also worked. Notes:

1. send-keys with new keys requires a deploy first as the list of keys is cached
2. deployment.keyFile actually is a _file_: directories are not (and have never been, afaik) supported.

PR #1326, opened 2020-04-27
tests: Add functional tests using NixOS in Podman (Docker)

adisbladis:
We want to extend "real world" testing by a lot in NixOps and containers is a way we can do so within the confines of free Github Actions runners.


tests/functional/test_deploy.py
```diff
@@ -0,0 +1,308 @@
...
+    @property
+    def started(self) -> bool:
+        return self.container_id is not None
+
+    def run(self):
+        process = subprocess.run(
+            [
+                "docker",
+                "run",
+                "--privileged",
```
grahamc, line 96: So.

Funny thing about this.

`docker` from `podman` makes `--privileged` not so interesting.

`docker` from `docker` makes `--privileged` .......much more interesting.

signal-2020-04-29-170151

It actually took over my entire system. There are even 2 gettys running: ee the alternating hostnames.

lol

so I think actually we should figure out how to force this to only use podman, and force it to only use podman in user-mode?

tests/functional-1.7/__init__.py
```diff
@@ -0,0 +1,14 @@
+import os
```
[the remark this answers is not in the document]
adisbladis: Let's do that in a follow-up?

tests/functional/test_deploy.py
adisbladis: We're now only calling `podman`. The CI still runs on Docker because we never got rootless Podman to work on Github Actions.

It does so by putting a symlink from `docker` -> `podman` which is a total hack but OK if only in CI.

adisbladis:
This is too complex.. I think we'll just bite the bullet and accept that Github Actions is going to be slow and use the NixOS VM test infra regardless.

PR #1327, opened 2020-04-28
Restore private ipv4 support

adisbladis:
This was accidentally removed in de59e4887b3f51079726a0faf95e390a80c2f3b7

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -362,7 +362,7 @@ class MachineState(nixops.resources.ResourceState):
         return None
 
     @property
-    def private_ipv4(self):
+    def private_ipv4(self) -> Optional[str]:
         return None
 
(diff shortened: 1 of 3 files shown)


grahamc:
I think if we add a type for `private_ipv4` and `public_ipv4`, we'll get the ratchet to pass, and it is sort of in the spirit of the ratchet too. Do you mind doing that?

grahamc:
The ratchet is failing because of this Any:

```
        attrs_per_resource: Dict[str, List[Dict[Tuple[str, ...], Any]]] = {
            m.name: [] for m in active_resources.values()
        }
```

and `attrs_list` has propagated the explicitly annotated Any type. Touching this list twice more is what is causing the problem. It'd be nice to get rid of that Any if we can, but not in this PR. Thanks!

Issue #1320, opened 2020-04-24
Make networks buildable by Hydra


grahamc:
Users would like to be able to pre-build their networks in Hydra as part of CI. This is pretty nice, but Hydra doesn't make this very easy or possible. This could be especially cool for if NixOps were to one day support building netboot images or AMIs automatically.

cleverca22:
both me and @deepfire have done that in the past, and could link some examples

PR #1331, opened 2020-04-28
eval-machine-info.nix: add nixpkgsPath argument.

dhess:
This allows the caller to use a pinned nixpkgs, rather than the `<nixpkgs>` default value. Among other things, this makes it possible to build at least some NixOps networks on a Hydra (or from any Nix expression), which addresses #1320.

### DIFF
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -1,4 +1,5 @@
 { system ? builtins.currentSystem
+, nixpkgsPath ? <nixpkgs>
 , networkExprs
 , checkConfigurationOptions ? true
 , uuid
@@ -7,7 +8,7 @@
 , pluginNixExprs
 }:
 
-with import <nixpkgs> { inherit system; };
+with import nixpkgsPath { inherit system; };
 with lib;
 
 
@@ -52,7 +53,7 @@ rec {
           networks;
       in
       { name = machineName;
-        value = import <nixpkgs/nixos/lib/eval-config.nix> {
+        value = import (nixpkgsPath + "/nixos/lib/eval-config.nix") {
           modules =
             modules ++
             defaults ++


lovesegfault:
It'd be nice if this added docs on how to pin nixpkgs too :)

dhess:
@lovesegfault I think that's out of scope for this change, or even for NixOps in general. Pinning nixpkgs is now a fairly common practice in nixpkgs-based projects, it can be done in a multitude of ways, and it wouldn't be practical to document it everywhere that it's supported :)

My preference these days is to use `niv` for this, and the general method I use to do it can be found here:

https://github.com/hackworthltd/hacknix-lib/blob/7d2cf1abaf8ef6ea30bb888a725123291227424c/nix/default.nix

But that's just one way to do it and it would be a bit much to try to document this in NixOps, in my opinion.

lovesegfault:
@dhess I completely failed to express what I meant over there, my bad!

I meant to ask whether it would make sense to add docs on how to/where to se `nixpkgsPath` when using `nixops`. In other words, I already use `niv`, I have my own `nix/default.nix`, now how do I point `NixOps` to it?

dhess:
@lovesegfault Ahh, I see what you meant. No worries!

It's really quite easy now that @grahamc and @adisbladis have done all this nice refactoring work. Assuming you're using `niv` with a pinned `nixops` and `nixpkgs`, just do something like this:

```
nixpkgsPath = (import sources.nixpkgs {}).path;
nixops = import sources.nixops { inherit nixpkgsPath; };
```

I can see how that would be helpful in the NixOps documentation, but I'm not sure how to express concisely without dragging `niv` into it, as the non-`niv` ways are quite a bit more involved.

lovesegfault:
Maybe something like:

```markdown
# Pinning Nixpkgs

You can pin `nixpkgs` by enforcing `nixpkgsPath` on your call to `nixops`. 
For example, assuming you had `myNixpkgs` pinned you can do 
`callPackage nixops { nixpkgsPath = myNixpkgs; }`

For more information on pinning nixpkgs see [insert helpful link here maybe https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs].
```

lovesegfault:
cc. @cole-h who is good at explaining

cole-h:
"Good" is stretching it. I've slightly improved upon your suggestion with the following:

```markdown
# Pinning Nixpkgs

You can pin the `nixpkgs` that `nixops` will use by setting the `nixpkgsPath`
argument.

Assuming you have `myNixpkgs` pinned (see the [FAQ on pinning
Nixpkgs](https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs) for more information), you
could overlay `nixops` with the following:

```nix
{
	# myNixpkgs = ...
    nixpkgsPath = myNixpkgs.path;
    nixops = callPackage ./path/to/nixops { inherit nixpkgsPath; };
}
whycantiescapethesewithabackslash```
```

I only worry that this is too abstract... Maybe another example following it that takes advantage of `niv`, to put the execution into more concrete terms?

adisbladis:
We have merged support for defining a network in a flake, this will use the flake-pinned nixpkgs version.

Is this good enough for you @dhess?

dhess:
@adisbladis Tough for me to say as I haven't used flakes yet, but I can see where the winds are blowing. Shall I close this, then?

lovesegfault:
I really hope that it is possible to pin nixpkgs without having to use flakes

adisbladis:
I think https://github.com/NixOS/nixops/pull/1369 is a better solution to this problem.

lovesegfault:
Works for me :)

adisbladis:
Closing this in favour of #1369

Issue #1279, opened 2020-04-08
NixOps' host key utility requires knowing an SSH key to remove it


grahamc:
This is not a realistic requirement since we have an IP and don't know what the old state of the system was. We should be able to remove the key with just the IP.

grahamc:
Here is the source of the trouble: https://github.com/NixOS/nixops/blob/master/nixops/known_hosts.py#L11-L52 this code is very complicated. Maybe `ssh-keygen -R` would be a better approach?

PR #1332, opened 2020-04-29
known_hosts: Don't update a key to an empty IP address

jhillyerd:
Remove keys that have an empty IP address instead of trying to re-add
them during the update phase.

Should resolve error seen in
https://github.com/nix-community/nixops-vbox/pull/3#issuecomment-620104958

### DIFF
--- a/nixops/known_hosts.py
+++ b/nixops/known_hosts.py
@@ -66,4 +66,5 @@ def update(prev_address: str, new_address: str, public_host_key: str) -> None:
     # FIXME: this rewrites known_hosts twice.
     if prev_address != new_address:
         remove(prev_address, public_host_key)
-    add(new_address, public_host_key)
+    if new_address is not None:
+        add(new_address, public_host_key)


jhillyerd:
I tested this on nixops-vbox with `tests/functional/single_machine_vbox_base.nix` as well as a dual machine network of my own with create, deploy, stop (and no stop), destroy, and delete.  No issues encountered.

nixops/known_hosts.py
```diff
@@ -66,4 +66,5 @@ def update(prev_address: str, new_address: str, public_host_key: str) -> None:
     # FIXME: this rewrites known_hosts twice.
     if prev_address != new_address:
         remove(prev_address, public_host_key)
-    add(new_address, public_host_key)
+    if new_address is not None:
```
domenkozar, line 69: shouldn't remove also be conditioned to this check?

jhillyerd: Possibly.  I don't know a good way to test changes do this code, so I was not comfortable going beyond a break-fix.

My guess is that none of these were meant to be called with empty IP addresses, and that the Virtualbox plugin is broken somewhere.

grahamc:
I think the better option is to delete the `update` function, and make users call remove and/or add as appropriate. Note this semi-related issue: https://github.com/NixOS/nixops/issues/1279 and with that, I'm inclined to close this PR, unless you wanted to do that here. What do y'all think?

jhillyerd:
Good to close.  👍

PR #1334, opened 2020-05-01
fix op_check

manveru:


grahamc:
I won't block on the ratchet, but if anyone wants to add a commit improving type coverage in this file that would be cool :).

manveru:
@grahamc what is this ratchet thing, and why does it not like the commit?

adisbladis:
@manveru It's to force us to increase type coverage over time, so we measure the coverage before and after the PR.

PR #1337, opened 2020-05-07
ReadTheDocs: fixup and get a successful render

grahamc:
ReadTheDocs uses a very old (1.x) Sphinx, but we're using a 3.x. In order to update it, we have to create a requirements.txt and point RTD to it with a file in the root of the project. A bit ugly, but at least  we can do it! (Though it'd be cool if they supported pyproject / poetry.)

To keep the requirements.txt consistent, CI now verifies that it matches the poetry.lock (and also that poetry.lock is up to date with pyproject.toml.)


grahamc:
I created my own project on RTD to test this, here it is rendered: https://nixops-testing.readthedocs.io/en/fixup-rtd/

Issue #1341, opened 2020-05-12
Use Niv for development shell


adisbladis:
And replace with Flakes when flakes is in Nix.

edolstra:
Why would we use this instead of flakes?

adisbladis:
Because Flakes is not yet released and I don't expect all contributors to run on a branch with flakes enabled for quite some time.

Niv should only be seen as a stop gap solution.

PR #1342, opened 2020-05-12
Use niv to pin nixpkgs channel for nix shell

adisbladis:
This will simplify contributing for people on other channels than unstable.


edolstra:
:-1: on this. We already have a flake branch that we're using in production, let's not complicate things by bringing in another tool.

adisbladis:
Copying my comment from the issue https://github.com/NixOS/nixops/issues/1341#issuecomment-627360762:

Flakes is not yet released and I don't expect all contributors to run on a branch with flakes enabled for quite some time.

Niv should only be seen as a stop gap solution.

gilligan:
my 2cents: `niv` seems like a reasonable approach to me. I don't use an umerged nix branch and i personally wouldn't recommend this to anyone outside of playing around with it.

Let's remind ourselves that we are discouraging Nix users from the new, experimental nix commands. This seems very much at odds with elsewhere suggesting "just build this unreleased branch of nix and use that".

Last but not least. It is trivial to remove `niv` again and use flakes instead once they are released.

edolstra:
Either way you need a non-standard command so I don't see an advantage of one over the other, except that I don't want to endorse niv because it would send the message that we're standardizing on that rather than flakes.

adisbladis:
> Either way you need a non-standard command so I don't see an advantage of one over the other, 

The advantage of using `niv` in this context is that it works with whatever version of Nix you have on your system already.
With flakes we can't expect users to be able to simply invoke `nix-shell` and get a working development environment, at least not yet.

What's the alternative stop-gap to pin nixpkgs until we can rely on flakes?
I'd rather not manually pin a revision with an updater script but I'm OK with that if we need to.

PR #1343, opened 2020-05-13
Add flake8 linter

adisbladis:
This is mainly an addition to avoid import loops by detecting unused imports early and force them to be removed.

Flake8 also detects other things which may be of interest of us such as bad variable naming, cyclomatic complexity (this one is of great interest to me at least).
    
We opt out of a number of checks that conflict with black's auto formatting rules.


grahamc:
Ratchet is whatever, but the requirements.txt for the docs is out of date

grahamc:
Ignoring the ratchet since this is strictly an improvement, and a few deleted imports shouldn't be penalized.

PR #1345, opened 2020-05-15
backends: s/provison/provision/

cole-h:
Sure hope this works. I just ran `sed 's/provison/provision/g' -i **.py` and don't know how to test this.


grahamc:
Thanks!

PR #1346, opened 2020-05-16
Plugin types: Tie a ResourceState to its corresponding ResourceDefinition

grahamc:
Using a protocol here makes mypy much more picky about variable definitions in the base classes (I think this is a good thing.) This helped find some hidden assumptions which are more correctly typed now.

PR #1348, opened 2020-05-18
deployment.py: fixup mistakes in #1346

grahamc:


tewfik-ghariani:
Can we use 
```python
if dep._destroyed_event:
```
instead of 
```python
if dep._destroyed_event is not None:
```

As far as I remember, it improves slightly the performance and surely it's easier to read

grahamc:
I think that is probably fine, what do you think @adisbladis?

adisbladis:
@grahamc Fine by me, since it's `Optional[Event]` it's always truthy when set.

grahamc:
Thanks, @tewfik-ghariani and @adisbladis?

PR #1353, opened 2020-05-25
Add support for pluggable transports

adisbladis:
SSH is not the only way to connect to machines.
I for one am using NixOps to deploy to my local machine and would like to be able to do that without going over SSH.

Another use case this opens up for is one I've been thinking a lot about but is too big for the scope of this PR, namely serial ports to deploy to embedded and air gapped machines that doesn't even have networking.

Additionally this PR contains quite a sizeable refactor to decouple command issuance from SSH which in itself makes all this effort worth it.


nixops/backends/__init__.py
```diff
@@ -80,9 +113,13 @@ class MachineState(
         "hasFastConnection", False, bool
     )
 
-    ssh: nixops.ssh_util.SSH
-    ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
-    _ssh_pinged_this_time: bool = False
+    _transport: Optional[Transport]
+    transport_type: Type[Transport]
+
+    # The attr_proporty name is sshPinged for legacy reasons
```
cole-h, line 119:
```suggestion
    # The attr_property name is sshPinged for legacy reasons
```

nixops/command.py
```diff
@@ -0,0 +1,20 @@
...
+    command: str,
+    user: str,
+    allow_ssh_args: bool,
+    privilege_escalation_command: List[str],
+) -> List[str]:
+    if allow_ssh_args:
+        return shlex.split(command)
+
+    # Don't make assumptions about remote login shell
+    cmd: List[str] = ["bash", "-c", command]
```
pasqui23, line 15:
```suggestion
    cmd: List[str] = ["sh", "-c", command]
```

PR #1355, opened 2020-05-27
Nixops aws compat

adisbladis:
Some changes required for https://github.com/NixOS/nixops-aws/pull/103 to land.

### DIFF
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -33,6 +33,8 @@ from typing import (
     Tuple,
     Union,
     cast,
+    TypeVar,
+    Type,
 )
 import nixops.backends
 import nixops.logger
@@ -53,6 +55,9 @@ class UnknownBackend(Exception):
 
 
 DEBUG = False
+
+
+TypedResource = TypeVar("TypedResource")
 
 
 class Deployment:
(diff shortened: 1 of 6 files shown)


nixops/deployment.py
```diff
@@ -138,22 +143,24 @@ def active_resources(self) -> Dict[str, nixops.resources.GenericResourceState]:
...
-        return res
+        if res.get_type() != type_name:
+            raise Exception(
+                "resource ‘{0}’ is not of type ‘{1}’".format(name, type_name)
+            )
+        return res  # type:ignore
 
-    def get_machine(self, name: str) -> nixops.resources.GenericResourceState:
+    def get_machine(self, name: str, type: Type[TypedResource]) -> TypedResource:
         res = self.active_resources.get(name, None)
```
grahamc, line 158: I wonder if we should have a get_generic_machine/resource, and a get_typed_machine/resource. What do you think?

nixops/diff.py
```diff
@@ -188,7 +188,7 @@ def retrieve_def(d):
                 name = d[4:].split(".")[0]
                 res_type = d.split(".")[1]
                 k = d.split(".")[2] if len(d.split(".")) > 2 else key
-                res = self._depl.get_typed_resource(name, res_type)
+                res = self._depl.get_typed_resource(name, res_type, Any)
```
grahamc, line 191: the _generic method would solve this one a bit nicer I think.

nixops/resources/__init__.py
```diff
@@ -103,13 +102,20 @@ def get_type(cls) -> str:
     _created_event: Optional[Event] = None
     _destroyed_event: Optional[Event] = None
     _errored: Optional[bool] = None
+
+    # While this looks like a rookie mistake where the list is going  get shared
+    # across all class instances it's not... It's working around a Mypy crash.
+    #
+    # We're overriding this value in __init__.
+    # It's safe despite there being a shared list on the class level
```
grahamc, line 110: Can you include a link to a bug ref, and/or a version number, or some more specific information for future spelunkers to explain exactly what code is related to this, and what they can delete once it is fixed?

nixops/util.py
```diff
@@ -154,6 +154,19 @@ def _transform_value(key: Any, value: Any) -> Any:
...
+                for v in value:
+                    for subann in ann.__args__:
+                        if inspect.isclass(subann) and issubclass(
+                            subann, ImmutableValidatedObject
+                        ):
+                            new_value.append(subann(**v))
+                        else:
+                            new_value.append(v)
+                value = tuple(new_value)
+
```
grahamc, line 169: How about a test for this?

nixops/deployment.py
adisbladis: I was considering returning to
```python 
def get_machine(self, name: str) -> nixops.resources.GenericResourceState:
```
and let the caller use [cast](https://docs.python.org/3/library/typing.html#typing.cast) instead.

This results in exactly the same type safety as 
```python 
def get_machine(self, name: str, type: Type[TypedResource]) -> TypedResource:
```
with only some slight inconvenience for the caller.

Wdyt?

grahamc: I like the second option much better. It is really awkward to use in other libraries: https://github.com/input-output-hk/nixops-packet/pull/21/files#diff-78f3655b013910dfdb430d53dae85085R11-R18 this feels like something that should definitely be implemented in core, to provide a safe API.

PR #1358, opened 2020-06-05
Add Python API reference docs

adisbladis:


ci/lint-docs.sh
```diff
@@ -3,6 +3,6 @@
 
 set -eux
 
-git ls-files | xargs codespell
+git ls-files | xargs codespell -L keypair,iam,hda
```
nlewo, line 6: Is `keypair` keyword still needed here since there are fixed?

ci/lint-docs.sh
```diff
@@ -3,6 +3,6 @@
...
 sphinx-build -M clean doc/ doc/_build
-sphinx-build -nW doc/ doc/_build
+sphinx-build -n doc/ doc/_build
```
nlewo, line 8: Do we really need to remove `W`?

ci/lint-docs.sh
adisbladis: Yes, the nix expressions still says `keyPair`.

ci/lint-docs.sh
adisbladis: Yes :/
We have circular imports guarded by `typing.TYPE_CHECKING`. That's not a problem for Mypy but it is a problem for Sphinx.

nlewo: Ok :(

PR #1359, opened 2020-06-05
docs: Add custom rst directive to include plugin documentation

adisbladis:


doc/index.rst
```diff
@@ -27,6 +27,8 @@ NixOps
    release-notes/1.x
 
 
+.. nixops_plugins_doc::
```
FRidh, line 30: What is the purpose of the directive?

adisbladis: To dynamically (at build time) include docs from all plugins in the closure.

adisbladis: It's possible that there is a better way to accomplish this. I'm not really too familiar with Sphinx.

PR #1202, opened 2019-10-04
[WIP] Flake support

edolstra:
This PR adds flake support to NixOps, meaning that you can do things like
```
# nixops create -d hydra-ec2-demo --flake github:edolstra/hydra-ec2-demo
# nixops deploy
```
This will create an [EC2 instance running Hydra](https://github.com/edolstra/hydra-ec2-demo/blob/master/flake.nix), using exactly the configuration locked by [`flake.lock`](https://github.com/edolstra/hydra-ec2-demo/blob/master/flake.lock).

Currently evaluation is not as hermetic as we'd like since the NixOps-specific NixOS modules are provided impurely. Ideally, the flake would lock a particular version of the NixOps flake as input.

Likewise, NixOps needs to pass in the generated file `physical.nix` as an impure input. This is something of an inherent issue with the NixOps model. Maybe for flakes in local Git repositories, `physical.nix` could be auto-committed.

This PR also converts `release.nix` etc. to `flake.nix`. `update-all` (the script for fetching NixOps plugins) is no longer needed since `nix flake update` can do the same.

PR #1364, opened 2020-06-09
Introduce flake.nix!

nlewo:
This project can however still be built with the stable Nix since they
are a compatibility function in the `default.nix` file.

Note this is the first time I play with the "flakes"...


edolstra:
Note that there already is a branch that adds flake support, see https://github.com/NixOS/nixops/pull/1202.

PR #1366, opened 2020-06-11
Use flake check for the documentation linting test

nlewo:
The goal is to easily run the documentation CI tests locally. 

- It moves the documentation linting test as a `flake check` attribute.
- The script `lint-docs` is now part of the shell environment.

The idea would be to move more CI scripts to `nix flake check`: it is easier to discover them and to reproduce them.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,6 +7,23 @@
 
   outputs = { self, nixpkgs, utils }: utils.lib.eachDefaultSystem (system: let
     pkgs = import nixpkgs { inherit system; };
+
+    sphinx = pkgs.python3.withPackages(p: [ p.sphinx ]);
+    linters.doc = pkgs.writers.writeBashBin "lint-docs" ''
+      set -eux
+      # When running it in the Nix sandbox, there is no git repository
+      # but sources are filtered.
+      if [ -d .git ];
+      then
+          FILES=$(${pkgs.git}/bin/git ls-files)
+      else
+          FILES=$(find .)
+      fi
+      echo "$FILES" | xargs ${pkgs.codespell}/bin/codespell -L keypair,iam,hda
+      ${sphinx}/bin/sphinx-build -M clean doc/ doc/_build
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -7,6 +7,28 @@
 
   outputs = { self, nixpkgs, utils }: utils.lib.eachDefaultSystem (system: let
     pkgs = import nixpkgs { inherit system; };
+    # TODO: need to be moved as flake input once this repository is a flake
+    gitignore = import (builtins.fetchTarball {
```
FRidh, line 11: I don't think we can use `fetchTarball` on Hydra.

flake.nix
```diff
@@ -7,6 +7,28 @@
...
+      url = https://github.com/hercules-ci/gitignore/archive/647d0821b590ee96056f4593640534542d8700e5.tar.gz;
+      sha256 = "0ks37vclz2jww9q0fvkk9jyhscw0ial8yx2fpakra994dm12yy1d";
+    }) { inherit (pkgs) lib; };
+
+    sphinx = pkgs.python3.withPackages(p: [ p.sphinx ]);
```
adisbladis, line 16: Use the environment provided by `mkPoetryEnv`, we don't want to be using multiple different versions of `sphinx`.
https://github.com/NixOS/nixops/blob/master/flake.nix#L13-L15

flake.nix
nlewo: This is correct (excepting if `allowed_uris` is set). 

I've planned to send a PR to the `hercules-ci/gitignore` to add the flake support.

Is it a short term issue? 
In this case, we can wait until the `gitignore` flake support is merged to merge this PR: I could then put this lib as a flake input.

nlewo: I finally switch to the `poetry2nix.cleanPythonSources` which works with the restricted evaluation.

nlewo:
@adisbladis I think I've addressed all requests. Could you take a look?

Issue #1370, opened 2020-06-13
Cannot use with fetchTarball after recent flakes migration due to builtins.fetchGit


cole-h:
The following `shell.nix` does not work:

```nix
let
  pkgs = import <nixpkgs> {};
  nixops = import (fetchTarball "https://github.com/NixOS/nixops/archive/master.tar.gz");
in
pkgs.mkShell {
  buildInputs = [ nixops ];
}
```

This is because of https://github.com/NixOS/nixops/blob/5a36c7b77c96ea0f843aed9a2e1e0ea451635886/default.nix#L5 -- there is no `.git` served with the master tarball, so no `git` commands will work.

Output of `nix-shell --show-trace shell.nix`:

```
fatal: '/nix/store/016174sf6gbrw96szs3f1nf6cgiz49xv-source' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
error: while evaluating the attribute 'buildInputs' of the derivation 'nix-shell' at /nix/var/nix/profiles/per-user/root/channels/nixos/pkgs/build-support/mkshell/default.nix:28:3:
while evaluating 'getOutput' at /nix/var/nix/profiles/per-user/root/channels/nixos/lib/attrsets.nix:464:23, called from undefined position:
while evaluating anonymous function at /nix/var/nix/profiles/per-user/root/channels/nixos/pkgs/stdenv/generic/make-derivation.nix:143:17, called from undefined position:
while evaluating the file '/nix/store/016174sf6gbrw96szs3f1nf6cgiz49xv-source/default.nix':
while evaluating the attribute 'defaultNix' at /nix/store/y91vx4bq08c04r1chnc97jidhxnzhsjz-source/default.nix:114:5:
program 'git' failed with exit code 128
```

PR #1371, opened 2020-06-13
{default,shell}.nix: don't use builtins.fetchGit

cole-h:
Otherwise, this would prevent one from using `import (fetchTarball
"https://github.com/NixOS/nixops/archive/master.tar.gz")` to run NixOps
from the master branch, for example. If you try to do so, you would see
some output like the following:

    fatal: '/nix/store/nysvzmqi11zfchympbksb8wcb2dxw6yz-source' does not appear to be a git repository
    fatal: Could not read from remote repository.

---

Fixes https://github.com/NixOS/nixops/issues/1370.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,6 @@
 (import (
   fetchTarball {
-    url = https://github.com/edolstra/flake-compat/archive/f012cc5092f01fc67d5cd6f04999f964c3e05cbf.tar.gz;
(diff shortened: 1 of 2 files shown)


cole-h:
This bumps flake-compat to include https://github.com/edolstra/flake-compat/pull/4 (as well as support for versions 6 and 7 of the lockfile) -- now, `builtins.fetchGit` will be used automatically in the case that `.git` is present.

PR #1372, opened 2020-06-15
Use flake-compat without fetchGit

adisbladis:
An alternative approach to #1371 that doesn't force you to rebuild everything on every change.

This fixes 2 different issues:
1. Github actions Checkout does a shallow clone by default.
   We want to use a shallow clone but force a deep clone because
   builtins.fetchGit doesn't seem to work on shallow clones.

2. It doesn't work when this repo is a store path and .git is removed.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,2 @@
-(import (
-  fetchTarball {
-    url = https://github.com/edolstra/flake-compat/archive/f012cc5092f01fc67d5cd6f04999f964c3e05cbf.tar.gz;
-    sha256 = "1n8q7v7alq802kl3b6zan6v27whi2ppbnlv8df6cz64vqf658ija"; }) {
-      src = builtins.fetchGit ./.;
-}).defaultNix
+{ pkgs ? import <nixpkgs> {} }:
+(import ./flake-compat.nix { inherit pkgs; }).defaultNix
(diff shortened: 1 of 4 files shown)


default.nix
```diff
@@ -1,6 +1,2 @@
-(import (
-  fetchTarball {
-    url = https://github.com/edolstra/flake-compat/archive/f012cc5092f01fc67d5cd6f04999f964c3e05cbf.tar.gz;
-    sha256 = "1n8q7v7alq802kl3b6zan6v27whi2ppbnlv8df6cz64vqf658ija"; }) {
-      src = builtins.fetchGit ./.;
-}).defaultNix
+{ pkgs ? import <nixpkgs> {} }:
```
edolstra, line 1: Making `default.nix` and `shell.nix` depend on `<nixpkgs>` is a bad idea. 1) It's inefficient, since now evaluation requires at least *2* versions of Nixpkgs. 2) It goes against the whole point of flakes, which is hermetic evaluation. It reintroduces a dependency on `$NIX_PATH` which is what we want to get rid of.

edolstra:
I don't understand the motivation. Why would you need to rebuild everything on every change? That's what nix-shell is for right?

edolstra:
I think this would be better addressed in flake-compat, where we could have some logic to check whether `src` is a git tree or not. The `default.nix` and `shell.nix` files that call flake-compat should have very little logic themselves.

nlewo:
I've submitted a PR to use `fetchGit` if src is a git tree in flake-compat:  https://github.com/edolstra/flake-compat/pull/4

Issue #1375, opened 2020-06-20
Too many nixops SSH "master" sockets


grahamc:
While waiting for SSH to be available, I noticed a ton of password prompts. Way too many, in fact :).

```
[grahamc@Petunia:~]$ pgrep -af ssh
435 ssh -x root@147.75.192.191 -S /run/user/1000/nixops-ssh-tmp5ufnlyqo/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -p 22 -i /run/user/1000/nixops-tmpq4iwawkj/id_nixops-check2 -o StrictHostKeyChecking=accept-new
579 ssh -x root@147.75.192.191 -S /run/user/1000/nixops-ssh-tmp_y3sdyy4/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -p 22 -i /run/user/1000/nixops-tmpq4iwawkj/id_nixops-check2 -o StrictHostKeyChecking=accept-new
650 ssh -x root@147.75.192.191 -S /run/user/1000/nixops-ssh-tmpqjtkf914/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -p 22 -i /run/user/1000/nixops-tmpq4iwawkj/id_nixops-check2 -o StrictHostKeyChecking=accept-new
761 ssh -x root@147.75.192.191 -S /run/user/1000/nixops-ssh-tmpwpuqjc4m/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -p 22 -i /run/user/1000/nixops-tmpq4iwawkj/id_nixops-check2 -o StrictHostKeyChecking=accept-new
831 ssh -x root@147.75.192.191 -S /run/user/1000/nixops-ssh-tmpoa8ixxyc/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -p 22 -i /run/user/1000/nixops-tmpq4iwawkj/id_nixops-check2 -o StrictHostKeyChecking=accept-new
[...]

PR #1380, opened 2020-07-05
Nixops ssh conn quantity and stderr fixup

johnalotoski:
* Fixes issue #1375, too many ssh conns, as observed in testing with nixops-packet plugin https://github.com/input-output-hk/nixops-packet/pull/21.  Not yet tested with other plugins.
* The recent tcp port probe to ssh probe change (362d049) introduced threaded workers which persist in their ssh conn task despite event blocking timing out.  Adding an ssh connecttimeout to match the event block timeout resolves the buildup of worker threads.
* An option for suppression of ssh master stderr was added to enable ssh status probes to avoid outputting large quantities of expected stderr messages to the user.

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -280,6 +280,7 @@ class MachineState(
                     timeout=1,
                     logged=False,
                     connection_tries=1,
+                    ssh_quiet=True,
                 )
             except Exception:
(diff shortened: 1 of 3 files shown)


adisbladis:
Great stuff, works nicely. Thanks!

PR #1382, opened 2020-07-06
SSH: don't force port 22 if none is given

### LINKED ISSUE #1053: SSH port in ~/.ssh/config is ignored
I have a machine available on my local network  as `foo.home`, with SSH on port 22. So I can do `deployment = { targetHost = "foo.home"; targetEnv = "none"; }`. This works as expected.

If I add this section to `~/.ssh/config`:

```
Host foo.home
  HostName foo.example.org
  Port 10022
```

`nixops deploy` no longer works, even though I can still `ssh foo.home` as I normally would. NixOps is ignoring the port set in my SSH config, which is annoying as I now have to maintain this port in two places. Ideally it should take the default port used by the SSH client if not configured otherwise.

wmertens:
This way, the port specified in the SSH config can be used.

Note, this is completely untested, I just wanted to write this down to see what it would entail to fix #1053

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -84,7 +84,7 @@ class MachineState(
     ssh: nixops.ssh_util.SSH
     ssh_pinged: bool = nixops.util.attr_property("sshPinged", False, bool)
     _ssh_pinged_this_time: bool = False
-    ssh_port: int = nixops.util.attr_property("targetPort", 22, int)
+    ssh_port: int = nixops.util.attr_property("targetPort", None, int)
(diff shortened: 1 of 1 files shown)


wmertens:
:tada: I can confirm this works.

adisbladis:
I am a bit torn about this PR.

I would consider making NixOps completely disregard `~./config/ssh` unless opted in to as it's a configuration file not managed in the NixOps network.

For now I'll just merge this as-is. It does no harm and I'm not entirely convinced my reasoning is correct.

PR #1383, opened 2020-07-06
Define plugins hooks as classes

adisbladis:
This is a better way to deal with plugins as we can statically type each individual function, something not possible with pluggy hooks.

### DIFF
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -107,11 +107,15 @@ Important Notes
 
    .. code-block:: python
 
-     @nixops.plugins.hookimpl
-     def nixexprs():
-         return [
-             os.path.dirname(os.path.abspath(__file__)) + "/nix"
-         ]
+     from nixops.plugins import Plugin
+
+     class NeatCloudPlugin(Plugin):
+
+         @staticmethod
+         def nixexprs():
+             return [
+                 os.path.dirname(os.path.abspath(__file__)) + "/nix"
+             ]
 
 5. Resource subclasses must now work with Python objects instead of XML
 
@@ -261,7 +265,7 @@ return it in the appropriate plugin hook:
   from nixops.plugins import Plugin
   import nixops.plugins
 
-  class MyPlugin(Plugin):
+  class NeatCloudPlugin(Plugin):
 
       @staticmethod
       def nixexprs():
@@ -269,7 +273,7 @@ return it in the appropriate plugin hook:
 
   @nixops.plugins.hookimpl
   def plugin():
-      return MyPlugin()
+      return NeatCloudPlugin()
 
 
 
--- a/doc/plugins/authoring.rst
+++ b/doc/plugins/authoring.rst
@@ -250,6 +250,29 @@ discover and load plugins. The glue which hooks things together is in
 NixOps implements a handful of hooks which your plugin can integrate
 with. See ``nixops/plugins/hookspec.py`` for a complete list.
 
+Defining a plugin
+-----------------
+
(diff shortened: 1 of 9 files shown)


flake.nix
```diff
@@ -1,7 +1,7 @@
 {
   description = "NixOps: a tool for deploying to [NixOS](https://nixos.org) machines in a network or the cloud";
 
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs-channels/nixos-unstable";
```
cole-h, line 4: Why `nixpkgs-channels`? I thought this was deprecated and added to `nixpkgs` proper long ago.

adisbladis: Force of habit.

doc/plugins/authoring.rst
```diff
@@ -250,6 +250,29 @@ discover and load plugins. The glue which hooks things together is in
...
+.. code-block:: python
+
+  from nixops.plugins import Plugin
+  import nixops.plugins
+
+  class MyPlugin(Plugin):
+
+      @staticmethod
+      def nixexprs():
+          return [ ]  # List of paths to nix expressions
```
grahamc, line 268: The doc above suggests changing some code to this format:

https://github.com/NixOS/nixops/blame/881c0b4baa90a57aa1be706beb1e897c6e8732bc/doc/plugins/authoring.rst#L94-L104

I think we should update those examples above, and then maybe include an example here of how to include a file.

Also, can we include a copy of the entire interface here somehow?

nixops/deployment.py
```diff
@@ -613,6 +616,14 @@ def get_physical_spec(self) -> Any:
...
         }
 
+        for plugin in get_plugins():
+            deployment_hooks = plugin.deployment_hooks()
+            if not deployment_hooks:
+                continue
+
+            for name, attrs in deployment_hooks.physical_spec(self).items():
+                attrs_per_resource[name].extend(attrs)
+
```
grahamc, line 626: I think we should not pepper too much of the plugin specific code here, and instead have something like:

```
plugin_manager.physical_spec(self)
```

to handle the iterating, skipping, etc. I'm not sure how to handle the extension of the attrs_per_resource. If we pass it in, it'll get modified, right?
```

File: nixops/deployment.py [hunk 4]
```diff
@@ -679,7 +690,7 @@ def do_machine(m: nixops.backends.GenericMachineState) -> None:
             do_machine(m)
 
         def emit_resource(r: nixops.resources.GenericResourceState) -> Any:
-            config = []
+            config: List[Dict[Tuple[str, ...], Any]] = []
```
grahamc, line 693: Let's give this type a name?

nixops/deployment.py
```diff
@@ -747,15 +758,19 @@ def emit_resource(r: nixops.resources.GenericResourceState) -> Any:
...
 
-        return (
+        v = (
             py2nix(
                 reduce(
                     nixmerge, [emit_resource(r) for r in active_resources.values()], {}
                 )
             )
-            + "\n"
+            # + "\n"
```
grahamc, line 767:
```suggestion
```

(but why?)

nixops/deployment.py
```diff
@@ -747,15 +758,19 @@ def emit_resource(r: nixops.resources.GenericResourceState) -> Any:
...
         )
 
+        # raise ValueError(v)
```
grahamc, line 770:
```suggestion
```

(but why again?)

PR #1386, opened 2020-07-10
resources: Improve error messages when no `config` annotation exists or type is wrong

adisbladis:


nixops/resources/__init__.py
```diff
@@ -42,6 +42,17 @@ def get_resource_type(cls):
 
     def __init__(self, name: str, config: ResourceEval):
         config_type = self.__annotations__.get("config", ResourceOptions)
+
+        if isinstance(config_type, str):
+            if config_type == "ResourceOptions":
+                raise TypeError(
+                    f'{self.__class__} is missing a "config" attribute, for example: `config: nixops.resources.ResourceOptions`, see https://nixops.readthedocs.io/en/latest/plugins/authoring.html'
```
cole-h, line 49: Instead of "for example," maybe we should link to an anchor on the readthedocs page that explains the existence/necessity/usage of `config: nixops.resources.ResourceOptions`. Then this could become:

```suggestion
                    f'{self.__class__} is missing a "config" attribute, see: https://nixops.readthedocs.io/en/latest/plugins/authoring.html#some-anchor-with-info'
```

nixops/resources/__init__.py
```diff
@@ -42,6 +42,17 @@ def get_resource_type(cls):
...
+                )
+            else:
+                raise TypeError(
+                    f"{self.__class__}.config is not allowed to be a string"
```
grahamc, line 53:
```suggestion
                    f"{self.__class__}.config's type annotation is not allowed to be a string, see: https://nixops.readthedocs.io/en/latest/plugins/authoring.html"
```

PR #1388, opened 2020-07-22
ResourceState.definition_type: introduce for run-time accessible definition type validation

grahamc:
This is paired with https://github.com/NixOS/nixops-aws/pull/108. Adding these types here showed zillions of type errors in the nixops-aws plugin. #108 fixes a zillion issues in the types, which makes it actually work. :)

### DIFF
--- a/nixops/resources/__init__.py
+++ b/nixops/resources/__init__.py
@@ -4,8 +4,8 @@ from __future__ import annotations
 import re
 import nixops.util
 from threading import Event
-from typing import List, Optional, Dict, Any, TypeVar, Union, TYPE_CHECKING
-from nixops.monkey import Protocol
+from typing import List, Optional, Dict, Any, Type, TypeVar, Union, TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


nixops/resources/__init__.py
```diff
@@ -70,14 +70,15 @@ def show_type(self) -> str:
         return self.get_type()
 
 
-ResourceDefinitionType = TypeVar(
-    "ResourceDefinitionType", bound="ResourceDefinition", contravariant=True
-)
+ResourceDefinitionType = TypeVar("ResourceDefinitionType", bound="ResourceDefinition")
```
grahamc, line 73: I think this is nice, actually, since we really did want it to be invariant I think.

grahamc:
Ratchet going down was just about unavoidable in this case, so I think it is good to go!

andir:
I think this is in a good enough state to merge it. I've been able to deploy to AWS (with fixes in the AWS plugin) and to libvirt (without any modifications on the code) with this nixops branch.

PR #1393, opened 2020-08-14
Fix diff evaluation logic

lukebfox:
Fix logic preventing changes in configurations of deployed resources from being picked up by the diff engine planner. Also, when assembling the key list to iterate over during diff evaluation, use sets to properly union the state and config keys.


adisbladis:
Thank you :)

PR #1401, opened 2020-09-12
Fix legacy 1.7 build [with potential downsides]

allgreed:
There was some incompatibilities that prevented the build from
completing, I've patched them to make it work, commenting out a few
lines of code (marked in README) - this is likely break GCE and Azure
potentially, but I haven't tested. Works on existing (pre 2.0) nixops
project out of the box (with DigitalOcean as the cloud of choice).

But I guess that partially working legacy nixops is better than one that cannot be built

### DIFF
--- a/nix/azure.nix
+++ b/nix/azure.nix
@@ -390,7 +390,7 @@ in
     };
 
     #fileSystems = mkOption {
-    #  type = with types; loaOf (submodule fileSystemsOptions);
+    #  type = with types; attrsOf (submodule fileSystemsOptions);
     #};
 
   };
--- a/nix/azure.nix
+++ b/nix/azure.nix
@@ -389,9 +389,9 @@ in
 
     };
 
-    fileSystems = mkOption {
-      type = with types; loaOf (submodule fileSystemsOptions);
-    };
+    #fileSystems = mkOption {
+    #  type = with types; loaOf (submodule fileSystemsOptions);
+    #};
 
   };
 
--- a/nix/gce.nix
+++ b/nix/gce.nix
@@ -402,7 +402,7 @@ in
     };
 
     #fileSystems = mkOption {
-    #  type = with types; loaOf (submodule fileSystemsOptions);
+    #  type = with types; attrsOf (submodule fileSystemsOptions);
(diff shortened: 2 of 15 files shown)


nix/azure.nix
```diff
@@ -389,9 +389,9 @@ in
 
     };
 
-    fileSystems = mkOption {
-      type = with types; loaOf (submodule fileSystemsOptions);
-    };
+    #fileSystems = mkOption {
+    #  type = with types; loaOf (submodule fileSystemsOptions);
```
cole-h, line 393: `loaOf` is an alias to `attrsOf` now, so why not just change the type to `attrsOf` proper and be done with it? Doesn't seem necessary to comment it out and break things.

nix/gce.nix
```diff
@@ -401,9 +401,9 @@ in
 
     };
 
-    fileSystems = mkOption {
-      type = with types; loaOf (submodule fileSystemsOptions);
-    };
+    #fileSystems = mkOption {
+    #  type = with types; loaOf (submodule fileSystemsOptions);
```
cole-h, line 405: Same.

cole-h (changes requested):

allgreed:
@cole-h I've changed the `loaOf` as per your suggestion (in the whole project), however then the entire thing breaks:
```
error: The option `fileSystems.<name>.device' defined in `[ path to repo ]/nixops/nix/gce.nix' does not exist.
```

also: any idea what to put instead of `type = types.listOf types.string;`? This one seems to be deprecated as well

mystfox:
There's not a one-size-fits all replacement, https://github.com/NixOS/nixpkgs/pull/66346 has a summary of the options.

allgreed:
@mystfox thanks for the suggestions, all done [and Travis is now passing <3]. Previously I was just trying to see if it will work ;)

Unfortunately I have no idea how to fix the `fileSystems` issue with azure and gce

adisbladis:
Closing this as it's very likely stale and 1.7 is not seeing any development.

Feel free to reopen if it's still relevant.

PR #1405, opened 2020-09-22
nixops copy-closure: use the copy mechanism as exposed by the machine

andir:
For one this behaves different from the closure copying that happends
during deployment and it also makes use of unstable nix features that
might change without any notice.


grahamc:
:partying_face: thanks!

PR #1407, opened 2020-09-23
Document NixOS module system options in generated sphinx docs

andir:
This is my current WIP for generating module option docs as part of the (new) regular docs build. It is yet unclear how we handle plugins, how readthedocs.org will actually `nix-build` the docs etc..


andir:
A rendered example of the NixOS options can be seen here: https://s.rammhold.de/nixops-doc-preview/manual/options.html

Issue #1418, opened 2020-11-04
NixOps can't deploy to a different system type anymore


regnat:
I have a NixOps deployment with several `x86_64-linux` machines and a `aarch64-linux`.
With NixOps 1.7 I could just set `nixpkgs.localSystem` for the aarch64 machine and everything was working fine.

This doesn't work anymore with NixOps 2.0 (I guess for the same reasons than https://github.com/NixOS/nixops/issues/1389), and there's (I think) no way anymore to specify that a machine should have a different `system` than the local one.

The system is set to `builtins.localSystem` in https://github.com/NixOS/nixops/blob/c9fdc936ba7066564993b6b34557c2f0f77f43ab/nix/eval-machine-info.nix#L1 and there's no way to override it (neither per-machine nor globally)

maksar:
There is a trick (thanks to https://github.com/NixOS/nixops/blob/master/nix/eval-machine-info.nix#L51) to deploy from, lets say x86_64-darwin to x86_64-linux:
```
{
  network = {
    description = "...";
    enableRollback = true;
    nixpkgs = (builtins.getFlake "github:NixOS/nixpkgs/release-20.09").legacyPackages."x86_64-linux";
  };
...
```
Or, if you don't want to pin your nixpkgs:
```
{
  network = {
    description = "...";
    enableRollback = true;
    nixpkgs = import <nixpkgs> { system = "x86_64-linux"; };
  };
...
```

regnat:
Thanks for the nice trix @maksar. Unfortunately I don't think that can work for mixing different system tyes in the same deployment :(

aaronjanse:
Sadly that workaround also doesn't work for nixops flakes

roberth:
I'd like to re-wire the network expression using the module system, so we can have proper network-wide options. If I do that, solving this issue is a matter of paying attention.
Using freeform modules I'm confident that we can even retain compatibility with the current format that mixes machines and other options.
Any objections @adisbladis @grahamc @AmineChikhaoui ?

regnat:
I didn't notice that gh didn't mention it there, but this specific issue should actually be closed by https://github.com/NixOS/nixops/pull/1422 that got merged a couple weeks ago

adisbladis:
@roberth I've wanted to do the same so absolutely no objections from me :)

PR #1422, opened 2020-12-02
Don't hijack `pkgs` in the nixos config

### LINKED ISSUE #1389: NixOps machines can no longer use `nixpkgs` options or specify overlays after #1369
After #1369 overlays are ignored, and also user overlays are always loaded impurely:

See:
* https://github.com/NixOS/nixops/pull/1369/files#diff-7f8400a57a1b63ade4fe16a6154cc092R89
* https://github.com/adisbladis/nixops/blob/9294637ea838d2d9fc27035cadbe0b257ca2ddfb/nix/eval-machine-info.nix#L51

regnat:
Don't pass `pkgs` to `evalConfig` as

1. This prevents most `nixpkgs.*` option from working
2. This doesn't add any encapsulation guaranty as by default NixOS gets nixpkgs from the path of the toplevel module − which is the same as what NixOps does (see https://github.com/NixOS/nixpkgs/blob/4d43de3/nixos/modules/misc/nixpkgs.nix#L58-L60)

Hopefully fixes #1389 and #1418

### DIFF
--- a/nix/eval-machine-info.nix
+++ b/nix/eval-machine-info.nix
@@ -86,7 +86,6 @@ in rec {
       in
       { name = machineName;
         value = evalConfig {
-          inherit pkgs;
           modules =
             modules ++
             defaults ++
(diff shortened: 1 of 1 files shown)


aaronjanse:
I can confirm that this PR allowed me to deploy nixops to an aarch64 Raspberry Pi 4

PR #1423, opened 2020-12-12
Bump pretty tables version to 2.0.0

ixxie:
nixops package is broken; attempts to install it yields the error:

ERROR: Package 'prettytable' requires a different Python: 2.7.18
not in '>=3.6'

This bumps the version of prettytables to 2.0.0 which resolves the
issue.


ixxie:
After further review, I'm getting the feeling the issue isn't caused by this version.

PR #1424, opened 2020-12-19
upgrading install-nix-action to fix build

tewfik-ghariani:
All nixops as well as nixops-plugins' builds are failing at the 'nix installation' part

```Error: Unable to process command '::add-path::/nix/var/nix/profiles/per-user/runner/profile/bin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
Error: Unable to process command '::add-path::/nix/var/nix/profiles/default/bin' successfully.
Error: The `add-path` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
Error: Unable to process command '::set-env name=NIX_PATH::/nix/var/nix/profiles/per-user/root/channels' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

Ref : https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

It was fixed per https://github.com/cachix/install-nix-action/pull/52 so we need to upgrade to v12

cc @adisbladis

PR #1427, opened 2020-12-24
Fix nixops check: Ignore non-existing home.mount

roberth:
Exactly like `tmp.mount`.

quote
```
                # Currently in systemd, failed mounts enter the
                # "inactive" rather than "failed" state.  So check for
                # that.  Hack: ignore special filesystems like
                # /sys/kernel/config and /tmp. Systemd tries to mount these
                # even when they don't exist.
```


roberth:
Please merge https://github.com/NixOS/nixops/pull/1424 to fix the ci.

roberth:
@grahamc could you review this?

adisbladis:
Rebased and merged in #1435

PR #1438, opened 2021-02-10
fix #1437

### LINKED ISSUE #1437: Diff engine planner buggy for list definitions
Resources using the diff engine manage their state with a `StateDict` object. Setting and getting state data via this class causes the diff planner to generate false positives when computing a diff, in some cases. 

When storing data in the state (really just sqlite3) python objects such as lists and dicts which cannot be inserted into a database are converted into a JSON encoded string which is stored instead. Then, when getting data out, the JSON string is decoded back into python objects. Neat, except that there an inconsistency in the conversion process regarding tuples. Python's `json` reference gives:
- [Encoding](https://docs.python.org/3/library/json.html#py-to-json-table): List / Tuple -> JSON array
- [Decoding](https://docs.python.org/3/library/json.html#json-to-py-table): JSON array -> List

The diff planner compares *definitions* with *state*. Definitions of `types.list` in nix are parsed into `typing.Sequence` in python. Such definitions are encoded to JSON array when storing in state, and decoded to `typing.List` when the state is queried for the comparison. This means the diff planner will always compare tuples with lists and trigger the handler for such definitions.

lukebfox:
fixes #1437:
When querying state, casting list typed values into tuples makes diff comparisons type-consistent. 

The JSON encoder may squish both lists and tuples into JSON arrays, however this information loss doesn't affect us because we never parse nix definitions in as lists, only as tuples. As such, decoding all JSON arrays to tuples is safe, and solves our problem.

### DIFF
--- a/nixops/state.py
+++ b/nixops/state.py
@@ -46,7 +46,10 @@ class StateDict(collections.MutableMapping):
             row: Tuple[str] = c.fetchone()
             if row is not None:
                 try:
-                    return json.loads(row[0])
+                    v = json.loads(row[0])
+                    if isinstance(v, list):
+                        v = tuple(v)
+                    return v
                 except ValueError:
                     return row[0]
             raise KeyError("couldn't find key {} in the state file".format(key))


adisbladis:
Thanks!

Could you improve your commit message? The first line should read something descriptive about either why or what the change is.

Maybe you could also do some mashup of the issue report in #1437 (which is an excellent issue report, kudos) and this PR body?

lukebfox:
Thanks! Hope this is alright?

adisbladis:
Looks great!

PR #1450, opened 2021-03-29
Mention keyFile in the 'Managing keys' section

jleeuwes:
The 'Managing keys' section suggested that `text` is required, and did not mention that `keyFile` can be used instead.

I simply added a `keyFile` mention without further changing of the example, mainly because I don't know how to best indicate the two choices in the code example.


cab404:
@jleeuwes Please, also add that `keyCommand` exists.

jleeuwes:
@cab404 Done. I also changed another file that seems to be sort of the same file, to keep it in sync.

PR #1464, opened 2021-08-06
Always provide our known_hosts in addition to user known_hosts

roberth:
Solves one of the problems from https://github.com/NixOS/nixops/pull/1264#issuecomment-889884626

It also solves a problem where multiple deployments to distinct private networks fight over the key for the same ip that occurs in both.

> `nixops import --include-keys`-like functionality doesn't seem to be included.

Prevously, you'd
1. create a deployment
2. upload the state
3. download the deployment state on a fresh machine (eg colleague)
4. nixops ssh doesn't know the keys
5. non-interactive deployments fail

With this pr, we don't have to rely on user configuration to include
known_hosts entries for the deployments.

It makes `nixops import --include-keys` unnecessary, unless you
use those entries outside of nixops.

Since recently we can get our deployment state from remote storage
backends, but we didn't have a way to get configure the known_hosts
yet. This is now largely unnecessary.

This functionality requires some cooperation from the plugins. For
instance, here's what ec2 needs to do: (pun intended)
    
    +    def get_ssh_host_keys(self):
    +        return self.private_ipv4 + " " + self.public_host_key + "\n" + self.public_ipv4 + " " + self.public_host_key + "\n"

(https://github.com/NixOS/nixops-aws/pull/141)

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -469,22 +469,17 @@ class MachineState(
         return None
 
     def _get_ssh_ambient_options(self) -> Dict[str, str]:
-        proc = subprocess.Popen(
-            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE
-        )
-        opts: Dict[str, str] = {}
-        if proc.stdout is None:  # mostly for mypy; Popen won't do this to us.
+        with subprocess.Popen(
+            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE, text=True
+        ) as proc:
+            assert proc.stdout is not None
+            opts: Dict[str, str] = {}
+            for line in proc.stdout:
+                s = line.rstrip("\r\n").split(" ", 1)
+                if len(s) == 2:
+                    opts[s[0].lower()] = s[1]
+
             return opts
-        while True:
-            line = proc.stdout.readline()
-            if not line:
-                break
-
-            s = line.decode("utf-8").rstrip("\r\n").split(" ", 1)
-            if len(s) == 2:
-                opts[s[0].lower()] = s[1]
-
-        return opts
 
     def get_known_hosts_file(self, *args, **kwargs) -> Optional[str]:
         k = self.get_ssh_host_keys()
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -6,6 +6,7 @@ from typing import (
     Mapping,
     Match,
     Any,
+    Dict,
     List,
     Optional,
     Union,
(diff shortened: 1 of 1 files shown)


nixops/backends/__init__.py
```diff
@@ -453,15 +454,62 @@ def get_keys(self):
...
+        """
+        return None
+
+    def _get_ssh_ambient_options(self) -> Dict[str, str]:
+        proc = subprocess.Popen(
+            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE
+        )
+        opts: Dict[str, str] = {}
+        if proc.stdout is None:  # mostly for mypy; Popen won't do this to us.
+            return opts
```
Mic92, line 477: This would be a bug:

```suggestion
        assert proc.stdout is not None  # mostly for mypy; Popen won't do this to us.
```

nixops/backends/__init__.py
```diff
@@ -453,15 +454,62 @@ def get_keys(self):
...
-            return ["-P", str(self.ssh_port)] if self.ssh_port is not None else []
+    def get_ssh_host_keys(self) -> Optional[str]:
+        """
+        Return the public host key in known_hosts format or None if not known.
+        """
+        return None
+
+    def _get_ssh_ambient_options(self) -> Dict[str, str]:
+        proc = subprocess.Popen(
+            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE
```
Mic92, line 473:
```suggestion
            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE, text=True
```

nixops/backends/__init__.py
```diff
@@ -453,15 +454,62 @@ def get_keys(self):
...
+        )
+        opts: Dict[str, str] = {}
+        if proc.stdout is None:  # mostly for mypy; Popen won't do this to us.
+            return opts
+        while True:
+            line = proc.stdout.readline()
+            if not line:
+                break
+
+            s = line.decode("utf-8").rstrip("\r\n").split(" ", 1)
```
Mic92, line 483:
```suggestion
            s = line.rstrip("\r\n").split(" ", 1)
```

nixops/backends/__init__.py
```diff
@@ -453,15 +454,62 @@ def get_keys(self):
...
+        proc = subprocess.Popen(
+            ["ssh", "-G", self.get_ssh_name()], stdout=subprocess.PIPE
+        )
+        opts: Dict[str, str] = {}
+        if proc.stdout is None:  # mostly for mypy; Popen won't do this to us.
+            return opts
+        while True:
+            line = proc.stdout.readline()
+            if not line:
+                break
```
Mic92, line 481:
```suggestion
        for line in proc.stdout:
```

nixops/backends/__init__.py
```diff
@@ -453,15 +454,62 @@ def get_keys(self):
...
-        if scp:
-            return ["-P", str(self.ssh_port)] if self.ssh_port is not None else []
+    def get_ssh_host_keys(self) -> Optional[str]:
+        """
+        Return the public host key in known_hosts format or None if not known.
+        """
+        return None
+
+    def _get_ssh_ambient_options(self) -> Dict[str, str]:
+        proc = subprocess.Popen(
```
Mic92, line 472: Change to:
```
with subprocess.Popen(...) as proc:
   # here comes the rest
```

This ensures the process is actually terminated afterwards.

roberth:
Thanks @Mic92 that was very helpful!

AmineChikhaoui (approved):

PR #1470, opened 2021-08-24
Locking usability improvements

roberth:
With remote state and locking, concurrent use of `nixops` commands has become impossible. This is bad when, for example, you want to `nixops ssh` while a `nixops deploy` is running.

This PR solves a couple of locking related problems by
 - creating a distinction between read-only and read-write NixOps commands
    - allowing read-only commands to run concurrently if supported by the lock backend
    - prevent corrupting state by uploading only when in read-write mode
 - making `nixops ssh` unlock before the session starts, so read-write operations can resume while presumably-interactive sessions exist
 - adding `nixops ssh --now` which uses whatever state it can get but does not lock at all

More details are in the commit messages.

### DIFF
--- a/nixops/args.py
+++ b/nixops/args.py
@@ -352,6 +352,13 @@ subparser.add_argument("machine", metavar="MACHINE", help="identifier of the mac
 subparser.add_argument(
     "args", metavar="SSH_ARGS", nargs=REMAINDER, help="SSH flags and/or command",
 )
+subparser.add_argument(
+    "--now",
+    dest="now",
+    default=False,
+    action="store_true",
+    help="do not acquire a lock before fetching the state",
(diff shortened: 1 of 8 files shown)


roberth:
Ok, I've improved typing precision by quite a bit, but `nixops.args` is hopeless. It doesn't seem like we can avoid explicit `Any`s whenever we add `add_argument`s. They're just part of its signature and they're counted even if we don't make use of the `Any`-typed params. It's not a real problem.
In fact this PR now improves typing precision.

roberth:
cc @adisbladis, would you mind reviewing this?

roberth:
This is sorely needed and has been up for review for nearly 3 months.

PR #1480, opened 2021-10-21
Make --network accept a file argument and add a --flake flag

talyz:
To make it possible to keep multiple network definitions in the same directory, make the `--network` flag accept a file in addition to a directory. Also, add a corresponding `--flake` flag and create short flags for both (`-n` and `-f` respectively).

### DIFF
--- a/nixops/script_defs.py
+++ b/nixops/script_defs.py
@@ -37,27 +37,42 @@ PluginManager.load()
 
 
 def get_network_file(args: Namespace) -> NetworkFile:
-    network_dir: str = os.path.abspath(args.network_dir)
+    network_path: str = ""
+    flake_dir: str = ""
+    flake_path: str = ""
 
-    if not os.path.exists(network_dir):
-        raise ValueError(f"{network_dir} does not exist")
+    network_path = args.network_path or os.path.join(os.getcwd(), "nixops.nix")
+    network_path = os.path.abspath(network_path)
+    if os.path.isdir(network_path):
+        network_path = os.path.join(network_path, "nixops.nix")
+    network_exists: bool = os.path.exists(network_path)
+    if args.network_path and not network_exists:
+        raise ValueError(f"{network_path} does not exist")
 
-    classic_path = os.path.join(network_dir, "nixops.nix")
-    flake_path = os.path.join(network_dir, "flake.nix")
-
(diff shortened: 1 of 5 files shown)


roberth:
This needs documentation and tests.

We have unit tests in tests.py and the tests directory, or https://github.com/NixOS/nixpkgs/pull/142035 could be added here for end to end tests.

Not sure if eval-machine-info needs to use the name, instead of `nixopsConfigurations.default`. Perhaps that  should be independent and part of the flake fragment: `--flake .#other-deployment`.

Documentation can be updated in doc/manual/nixops.rst

talyz:
> This needs documentation and tests.
> 
> We have unit tests in tests.py and the tests directory, or [NixOS/nixpkgs#142035](https://github.com/NixOS/nixpkgs/pull/142035) could be added here for end to end tests.

I'm not sure how to add this to the unit tests - it doesn't really feel suited for it. The NixOS test could be updated for it, though.

> Not sure if eval-machine-info needs to use the name, instead of `nixopsConfigurations.default`. Perhaps that should be independent and part of the flake fragment: `--flake .#other-deployment`.

I'm not really sure what this is referring to in relation to this change.

> Documentation can be updated in doc/manual/nixops.rst

I've added some basic parameter documentation now and removed the references I could find to the previous `-n` flag. There's more to document and more in the documentation to update, but I think it makes more sense to do a bigger rewrite when we know how the interface looks - should deployments and the legacy backend be removed or should legacy deployments be fixed instead? As I see it, the prior alternative would require more substantial documentation updates (and be a lot of work in general), while the latter would give us a confusing UX with a lot of old cruft.

nixops/script_defs.py
```diff
@@ -37,27 +37,41 @@
...
-        raise ValueError("f{network_dir} does not exist")
-
-    classic_path = os.path.join(network_dir, "nixops.nix")
-    flake_path = os.path.join(network_dir, "flake.nix")
-
-    classic_exists: bool = os.path.exists(classic_path)
+    network_path: str = ""
+    flake_path: str = ""
+
+    if args.network_path is not None:
```
pasqui23, line 43:
```suggestion
    if args.network_path is not None and args.flake_path is not None:
        raise ValueError("You can only provide one of either -f or -n")
    
    if args.network_path is not None:
```

PR #1482, opened 2021-11-01
util.ImmutableValidatedObject: Support Optional[ImmutableValidatedObject]

rehno-lindeque:
Similar to 88494fb80f6f9c3aa0f0f1591475a6ca181b2f98 this adds support for `Optional[ImmutableValidatedObject]`

### Usage

* Support for optional sub-modules with required sub-options.
* Simplify auto-generating plugin code from API definitions that follow this pattern.

E.g. this is not currently possible:

```nix
{
  misc = mkOption {
    default = null;
    type = types.nullOr (types.submodule miscOptions);
  };
}
```

#### Supporting additional generic types

It's trivial to add support for`Mapping[Any, ImmutableValidatedObject]` as well, but I didn't want to add new code that isn't actively used in a plugin.

#### Satisfying mypy

Mypy needs to be upgraded to remove `type: ignore[attr-defined]` after `typing.get_args` and `typing.get_origin`.

Never-the-less the tests pass just fine with this version of python:

```sh
$ python --version
Python 3.8.7
```


nixops/util.py
```diff
@@ -143,30 +146,37 @@ def __init__(self, *args: ImmutableValidatedObject, **kwargs):
...
+                and issubclass(type_args[0], ImmutableValidatedObject)
+            ):
+                # Support Sequence[ImmutableValidatedObject]
+                if isinstance(value, Sequence) and issubclass(tuple, type_origin):
+                    value = tuple(_transform_value(v, type_args[0]) for v in value)
+
+                # Support Optional[ImmutableValidatedObject]
+                if type_origin is Union:
+                    if value is not None:
+                        value = _transform_value(value, type_args[0])
```
rehno-lindeque, line 177: I discovered that the logic for this is slightly off, so hold off reviewing for now please.

Issue #1027, opened 2018-10-22
Cannot deploy trivial example


mcarvalho:
trivial.nix
```yml
{
    network.description = "Web Server";
    
    webserver = { config, pkgs, ... }: { 
        services.httpd.enable = true;
        services.httpd.adminAddr = "alice@example.org";
        services.httpd.documentRoot = "${pkgs.valgrind.doc}/share/doc/valgrind/html";
        networking.firewall.allowedTCPPorts = [80 22];     
    };
}
```
trivial-vbox.nix
```yml
{
  webserver =
    { config, pkgs, ... }:
    { deployment.targetEnv = "virtualbox";
      deployment.virtualbox.headless = true;
      deployment.virtualbox.memorySize = 1024; # megabytes
      deployment.virtualbox.vcpu = 2; # number of cpus
    };
}
```
after creating the deploy, when i run `$ nixops deploy -d trivial` the files above it fails to buld because nix can't find some hashes
```SHELL
copying path '/nix/store/y6lkznrphrb5q914my4cah6cnjcc7czw-findutils-4.6.0' to 'ssh://root@192.168.56.9'...
copying path '/nix/store/6wm0g6b6xwnx3qccn5af5mb3x8ksk035-stdenv-linux' to 'ssh://root@192.168.56.9'...
copying 1 paths...
copying path '/nix/store/fpffgjw3bvwba6kf3adwnmqlny6i49ps-append-initrd-secrets' from 'ssh://root@192.168.56.9'...
error: cannot add path '/nix/store/fpffgjw3bvwba6kf3adwnmqlny6i49ps-append-initrd-secrets' because it lacks a hash
builder for '/nix/store/l3awjif9sgvyml3qjkinpj3aqsx4yvn6-append-initrd-secrets.drv' failed with exit code 1
cannot build derivation '/nix/store/4ij1cky4czqqd9ijjisasrfg0yh5755v-nixos-system-webserver-19.03pre155614.ac7bf94f6df.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/js8hfd7902hwbv5viy0gljrkrfh2f62m-nixops-machines.drv': 1 dependencies couldn't be built
error: build of '/nix/store/js8hfd7902hwbv5viy0gljrkrfh2f62m-nixops-machines.drv' failed
error: unable to build all machine configurations
```

os version:
```SHELL
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.14
BuildVersion:	18A391
[...]

Issue #560, opened 2016-12-04
Help poor macOS users who can't build many closures on their host


3noch:
Poor macOS users often run into issues when trying to deploy with `nixops` because many packages must be built on Linux. Perhaps we can offer them some help in the documentation and potentially point them to "workarounds" like this one: https://github.com/3noch/nix-vbox-build-slave

copumpkin:
@puffnfresh is working on something that could help a lot with this

3noch:
@puffnfresh Do tell!

copumpkin:
He packaged HyperKit (Docker's fork of xhyve) and is making a wrapper for it to act as a background remote builder for Nix, sort of the way Docker for Mac works, except with an even more minimal OS in it that only builds Nix derivations. So basically a similar idea to the VBox build slave but fully managed by Nix, free, and with no kernel components to install (or even privileged processes).

It's a work in progress but what I've seen so far has been very promising!

domenkozar:
I have merged all darwin related issues into this one. There is also an issue for "buildin fully remotely" at https://github.com/NixOS/nixops/issues/260 and an unreviewed patch at https://github.com/NixOS/nixops/pull/412

proger:
Is it still desirable to build on the virtual machine you're deploying to?
I had to stop using nix-daemon on Darwin to make it just work out of the box — I assume this has something to do with https://github.com/NixOS/nix/issues/1061, if I'm testing on virtualbox I don't really want to run two VMs on a laptop.

Gabriel439:
@grahamc: In https://github.com/NixOS/nix/issues/1061#issuecomment-313860794 you noted that you were able to get distributed builds working on a multi-user Nix installation on OS X.  Is that documented anywhere?

grahamc:
I'm not sure I was able to get it working? Though, I'd definitely ping @lnl7 on that. I suspect nix-darwin helps out a lot here.

Gabriel439:
@grahamc: The reason I asked is that the way we got distributed builds working on multi-user Nix on OS X required running Nix commands as the `root` user (i.e. `sudo`).  In other words, in addition to setting up `NIX_BUILD_HOOK` and `NIX_REMOTE_SYSTEMS` we also had to run `sudo nix-build ...` for distributed build support to kick in.  However, having to run as `root` for distributed builds seems like the wrong way to do things.

3noch:
[...]

PR #1483, opened 2021-11-05
Clarify Linux requirement in README (macOS seems unsupported)

JanCVanB:
As a Nix(Ops) noob, this was particularly confusing - I assumed that since Nix supports macOS that NixOps would also support macOS. However, that does not seem to be true*. I think that adding a relevant note to the README will protect some future noobs from this confusion.

*Sources:
https://github.com/NixOS/nixops/issues/260
https://github.com/NixOS/nixops/issues/560
https://github.com/NixOS/nixops/issues/1027
https://github.com/NixOS/nixops/pull/412

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ $ nixops --version
 NixOps @version@
 ```
 
+**Note**: Currently NixOps only supports deploying from a Linux machine. macOS users should read
+[these](https://github.com/NixOS/nixops/issues/260)
+[open](https://github.com/NixOS/nixops/issues/560)
+[issues](https://github.com/NixOS/nixops/issues/1027).
+
 ### Building And Developing
 
 #### Building The Nix Package


roberth:
Did you try setting up a remote builder in a VM to build for linux?
like https://github.com/nix-community/linuxkit-nix
or https://github.com/LnL7/nix-docker#running-as-a-remote-builder

I would expect deployment to cloud vms and existing machines to work. It may not be optimal (#560) or to a local vbox (#1027), but there should be no reason for it not to work altogether when deploying from a mac.

JanCVanB:
I haven't tried those solutions, but if they are indeed the intended path for macOS users then we should add those links to the README instead. The NixOps User Guide doesn't contain any reference to "macOS" or "remote builder", and those linked repos' READMEs only mention NixOps once (in passing) between the two of them.

roberth:
A linux builder is the best solution.

In theory there could be other options, like cross compilation or building on the target, but those are sub-optimal to say the least. Not worth pursuing.

JanCVanB:
I tried using https://github.com/nix-community/linuxkit-nix, but the daemon wouldn't run for me on macOS 12 (Monterey). It looks like the repo is unmaintained and [hasn't worked since macOS 10 (Catalina)](https://github.com/nix-community/linuxkit-nix/issues/36).

JanCVanB:
I tried following the other repo's instructions [here](https://github.com/LnL7/nix-docker#running-as-a-remote-builder), but the daemon also wouldn't start. It looks like that repo is also unmaintained.

JanCVanB:
If there are no maintained first-party or third-party workarounds for macOS, then would it be accurate to say that NixOps doesn't support macOS?

roberth:
[UTM](https://mac.getutm.app/) is another alternative. It lets you run NixOS like you would on a physical machine. It's qemu underneath.
Plenty of other virtualization options exist too.

> would it be accurate to say that NixOps doesn't support macOS?

Technically, setting up a remote builder is a distinct activity that isn't the responsibility of NixOps. There's no inherent limitation that prevents NixOps from running on macOS, so saying that it is not supported would not be accurate.
I've updated the README with some minimal info. Perhaps you could add to it with information about your own linux builder setup?

roberth:
This isn't actionable now and the linux builder is more of an issue for the nix installer or nix-darwin to solve.

Issue #1282, opened 2020-04-14
Replacing nose by pytest


nlewo:
`nose` is no longer active [since 4 years](https://pypi.org/project/nose/) and could be replaced by `pytest`. Since `pytest` can execute `nose` tests, existing tests wouldn't need to be updated, while new tests could be written with the `pytest` framework.

This would also simplify how tests are executed since we would have to only execute a `pytest` command (instead of calling a Python script).

PR #1484, opened 2021-11-07
Migrate to Python 3.9 and pytest

lukebfox:
Upgraded Python from 3.7 to 3.9. Migrated from nose to pytest (#1282). Removed obsolete plugin tests*. Fixed typing issue revealed by newer mypy. Replaced strings with rawstrings in regex containing escaping to avoid deprecation warning. Dropped setup.cfg as the tools we use it for are now pep517 compliant, and config now resides in pyproject.toml.


Mic92:
I went over the changes and they look fine to me.
The ci error seems to be caused not by the changes here: https://github.com/NixOS/nixops/runs/4128582799?check_suite_focus=true#step:5:3710
but the lack of a set `NIX_PATH` in ci configuration.

doc/requirements.txt
```diff
@@ -4,264 +4,388 @@ alabaster==0.7.12; python_version >= "3.5" \
...
-    --hash=sha256:dacca89f4bfadd5de3d7489b7c8a566eee0d3676333fbb50030263894c38c0dc \
-    --hash=sha256:d2b5255c7c6349bc1bd1e59e08cd12acbbd63ce649f2588755783aa94dfb6b1a
-colorama==0.4.4; python_version >= "3.5" and python_full_version < "3.0.0" and sys_platform == "win32" or python_full_version >= "3.5.0" and sys_platform == "win32" and python_version >= "3.5" \
+certifi==2021.10.8; python_version >= "3.5" and python_full_version < "3.0.0" or python_full_version >= "3.6.0" and python_version >= "3.5" \
+    --hash=sha256:d62a0163eb4c2344ac042ab2bdf75399a71a2d8c7d47eac2e2ee91b9d6339569 \
+    --hash=sha256:78884e7c1d4b00ce3cea67b44566851c4343c120abd683433ce934a68ea58872
+charset-normalizer==2.0.7; python_full_version >= "3.6.0" and python_version >= "3.5" \
+    --hash=sha256:e019de665e2bcf9c2b64e2e5aa025fa991da8720daa3c1138cadd2fd1856aed0 \
+    --hash=sha256:f7af805c321bfa1ce6714c51f254e0d5bb5e5834039bc17db7ebe3a4cec9492b
+click==8.0.3; python_version >= "3.6" \
```
Mic92, line 25: the documentation build is somehow unhappy about click here:

```
    importlib-metadata from https://files.pythonhosted.org/packages/71/c2/cb1855f0b2a0ae9ccc9b69f150a7aebd4a8d815bd951e74621c4154c52a8/importlib_metadata-4.8.1-py3-none-any.whl#sha256=b618b6d2d5ffa2f16add5697cf57a46c76a56229b0ed1c438322e4e95645bd15 (from click==8.0.3->-r doc/requirements.txt (line 25))
```

Mic92: Looks like an issue in click itself.

Mic92: Here is the `setup.py` of click: https://github.com/pallets/click/blob/6411f425fae545f42795665af4162006b36c5e4a/setup.py#L7

roberth:
I noticed you updated mypy as well. I've improved the types a bit in https://github.com/NixOS/nixops/pull/1485 but I don't know if it's enough to satisfy the mypy-ratchet ci job. Feel free to take the useful bits from that draft pr.

doc/requirements.txt
lukebfox: Looks like the docs CI process is using python 3.7 for the virtualenv it's using, which meets the `python_version < '3.8'`conditions in click's `setup.py`. But we haven't pinned a version for this `importlib_metadata` in our `doc/requirements.txt` because we shouldn't be installing it. From docs:
```
$ python3.7 -mvirtualenv /home/docs/checkouts/readthedocs.org/user_builds/nixops/envs/1484
...
$ /home/docs/checkouts/readthedocs.org/user_builds/nixops/envs/1484/bin/python -m pip install --exists-action=w --no-cache-dir -r doc/requirements.txt
```
Unfortunately the docs CI is separate to nixops; I can't see/change how this is configured.

Issue #1290, opened 2020-04-20
Deploy as part of CI


grahamc:
At the end of the day, `nixops deploy` is the key feature we have here. We should do it as part of CI. Also nice would be being able to do deploys with various backends, like `nixops-aws` to see if PRs to nixops breaks a plugin.

roberth:
As a low hanging fruit, we could test the `none` backend in a `nixosTest`, because it does not depend on real cloud services. As such, it will be easy to run, which is actually kinda great.

roberth:
This seems like a good starting point: https://github.com/NixOS/nixpkgs/pull/142035

PR #1493, opened 2021-12-21
Include sshOptions when building ssh flags

cprussin:
Looks like this was broken in https://github.com/NixOS/nixops/commit/94a68844451ed5e81b25f1f864c27ee4ba6e5f76

### DIFF
--- a/nixops/backends/__init__.py
+++ b/nixops/backends/__init__.py
@@ -489,7 +489,7 @@ class MachineState(
             return None
 
     def get_ssh_flags(self, scp: bool = False) -> List[str]:
-        flags: List[str] = []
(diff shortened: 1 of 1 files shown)


cprussin:
cc @roberth

roberth (approved):
The original implementation did not pass the ssh options to `scp`. This does, so that's a difference, but I think it makes sense because it seems like the options will be about the connection and not about terminal stuff and such, so that should mostly work, except when using command-specific shortcuts.
It'd be great if you could add a regression test. Some prep work is necessary. https://github.com/NixOS/nixops/issues/1290#issuecomment-921072855

PR #1505, opened 2022-02-09
Exploratory: Add bootstrap script hook

exFalso:
Hi, this PR is more of an exploratory PR, to see whether this idea is useful for the general public. Internally at our company we're using nixops to deploy to **non-NixOS** systemd distributions. One of the issues we used to have is that nixops makes certain assumptions about the target machine, namely that it has a working multi-user nix installation, and that it has a `keys` group defined for secrets. For a while we've handled this explicitly outside nixops, we basically made the machine look like NixOS for the purposes of nixops.

However, with a small addition to nixops, we now tackle this issue much more elegantly. Namely, this PR adds a nullable config option `deployment.bootstrapScript`, which (if set) is run on the master SSH connection before any nix-specific actions are taken, but after the resource closure has been built. This allows us to do the nix installation and certain early setup (like mounting of /nix from a separate hard disk), independent of the /nix/store.

The actual mechanism I added is a bit broken as it only handles None type targets, but the control flow of the resource configuration is a bit too tricky and I couldn't figure out how to do it for a general resource type.

Question is, is this a useful enough feature to be included upstream? If so, can I get any pointers to how to expose the option to a more generic resource type? Also, the changes are on top of the 1.7 tag (the PR is submitted against release-1.7, which seems to have diverged from the tag(??)), would this feature go into 2.0 only? I tried upgrading to 2.0 but got into a bunch of unrelated issues with deployment.

Thank you!

### DIFF
--- a/nix/options.nix
+++ b/nix/options.nix
@@ -57,6 +57,16 @@ in
       '';
     };
 
+    deployment.bootstrapScript = mkOption {
+      type = types.nullOr types.lines;
+      default = null;
+      description = ''
+        This option specifies an optional bash script to be run on the target
+        before nix-specific operations are run. It only assumes `bash` to be
+        installed on the target.
+      '';
+    };
+
     deployment.targetPort = mkOption {
       type = types.int;
       description = ''
--- a/doc/manual/release-notes.xml
+++ b/doc/manual/release-notes.xml
@@ -5,7 +5,7 @@
 
 <!--==================================================================-->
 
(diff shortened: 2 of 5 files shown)


nix/options.nix
```diff
@@ -57,6 +57,16 @@ in
       '';
     };
 
+    deployment.bootstrapScript = mkOption {
+      type = types.nullOr types.lines;
+      default = null;
```
pasqui23, line 62:
```suggestion
      type = types.lines;
      default = "";
```

Issue #1486, opened 2021-11-18
`require` -> `imports`


roberth:
I believe a custom `require` feature was chosen in order to solve a bootstrapping problem.
This custom mechanism is not ideal because it adds to the growing list of "override", "overlay", "module", etc. It'd be better to reuse the module system. I suggest the following bootstrapping

1. Load the raw network expressions as Nix functions
2. Call them with bogus auto-arguments as in `module (mapAttrs (k: v: throw k) (functionArgs module)`
3. This should give us module attrsets containing a few attributes of interest that we'll evaluated and a bunch of things we don't need and can't evaluate at this stage. That's ok.
4. Use these partial modules to determine what `lib` should be.
5. Call `lib.evalModules` on the network expressions.
6. Have the module system at our disposal

A more radical approach, suitable for flakes, is to let the user inject `lib` manually. This would change the flake format to require a call like `nixopsConfigurations.default = inputs.nixops.lib.mkNetwork { inherit lib; configuration = /* ... */; }`, which is more robust than step 3 and 4. For non-flake `nixops.nix` we'll still want something like step 3 and 4 I think.

Pros:
 - module system is more familiar
 - module system is more powerful a custom composition method will be
    - allows the user to define network-wide options to capture the variability of environments (dtap, etc)
    - allows the user to define network-wide options to represent cross-machine concepts such as groups of machines
    - network arguments can become option values
    - familiar infrastructure for deprecations
    - replace custom, buggy network expression merging code

Cons:
 - breaks pre-release version (not that bad)
 - step 3 and 4 are somewhat fragile. I am confident that this can be mitigated

Myths:
 - "the module system is slow; look at NixOS": the module system scales with the number of options. It is very fast for small configuration systems. It is also quite lazy, so it is possible to avoid evaluating the machines for example. I see no reason for performance to degrade because of this change.

PR #1506, opened 2022-02-17
Switch from fixMergeModules to evalModules

talyz:
`fixMergeModules` is deprecated and using it produces deprecation warning messages in newer revisions of nixpkgs. Since it's implemented as a simple wrapper around `evalModules`, let's just use `evalModules` instead and set the module arguments through `config._module.args`.

Fixes #1507

Issue #1429, opened 2021-01-12
targetHost Address ist not updated in the database when changed in the nix file


madonius:
When I change the `deployment.targetHost` of an existing Host and update the corresponding entry in the Database is not being updated either.

tewfik-ghariani:
Did you run a **deploy** operation? What is the output of the "info" sub-command after the deployment?

madonius:
It asks for the root password, yet the machine I intended to connect to would be accessible with the ssh-key.
Running `ssh … -vvv` got me on the track that in truth it tries to connect to the previous address.

tewfik-ghariani:
I just gave it a quick test and I see that the 'targetHost' was updated without any issues. Which nixops version are you using? Master?

Can you please share the output of the `export` command ( while redacting any sensitive information )

madonius:
Yes, I could replicate it on master. Will do

PR #1508, opened 2022-02-20
eval-machine-info.nix: ported to modules

pasqui23:
Fixes https://github.com/NixOS/nixops/issues/1486
Done:
* [X] Made `requires` an error.
* [X] Implemented the `nodes` option.
* [X] Made a fallback for the old way of writing nodes to the 
To do:
* [ ] Implemented all `resources` options
* [ ] Implemented all `network` options
* [ ] Tested it


nix/eval-machine-info.nix
```diff
@@ -11,104 +11,136 @@
...
-      then (builtins.head network.nixpkgs).legacyPackages.${system}
-      else throw "NixOps network must have a 'nixpkgs' attribute"
-    else (builtins.head (network.network)).nixpkgs or (import <nixpkgs> { inherit system; });
+  inherit (lib) mkOption types;
 
-  inherit (pkgs) lib;
+  stage1Eval = lib.evalModules {
+    specialArgs = args;
+    modules = [
+      (lib.mkRemovedOptionModule [ "require" ] "Use the imports option instead.")
```
pasqui23, line 28: @roberth you mentioned that the module system supports `requires`,how?

roberth: `require` has the same meaning as `imports`: https://github.com/NixOS/nixpkgs/blob/d81f048ef4ce360ba2d11d3d7c94398be134b380/lib/modules.nix#L368

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,136 @@
...
-      networkExprClosure = builtins.genericClosure {
-        startSet = map exprToKey networkExprs;
-        operator = { key }: map exprToKey ((getNetworkFromExpr key).require or []);
-      };
-    in
-      map ({ key }: getNetworkFromExpr key) networkExprClosure
-      ++ optional (flakeUri != null)
-        ((call flakeExpr) // { _file = "<${flakeUri}>"; });
+  nixpkgs1 = <nixpkgs>; # this will be replaced on install by nixops' nixpkgs input
+  lib = import "${nixpkgs1}/lib";
```
roberth, line 21: We should use this exclusively for getting the `nixpkgs` option value, and then use its lib for the rest of the evaluation.

```suggestion
  libBootNixpkgs = <nixpkgs>; # this will be replaced on install by nixops' nixpkgs input
  libBootLib = import "${nixpkgs1}/lib";
```

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,136 @@
...
+                    deployment.targetHost = lib.mkOverride 900 name;
+                    environment.checkConfigurationOptions = lib.mkOverride 900 checkConfigurationOptions;
+                    nixpkgs.system = lib.mkDefault system;
+                  })
+                ];
+            });
+          };
+          defaults = mkOption {
+            type = with types;
+              # TODO: use types.raw once this PR is merged: https://github.com/NixOS/nixpkgs/pull/132448
```
roberth, line 111: I think we'll want a type that doesn't prohibit merging, but returns a list or a `mkMerge` instead.

This option is actually somewhat redundant, as you can merge modules into all nodes by redeclaring `options.nodes`, but `defaults` is expected and convenient.

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,134 @@
...
+  inherit (evalBoot.config) nixpkgs;
+
+  pkgs = nixpkgs.legacyPackages.${system} or import nixpkgs { inherit system; };
+  lib = nixpkgs.lib or pkgs.lib or libBoot;
+  inherit (lib) mkOption types removeAttrs;
+
+in
+rec {
+
+  network = (evalBoot.extendModules {
```
roberth, line 48: This `extendModules` comes from `libBoot`, so this should use `lib.evalModules` from scratch instead.

(Sorry if I'm repetitive; made some commit comments by accident earlier)

roberth:
Heads up about the resources; just a merged a PR that touches that.

pasqui23:
what pr @roberth ?

roberth:
https://github.com/NixOS/nixops/pull/1506

pasqui23:
So do I have to rebase on master?

pasqui23:
Ok I did rebase on master

pasqui23:
Ok I've tried and failed to build a network, it goes in infinite recursion while evaluating `scrubOptionValues` in `info.machines`

pasqui23:
It still goes in infinite recursion, I need  to know what nixops expects exactly from this file and what can be safely removed.

roberth:
> it goes in infinite recursion

Maybe the `attrsOf`s need to be `lazyAttrsOf`s.
`attrsOf` can only return its attrset by inspecting the children for `mkIf false` values and filtering out those attributes. That makes it more strict than `lazyAttrsOf`, which will return the spine of the attribute set (attr names + thunks) without inspecting those thunks.

roberth:
> I need to know what nixops expects exactly

I don't expect this to be documented anywhere.

pasqui23:
> > I need to know what nixops expects exactly
> 
> I don't expect this to be documented anywhere.

I don't need to be fully documented,only to tell me what the nixops cli and plugins take from the `eval-machine-info.nix` file *right now*. If, after merging this, you go and change that I would not care abut it.

pasqui23:
> > it goes in infinite recursion
> 
> Maybe the `attrsOf`s need to be `lazyAttrsOf`s. `attrsOf` can only return its attrset by inspecting the children for `mkIf false` values and filtering out those attributes. That makes it more strict than `lazyAttrsOf`, which will return the spine of the attribute set (attr names + thunks) without inspecting those thunks.

Tried with `lazyAttrsOf` but it still gives me the same problems.

pasqui23:
Ok it was an error on my part, I forgot `network.nodesExtraArgs = { inherit inputs; };` in my config. 🤦🤦🤦🤦

pasqui23:
Ok now the error I get is 
```
; nixops list
error: The option `resources' is used but not defined.
```

Which I don't know where even to begin

roberth:
> ```
> ; nixops list
> error: The option `resources' is used but not defined.
> ```

Just add a `default = { };` to the `resources` option.

It's some silly ux :/

pasqui23:
Ok latest error is definetly on the python side
```
; nixops list                       
def-matcher: command not found
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/bin/.nixops-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/__main__.py", line 56, in main
    args.op(args)
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/script_defs.py", line 200, in op_list_deployments
    depl.evaluate()
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 431, in evaluate
    self.evaluate_network()
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 423, in evaluate_network
    self.description = config.get("description", self.default_description)
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/util.py", line 493, in set
    self._set_attr(name, x)
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 250, in _set_attr
    self._set_attrs({name: value})
  File "/nix/store/nbrhfv3wcdsiish3y3k64jgs9826k2xc-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 243, in _set_attrs
    c.execute(
sqlite3.OperationalError: attempt to write a readonly database
```

roberth:
> sqlite3.OperationalError: attempt to write a readonly database

This is a known issue. https://github.com/NixOS/nixops/issues/1490

At this time, `nixops deploy` is more suitable for testing, as this command does request write access to the db.

pasqui23:
```
; nixops deploy                  
def-matcher: command not found
flatwoody> generating new SSH key pair... done
Password: 
flatwoody> setting state version to 22.05
flatwoody> waiting for SSH...
trace: warning: Please use the actual nodes.* option instead of assigning machines to the config's top level
error: 'flatwoody' at /tmp/nixops-tmpnvee6p43/physical.nix:2:15 called without required argument 'pkgs'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:199:25:

          198|                 file = def.file;
          199|                 value = def.value arg;
             |                         ^
          200|               }) defs);

       … while evaluating the attribute 'value'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:199:17:

          198|                 file = def.file;
          199|                 value = def.value arg;
             |                 ^
          200|               }) defs);

       … while evaluating 'getType'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:171:21:

          170|         let
          171|           getType = value:
             |                     ^
          172|             if isAttrs value && isCoercibleToString value

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:179:16:

          178|           commonType = foldl' (type: def:
          179|             if getType def.value == type
             |                ^
          180|             then type

       … while evaluating anonymous lambda

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:178:38:

          177|           # don't have the same type
          178|           commonType = foldl' (type: def:
             |                                      ^
          179|             if getType def.value == type

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:178:24:

          177|           # don't have the same type
          178|           commonType = foldl' (type: def:
             |                        ^
          179|             if getType def.value == type

       … while evaluating 'merge'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:169:20:

          168|       check = value: true;
          169|       merge = loc: defs:
             |                    ^
          170|         let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:195:38:

          194|             stringCoercibleSet = mergeOneOption;
          195|             lambda = loc: defs: arg: anything.merge
             |                                      ^
          196|               (loc ++ [ "<function body>" ])

       … while evaluating 'lambda'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:195:33:

          194|             stringCoercibleSet = mergeOneOption;
          195|             lambda = loc: defs: arg: anything.merge
             |                                 ^
          196|               (loc ++ [ "<function body>" ])

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:530:46:

          529|         coerce = unify: value: if isFunction value
          530|           then setFunctionArgs (args: unify (value args)) (functionArgs value)
             |                                              ^
          531|           else unify (if shorthandOnlyDefinesConfig then { config = value; } else value);

       … while evaluating 'unifyModuleSyntax'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:343:34:

          342|      of ‘options’, ‘config’ and ‘imports’ attributes. */
          343|   unifyModuleSyntax = file: key: m:
             |                                  ^
          344|     let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:530:39:

          529|         coerce = unify: value: if isFunction value
          530|           then setFunctionArgs (args: unify (value args)) (functionArgs value)
             |                                       ^
          531|           else unify (if shorthandOnlyDefinesConfig then { config = value; } else value);

       … while evaluating anonymous lambda

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:530:33:

          529|         coerce = unify: value: if isFunction value
          530|           then setFunctionArgs (args: unify (value args)) (functionArgs value)
             |                                 ^
          531|           else unify (if shorthandOnlyDefinesConfig then { config = value; } else value);

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:397:8:

          396|       # works.
          397|     in f (args // extraArgs)
             |        ^
          398|   else

       … while evaluating 'applyIfFunction'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:373:29:

          372|
          373|   applyIfFunction = key: f: args@{ config, options, lib, ... }: if isFunction f then
             |                             ^
          374|     let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:272:55:

          271|         if isFunction m || isAttrs m then
          272|           unifyModuleSyntax fallbackFile fallbackKey (applyIfFunction fallbackKey m args)
             |                                                       ^
          273|         else if isList m then

       … while evaluating 'unifyModuleSyntax'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:343:34:

          342|      of ‘options’, ‘config’ and ‘imports’ attributes. */
          343|   unifyModuleSyntax = file: key: m:
             |                                  ^
          344|     let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:272:11:

          271|         if isFunction m || isAttrs m then
          272|           unifyModuleSyntax fallbackFile fallbackKey (applyIfFunction fallbackKey m args)
             |           ^
          273|         else if isList m then

       … while evaluating 'loadModule'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:270:53:

          269|       # Like unifyModuleSyntax, but also imports paths and calls functions if necessary
          270|       loadModule = args: fallbackFile: fallbackKey: m:
             |                                                     ^
          271|         if isFunction m || isAttrs m then

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:311:22:

          310|           let
          311|             module = loadModule args parentFile "${parentKey}:anon-${toString n}" x;
             |                      ^
          312|             collectedImports = collectStructuredModules module._file module.key module.imports args;

       … while evaluating anonymous lambda

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:328:31:

          327|           disabledKeys = map moduleKey disabled;
          328|           keyFilter = filter (attrs: ! elem attrs.key disabledKeys);
             |                               ^
          329|         in map (attrs: attrs.module) (builtins.genericClosure {

       … from call site

       … while evaluating 'filterModules'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:324:36:

          323|       # modules recursively. It returns the final list of unique-by-key modules
          324|       filterModules = modulesPath: { disabled, modules }:
             |                                    ^
          325|         let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:335:7:

          334|     in modulesPath: initialModules: args:
          335|       filterModules modulesPath (collectStructuredModules unknownModule "" initialModules args);
             |       ^
          336|

       … while evaluating anonymous lambda

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:334:37:

          333|
          334|     in modulesPath: initialModules: args:
             |                                     ^
          335|       filterModules modulesPath (collectStructuredModules unknownModule "" initialModules args);

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:180:25:

          179|       merged =
          180|         let collected = collectModules
             |                         ^
          181|           (specialArgs.modulesPath or "")

       … while evaluating 'reverseList'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/lists.nix:393:17:

          392|   */
          393|   reverseList = xs:
             |                 ^
          394|     let l = length xs; in genList (n: elemAt xs (l - n - 1)) l;

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:184:33:

          183|           ({ inherit lib options config specialArgs; } // specialArgs);
          184|         in mergeModules prefix (reverseList collected);
             |                                 ^
          185|

       … while evaluating 'byName'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:449:25:

          448|       */
          449|       byName = attr: f: modules:
             |                         ^
          450|         zipAttrsWith (n: concatLists)

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:466:21:

          465|       # an attrset 'name' => list of submodules that declare ‘name’.
          466|       declsByName = byName "options" (module: option:
             |                     ^
          467|           [{ inherit (module) _file; options = option; }]

       … while evaluating the attribute 'matchedOptions'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:510:14:

          509|     in {
          510|       inherit matchedOptions;
             |              ^
          511|

       … while evaluating 'mapAttrsRecursiveCond'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/attrsets.nix:296:36:

          295|   */
          296|   mapAttrsRecursiveCond = cond: f: set:
             |                                    ^
          297|     let

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:192:28:

          191|           # For definitions that have an associated option
          192|           declaredConfig = mapAttrsRecursiveCond (v: ! isOption v) (_: v: v.value) options;
             |                            ^
          193|

       … while evaluating the attribute 'config'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:257:9:

          256|         options = checked options;
          257|         config = checked (removeAttrs config [ "_module" ]);
             |         ^
          258|         _module = checked (config._module);

       … while evaluating 'merge'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:569:22:

          568|         check = x: isAttrs x || isFunction x || path.check x;
          569|         merge = loc: defs:
             |                      ^
          570|           (base.extendModules {

       … from call site

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:649:59:

          648|       if isDefined then
          649|         if all (def: type.check def.value) defsFinal then type.merge loc defsFinal
             |                                                           ^
          650|         else let allInvalid = filter (def: ! type.check def.value) defsFinal;

       … while evaluating the attribute 'value'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/modules.nix:660:27:

          659|     optionalValue =
          660|       if isDefined then { value = mergedValue; }
             |                           ^
          661|       else {};

       … while evaluating anonymous lambda

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/lib/types.nix:413:22:

          412|       merge = loc: defs:
          413|         mapAttrs (n: v: v.value) (filterAttrs (n: v: v ? value) (zipAttrsWith (name: defs:
             |                      ^
          414|             (mergeDefinitions (loc ++ [name]) elemType defs).optionalValue

       … from call site

       … while evaluating the attribute 'config.system.build.toplevel'

       at /nix/store/kxlj5h6jss1szi1rs0lkccggr18g0lza-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nix/eval-machine-info.nix:161:32:

          160|   #TODO: take options and auter modules outputs for each node
          161|   nodes = lib.mapAttrs (n: v: {config = v;}) net.config.nodes;
             |                                ^
          162|

       … while evaluating anonymous lambda

       at /nix/store/kxlj5h6jss1szi1rs0lkccggr18g0lza-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nix/eval-machine-info.nix:309:54:

          308|         mkdir -p $out
          309|         ${toString (lib.attrValues (lib.mapAttrs (n: v: ''
             |                                                      ^
          310|           ln -s ${v.config.system.build.toplevel} $out/${n}

       … from call site

       … while evaluating the attribute 'buildCommand' of the derivation 'nixops-machines'

       at /nix/store/531v6sf2db04x3xqphfkh7zg2dgw135k-source/pkgs/stdenv/generic/make-derivation.nix:205:7:

          204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
          205|       name =
             |       ^
          206|         let
error: evaluation of the deployment specification failed

```

pasqui23:
Where it would ask for a pkgs param?

roberth:
The error seems to originate from the "physical" expression that's generated by nixops. You can see that expression with `nixops show-physical`.

pasqui23:
```
; nixops show-physical
def-matcher: command not found
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/bin/.nixops-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/__main__.py", line 56, in main
    args.op(args)
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/script_defs.py", line 813, in op_show_physical
    depl.evaluate()
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 431, in evaluate
    self.evaluate_network()
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 423, in evaluate_network
    self.description = config.get("description", self.default_description)
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/util.py", line 493, in set
    self._set_attr(name, x)
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 250, in _set_attr
    self._set_attrs({name: value})
  File "/nix/store/3y06h6hsw56r0bszn9pl7cdpz91qjbh1-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nixops/deployment.py", line 243, in _set_attrs
    c.execute(
sqlite3.OperationalError: attempt to write a readonly database
```

pasqui23:
https://github.com/NixOS/nixops/issues/1490 strikes again☹

roberth:
> #1490 strikes againfrowning_face

https://github.com/NixOS/nixops/pull/1511 should work around it.

pasqui23:
```
; nixops show-physical
def-matcher: command not found
{
  flatwoody = { config, lib, pkgs, ... }: {
    config = {
      boot.kernelModules = [];
      networking = { extraHosts = "\n"; firewall.trustedInterfaces = []; };
      system.stateVersion = ( lib.mkDefault "22.05" );
    };
    imports = [
      {
        config.users.extraUsers.root.openssh.authorizedKeys.keys = [
          "ssh-ed25519 XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX NixOps client key for flatwoody"
        ];
      }
    ];
  };
}
```

pasqui23:
The worst thing is that `pkgs` is not even used …

roberth:
It seems that the "physical" expression is using the legacy format compatibility which isn't working correctly.

```
trace: warning: Please use the actual nodes.* option instead of assigning machines to the config's top level
```

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,170 @@
...
+                    networking.hostName = lib.mkOverride 900 name;
+                    deployment.targetHost = lib.mkOverride 900 name;
+                    environment.checkConfigurationOptions = lib.mkOverride 900 checkConfigurationOptions;
+                    nixpkgs.system = lib.mkDefault system;
+                  })
+                ];
+            }).type;
+          };
+          defaults = mkOption {
+            type = types.anything;
```
roberth, line 142: `anything` would try to merge things on its own when multiple `defaults` values are defined. Instead we ought to defer this to the module system.
Something like https://github.com/NixOS/nixpkgs/pull/152785 or perhaps better, a type that returns a `mkMerge`.

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,170 @@
...
+            default = { };
+            description = ''
+              Extra NixOS options to add to all nodes.
+            '';
+          };
+        };
+        config = let
+          nodes = removeAttrs config (builtins.attrNames options);
+        in lib.mkIf ({} != nodes) { #TODO: actual warning/assert module impl.
```
roberth, line 151: Module system built-in functionality would be best (https://github.com/NixOS/nixpkgs/pull/97023#issuecomment-1059445608) but hasn't been implemented yet. It's a little tricky with strictness, but still feasible.

nix/eval-machine-info.nix
```diff
@@ -11,104 +11,170 @@
...
-  inherit (pkgs) lib;
+in
+rec {
+  inherit nixpkgs;
 
-  # Expose path to imported nixpkgs (currently only used to find version suffix)
-  nixpkgs = builtins.unsafeDiscardStringContext pkgs.path;
+  net = lib.evalModules {
+    specialArgs = args;
+    modules = baseMods lib ++ [
```
roberth, line 54: We'll probably want a more specific `freeformType`. `anything` was acceptable for bootstrapping because we don't use the value, but for legacy `nodes` we need to inject `pkgs`.

pasqui23:
> It seems that the "physical" expression is using the legacy format compatibility which isn't working correctly.
> 
> ```
> trace: warning: Please use the actual nodes.* option instead of assigning machines to the config's top level
> ```

No, I just have written the warning incorrectly

pasqui23:
> anything would try to merge things on its own when multiple defaults values are defined. Instead we ought to defer this to the module system.
> Something like https://github.com/NixOS/nixpkgs/pull/152785 or perhaps better, a type that returns a mkMerge.

How would a type return a mkMerge?

pasqui23:
> It seems that the "physical" expression is using the legacy format compatibility which isn't working correctly.
> 
> ```
> trace: warning: Please use the actual nodes.* option instead of assigning machines to the config's top level
> ```

No, I just have written the warning incorrectly

nix/eval-machine-info.nix
roberth: Let's keep it simple. `types.raw` should be fine in almost all cases. It seems unlikely that the module system would touch those values too early, and impossible when they're in function bodies, which is very often the case.

roberth:
> > It seems that the "physical" expression is using the legacy format compatibility which isn't working correctly.
> > ```
> > trace: warning: Please use the actual nodes.* option instead of assigning machines to the config's top level
> > ```
> 
> No, I just have written the warning incorrectly

To pick out the legacy nodes from the freeformtype more easily, you can postcompose a function that puts the all the freeform stuff in a separate attribute.

Here's what that looks like (crux: `freeformItems`), although it shouldn't use `// { merge = ... }`.
https://github.com/NixOS/nixpkgs/pull/163597/files#diff-619679c9f04a185497c93f14a481d1fe65cc75298f422169128a902ba907ffa2

It'd be better to use a copy of `postTransform` (https://github.com/NixOS/nixpkgs/pull/163601) instead of `// { merge = ... }`.

nix/eval-machine-info.nix
roberth: Here's a new type that accepts single modules and returns them as a list suitable for `imports` or `modules`. https://github.com/NixOS/nixpkgs/pull/163617

pasqui23:
Using your proposed `deferredModule` instead of `anything` in `freeformType` will give me:
``` 
; nixops create -n net.nix
def-matcher: command not found

WARNING: NixOps 1.0 -> 2.0 conversion step required

NixOps 2.0 added support for multiple storage backends.

Upgrade steps:
1. Open /home/me/src/my-nixos-conf
2. Add:
    network.storage.legacy = {
      databasefile = "~/.nixops/deployments.nixops";
    };
3. Rerun

See https://nixops.readthedocs.io/en/latest/manual/migrating.html#state-location for more guidance.
```
I suspect other storage backends would have similar problems as they would not be written with modules in mind.

pasqui23:
If  I try to use `deferredModule` for `default.type` it goes in infinite recursion while creating the net.

pasqui23:
```
; nixops deploy           
def-matcher: command not found
error: Module imports can't be nested lists. Perhaps you meant to remove one level of lists? Definitions:
       - In `/nix/store/v6mzngi4lmbl7lbw12w5f4fhbvdl2vak-python3.8-nixops-2.0.0/lib/python3.8/site-packages/nix/eval-machine-info.nix':
           [
             {
               _file = "defaults from <unknown-file>";
               imports = [
                 /nix/store/cad0w82676y2k2k06bpplnx944r2wafr-source/modules/core
           ...

       … while evaluating 'loadModule'

```
If only `defaults.type = deferredModules`

pasqui23:
Even after making `defaults.type = deferredModules` it still does the same error.

pasqui23:
Where is the code evaluating `physical.nix`?

roberth:
> Where is the code evaluating `physical.nix`?

Roughly here it seems
 https://github.com/NixOS/nixops/blob/9b3ac919d87db872854fb528064821b959ac38c1/nixops/deployment.py#L639-L660

roberth:
> I suspect other storage backends would have similar problems as they would not be written with modules in mind.

Probably the `network` attr is treated as a legacy node instead of options?

pasqui23:
> Probably the `network` attr is treated as a legacy node instead of options?

What do you mean? `nodes = removeAttrs config (builtins.attrNames options);` was written specifically to remove `network` and other from the legacy nodes.

pasqui23:
However the `network.storage.legacy`,as it is not defined by our config, it fall through to `freeformType` and gets merged as `[{storage.legacy=...;}]`.
Again without actual options to configure the storage options it still would not be merged without the `anything` type.

pasqui23:
> > Where is the code evaluating `physical.nix`?
> 
> Roughly here it seems
> 
> https://github.com/NixOS/nixops/blob/9b3ac919d87db872854fb528064821b959ac38c1/nixops/deployment.py#L639-L660

That is the part that seems to generate the file. I mean what part in the `nixops deploy` command that runs `nix-instantiate` with the generated `physical.nix`?

roberth:
> `nix-instantiate`

That'd be `nixops/evaluation.py`

roberth:
> However the `network.storage.legacy`,as it is not defined by our config, it fall through to `freeformType` and gets merged as `[{storage.legacy=...;}]`.
> Again without actual options to configure the storage options it still would not be merged without the `anything` type.

I don't think I fully understand this problem as it relates to the code. Did you intend to rename this option?

Perhaps you could define only `storage` as an empty submodule with an `anything` `freeformType`?

Sorry if I can only throw ideas at you now. I'll have a more thorough look later.

pasqui23:
> Perhaps you could define only `storage` as an empty submodule with an `anything` `freeformType`?

Tried that,I get
```
; nixops create -n net.nix
def-matcher: command not found
error: The option `network.storage' is used but not defined.
```

pasqui23:
> Perhaps you could define only `storage` as an empty submodule with an `anything` `freeformType`?

Tried that but I get:
```
; nixops create -n net.nix
error: The option `network.storage' is used but not defined.
```

pasqui23:
And yes adding `default ={}` did not solve it.

pasqui23:
Ok I've only tested it as the world's worst replacement for `nixos-rebuild` but it works!

pasqui23:
@roberth is there anything I have to do change about this pr?

pasqui23:
Also when I try to deploy my test it always ask me for password even if `deployment.provisionSSHKey` is defaulted to true.

roberth:
Hi @pasqui23,

I've tried your PR and made some followup changes in #1515 (and accidentally pushed here; reverted that)


> Also when I try to deploy my test it always ask me for password even if `deployment.provisionSSHKey` is defaulted to true. Is that related to #1429 ?

I don't know. Did you try without this PR? It's possible that recent changes to ssh flags handling haven't been fixed completely yet. #1464 (ssh keys for remote state) caused #1493 and subsequently #1509.

roberth:
> is there anything I have to do change about this pr?

I will test it some more and see if can find anything else. Perhaps you could review my additions in #1515?

pasqui23:
> I don't know. Did you try without this PR?

I can't, my config relies on `networking.nodesExtraArgs`

pasqui23:
@roberth I've added a simplified way to define resources with `defineResources`.

nix/net.nix
```diff
@@ -0,0 +1,126 @@
...
+let
+  inherit (lib) mkOption types mapAttrs warn;
+  #TODO: remove after merging https://github.com/NixOS/nixpkgs/pull/163617
+  deferredModule = with lib; mkOptionType {
+    name = "deferredModule";
+    description = "module";
+    check = t: isAttrs t || isFunction t || builtins.isPath t;
+    merge = loc: defs: map (def: lib.setDefaultModuleLocation "${showOption loc} from ${def.file}" def.value) defs;
+  };
+  # inherit (types) deferredModule;
```
roberth, line 11: Nixpkgs `deferredModule` now returns a singular module; see https://github.com/NixOS/nixpkgs/pull/163617/commits/4d94100d6ec131014bc2a8c2f1e06ee429c96967

pasqui23:
ok updated to only single import.
@roberth What do you think about `defineResource`?

pasqui23:
@roberth also why don't we contribute both on this branch instead of duplicating the work?

roberth:
> ok updated to only single import. @roberth What do you think about `defineResource`?

That looks really useful. Defining one was a little cumbersome. DRY-ing this also puts the nixops core in charge of precisely how resources are defined, which might prove useful one day.



> @roberth also why don't we contribute both on this branch instead of duplicating the work?

:+1:

nixops/deployment.py
```diff
@@ -439,6 +439,9 @@ def evaluate(self) -> None:
 
         # Extract info about other kinds of resources.
         for res_type, cfg in config["resources"].items():
+            # skip problematic resources entries
+            if res_type in ["deployment", "_name", "_type"]:
```
roberth, line 443: > @roberth what is about that `mypy-ratchet` failing test?

Apparently this new comparison doesn't have precise types. I don't think that's a big deal in itself.

I don't know where `_name` and `_type` are from, but I think you've added `deployment` through `deploymentInfoModule`. Those resource defaults can only realistically be used to set things in `_module`, because anything else will pollute actual resource parameters.
Setting things like aws defaults is enticing, but problematic and potentially confusing when later adding a non-aws resource.
`_module.args` doesn't seem that useful when resources can reference higher scopes including the whole network. You might as well reference those configs directly.
So I think the only use case for the resource defaults option is legacy compatibility.

pasqui23:
@roberth Is there anything missing before merging this pr?

pasqui23:
@roberth why is doc build failing?

pasqui23:
@roberth can we merge this or are there any blockers still?

pasqui23:
If it is of any consolation, I am using it in my homelab for a few months without any problems. I can however understand if it is a little too simple for commercial deployment.

pasqui23:
Ok now it gives me this error when the same deployment,same config, same flake.lock was ok just a few minutes ago…

```
; nixops deploy --include flatwoody
trace: warning: Please use nodes.flatwoody option instead of assigning machines to the config's top level
warning: binary cache 'http://127.0.0.1:12304' is for Nix stores with prefix 'Nix::Store::getStoreDir', not '/nix/store'
this derivation will be built:
  /nix/store/0vm5kgw4svxjwskx0di99z4i9qqkb95l-nixops-machines.drv
building '/nix/store/0vm5kgw4svxjwskx0di99z4i9qqkb95l-nixops-machines.drv'...
warning: you did not specify '--add-root'; the result might be removed by the garbage collector
flatwoody> copying closure...
flatwoody> copying 0 paths...
unnamed> closures copied successfully
flatwoody> uploading key ‘account_www.worldcommunitygrid.org.xml’ to ‘/var/lib/boinc/account_www.worldcommunitygrid.org.xml’...
flatwoody> chmod: modo non valido: "\"'0600'\""
flatwoody> Try 'chmod --help' for more information.
flatwoody> error: Traceback (most recent call last):
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/ssh_util.py", line 376, in run_command
    return nixops.util.logged_exec(cmd, self._logger, **kwargs)
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/util.py", line 372, in logged_exec
    raise CommandFailed(err, res)
nixops.util.CommandFailed: command ‘['ssh', '-oControlPath=/tmp/nixops-ssh-tmp2v174qig/master-socket', '-o', 'Port=22', '-o', 'StrictHostKeyChecking=accept-new', '-i', '/tmp/nixops-tmpffrbz4ga/id_nixops-flatwoody', '-x', 'root@flatwoody.local', '--', 'bash -c \'(  getent passwd \'"\'"\'root\'"\'"\' >/dev/null &&  getent group \'"\'"\'root\'"\'"\' >/dev/null &&  chown \'"\'"\'root:root\'"\'"\' \'"\'"\'/var/keys/.account_www.worldcommunitygrid.org.xml.tmp\'"\'"\' ); chmod \'"\'"\'0600\'"\'"\' \'"\'"\'/var/keys/.account_www.worldcommunitygrid.org.xml.tmp\'"\'"\'\'']’ failed on machine ‘flatwoody’ (exit code 1)

During handling of the above exception, another exception occurred:

  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 883, in worker
    m.send_keys()
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/backends/__init__.py", line 432, in send_keys
    self.run_command(
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/backends/__init__.py", line 599, in run_command
    return self.ssh.run_command(command, user=self.ssh_user, **kwargs)
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/ssh_util.py", line 378, in run_command
    raise SSHCommandFailed(exc.message, exc.exitcode)
nixops.ssh_util.SSHCommandFailed: command ‘['ssh', '-oControlPath=/tmp/nixops-ssh-tmp2v174qig/master-socket', '-o', 'Port=22', '-o', 'StrictHostKeyChecking=accept-new', '-i', '/tmp/nixops-tmpffrbz4ga/id_nixops-flatwoody', '-x', 'root@flatwoody.local', '--', 'bash -c \'(  getent passwd \'"\'"\'root\'"\'"\' >/dev/null &&  getent group \'"\'"\'root\'"\'"\' >/dev/null &&  chown \'"\'"\'root:root\'"\'"\' \'"\'"\'/var/keys/.account_www.worldcommunitygrid.org.xml.tmp\'"\'"\' ); chmod \'"\'"\'0600\'"\'"\' \'"\'"\'/var/keys/.account_www.worldcommunitygrid.org.xml.tmp\'"\'"\'\'']’ failed on machine ‘flatwoody’ (exit code 1)

  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/bin/.nixops-wrapped", line 9, in <module>
    sys.exit(main())
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/__main__.py", line 56, in main
    args.op(args)
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/script_defs.py", line 715, in op_deploy
    depl.deploy(
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 1365, in deploy
    self.run_with_notify("deploy", lambda: self._deploy(**kwargs))
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 1354, in run_with_notify
    f()
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 1365, in <lambda>
    self.run_with_notify("deploy", lambda: self._deploy(**kwargs))
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 1300, in _deploy
    self.activate_configs(
  File "/nix/store/fb1rfkicp42n9363ifhwvb8npvwiplzh-python3.10-nixops-2.0.0/lib/python3.10/site-packages/nixops/deployment.py", line 947, in activate_configs
    raise Exception(
Exception: activation of 1 of 1 machines failed (namely on ‘flatwoody’)
```

roberth:
> ```
> flatwoody> chmod: modo non valido: "\"'0600'\""
> ```

Looks like the deployment key mode got escaped/quoted too often.

pasqui23:
I think I know what happened: I put `users.defaultUserShell=pkgs.nushell` and on subsequent `nixops deploy` ssh ran nushell that then ran the bash command, but not before quoting it one more time.

This is a separate issue from this pr as this did not touch any python code.

pasqui23:
Ok why is `TestSSHKeyPairResource` failing?

roberth:
> Ok why is `TestSSHKeyPairResource` failing?

I've made some fixes so that a nixops_aws deployment with resources now works.

pasqui23:
On 20/03/23 08:01, Robert Hensing ***@***.***> wrote:
>     net.nix: removed nodes options, all network nodes are now defined in
>     resources.machines
> 
> This seems a bit too radical to me. |nodes| used to be consistent with 
> the NixOS tests

|nodes| is the only options nixops and the nixos test have in commons, I do not even belive that the nixos tests are implemented with the modules framework so it would only cause further confusions if we were to make the 2 more similar.

> and I believe the distinction between nodes and machine 
> resources may have been greater in the past, and completely merging the 
> two may be a very grave mistake.

On the contrary i joined the 2 for a reason of Single Source of Truth: the users are asked to set the nodes options yet the nixops script goes to look into a completely different option. It's not hard to imagine a bug where the 2 begin to diverge.

Worse such bug would be hard to even detect,let alone debug, as the user begin to look in the complete wrong position and the nix language support is extremely poor.

> There's no easy way out of such a decision.

|mkRenamedModule| and |mkRemovedModule| covers a multitude of sins 

> Besides the consistency and the risk, I feel like |nodes| is actually an 
> ingredient of a good solution, which is the idea that |resources| are 
> somewhat of a low level interface to the cloud, whereas almost all the 
> other options can be more high level.

I would go the opposite direction, with |machines| being merely one kind among many of cloud resources, and even being able to configure a network that has no machines in it.

It could even intrude into disnix's role and allow to configure,for example, a network of Java servlets on something like a Servlet as a Service or something.

> NixOS itself is actually one of 
> these more high level options. For instance, it can select between 
> multiple backends.
> When we support a generic resource backend like terraform, cloud support 
> would be largely be implemented in the module system,

Even now the we can access |resources.machines.*.system.build.toplevel| from inside the net module. It is still possible to implement the cloud supposrt from the module itself.

> and these NixOS 
> configurations won't be represented by a single resource that was 
> specifically written to be this all-encompassing NixOS + VM + SSH + 
> whatnot python object.

So we can still do all kind of object shuffling with the right |networkExpr| and they can still have all kind of backend-dependent magic that can be referred from the nodes themselves.

Hell we could even steal from |deploy,rs| and have |resource.upgradeScripts| all from the module system.

 
> So I'm a bit hesitant, but I can be convinced, maybe?

Remember that I have little experience with clud systems, let alone nixops specifically, so if any of the above sound wrong don't hesitate to call me out on it ☺

roberth:
I'll share a response with my thoughts so far. Haven't been able to look into it more deeply yet.

> I do not even belive that the nixos tests are implemented with the modules framework

They are.

> On the contrary i joined the 2 for a reason of Single Source of Truth: the users are asked to set the nodes options yet the nixops script goes to look into a completely different option. It's not hard to imagine a bug where the 2 begin to diverge.

If we have option definitions that assign the right thing, I don't see how divergence could reasonably occur, except perhaps through `mkForce` if at all possible, which users would avoid or do for a good reason.

I would prefer to keep the delta small until we can say something definitive about this `nodes` vs `resources.machines` choice.

What worries me is a problem that we have in NixOps evaluation, where it could take an arbitrary number of `nix-instantiate` invocations to properly wire each value from one resource's outputs to another resource's inputs, which produces an output that is then wired to yet another resource's inputs, and so forth. This is a serious issue that requires some rearchitecting to get right. I did not plan to solve this for a NixOps 2 release, as it depends on having proper Nix python bindings (in progress but will take a while to be sufficiently usable). So iiuc for now we're stuck with NixOps 1's phased evaluated that invokes the evaluator twice, and no more than that. It has some special logic for wiring up values between resources by name, so without going through the evaluator. NixOS basically does not fit within this resources paradigm.
For a more thorough and confident explanation I'd have to do more research, but realistically we're should not solve this specific problem for the NixOps 2 release, because we have more pressing issues and very limited resources.
So anyway, in the future we'll want to essentially replace the resource modules by a custom primop invocation backed by e.g. a terraform resource, and not much else. Having a whole NixOS configuration in there as a resource would not make much sense. It'd be like declaring your database configuration in `systemd.services`.

Anyway, I'll do a bit more research and some testing before I actually make a decision.

PR #1509, opened 2022-03-03
MachineState.run_command: remove duplicate flags

roberth:
The configured ssh options are already included since e21d5b3
https://github.com/NixOS/nixops/pull/1493

Solves errors like `Only a single -J option is permitted (use commas to separate multiple jump hops)`, provided that the duplicate `-J` isn't user error of course.


roberth:
Tested on a real deployment with jump host.

        deployment.sshOptions = [ "-J" "some-host" ];

PR #1515, opened 2022-04-06
eval-machine-info.nix: ported to modules [fork with additions]

roberth:
Work in progress.


pasqui23:
It still ask for password and if I try to deploy more than one node it fails to ssh connect.

pasqui23:
Also do I close https://github.com/NixOS/nixops/pull/1508 ?

roberth:
> It still ask for password and if I try to deploy more than one node it fails to ssh connect.

Did you try with `master` too? It could be an independent problem.

> Also do I close #1508 ?

No need.

pasqui23:
OK using your latest version I can, from flatwoody, deploy flatwoody without password prompt but it still ask for password prompt if I try to deploy nassie.
The network consist of only those 2 nodes and it can be found there:
https://gitlab.com/pasqui23/my-nixos-conf

pasqui23:
> Did you try with `master` too? It could be an independent problem.
Unfortunately I can't check as my config uses `networking.extraNodesArgs` introduced in our pr.

However your latest development in this pr fixes the problem of deploying it to flatwoody (the machine I ran nixops) but deployment to nassie still ask for password.

roberth:
@pasqui23 I don't think you've generated ssh keys with a NixOps resource, right? So then it _seems_ unlikely that these changes break your ssh config, without breaking other options like the aws ones, which work. I'm inclined to assume that you're experiencing a pre-existing problem, but let's see if we can find a solution.

We're going to need more info though, to see why it doesn't work.

 - When you're running the deployment, nixops waits for the `ssh` command to complete. Could you check with `ps` that the command line is correct? It may reference temporary files that should also exist for the duration of the ssh command. It might be `scp` instead of `ssh`.
 
  - Does your state file contain the private key? `nixops export` will print the state to stdout. Don't post it here, unless you have good reason to :)

  - You could add some debug `print`s to follow the nixops python code. (We might want to invest in some proper logging so we don't have to reinvent this every time...) (Also need better hacking docs)

pasqui23:
> * When you're running the deployment, nixops waits for the `ssh` command to complete. Could you check with `ps` that the command line is correct? It may reference temporary files that should also exist for the duration of the ssh command. It might be `scp` instead of `ssh`.

```
; ps 150476
    PID TTY      STAT   TIME COMMAND
 150476 pts/3    S+     0:00 ssh -x root@nassie.local -S /tmp/nixops-ssh-tmpfpe1egus/master-socket -M -N -f -oNumberOfPasswordPrompts=3 -oServerAliveInterval=60 -oControlPersist=600 -C -o Port=22
```

> * Does your state file contain the private key? `nixops export` will print the state to stdout. Don't post it here, unless you have good reason to :)

```
; nixops export
{
  "169e6beb-bc2a-11ec-abc6-44af2811d866": {
    "configsPath": "/nix/store/9zl1m3vrfd8ivpjz9w9qkxc79a40ba7j-nixops-machines",
    "description": "",
    "resources": {
      "flatwoody": {
        "configsPath": "/nix/store/9zl1m3vrfd8ivpjz9w9qkxc79a40ba7j-nixops-machines",
        "creationTime": "1649964952",
        "index": "0",
        "keys": present,
        "none.sshPrivateKey": "present",
        "none.sshPublicKey": "ssh-ed25519 present NixOps client key for flatwoody",
        "none.sshPublicKeyDeployed": "1",
        "privilegeEscalationCommand": "[\"sudo\", \"-H\", \"--\"]",
        "sshOptions": "[]",
        "sshPinged": "1",
        "state": "3",
        "stateVersion": "22.05",
        "targetHost": "flatwoody.local",
        "targetPort": "22",
        "toplevel": "/nix/store/yq6j72l6x1mq9ww6n0n825xgkqbsnlbz-nixos-system-flatwoody-22.05pre-git",
        "type": "none",
        "vmId": "nixops-169e6beb-bc2a-11ec-abc6-44af2811d866-flatwoody"
      },
      "nassie": {
        "creationTime": "1649965218",
        "index": "1",
        "keys": present,
        "none.sshPrivateKey": present",
        "none.sshPublicKey": "ssh-ed25519 present NixOps client key for nassie",
        "privilegeEscalationCommand": "[\"sudo\", \"-H\", \"--\"]",
        "sshOptions": "[]",
        "sshPinged": "1",
        "state": "3",
        "stateVersion": "22.05",
        "targetHost": "nassie.local",
        "targetPort": "22",
        "type": "none",
        "vmId": "nixops-169e6beb-bc2a-11ec-abc6-44af2811d866-nassie"
      }
    },
    "rollbackEnabled": "1"
  }
}
```
Notably `flatwoody` has `"none.sshPublicKeyDeployed": "1",` but not `nassie`

pasqui23:
Even after doing `nixops send-keys` it still does not have `"none.sshPublicKeyDeployed": "1",`

pasqui23:
Also `flatwoody` has a `configsPath` entry but `nassie` does not.

pasqui23:
The `configsPath` are directory that contain only a `flatwoody` directory, without sight of `nassie`.

pasqui23:
Ok, I've found the problem:
1. `nassie`'s config would build perfectly but some systemd units would fail to restart
2. Failing systemd unit restart would make `nixos-rebuild` exit with non-0
3. nixops would report the deployment operation as failed
4. nixops would not set `none.sshPublicKeyDeployed`
5. On next deployment nixops would use password auth as `none.sshPublicKeyDeployed` is not set.

I've managed to work around it by managing a `nixops deploy` fully successfull

pasqui23:
In my branch I've merged your latest edits and added defineResource, a simplified way to implement resources

roberth:
Resumed back in #1508 :tada:

Issue #1472, opened 2021-08-25
`network.storage.legacy.databasefile` setting has not been implemented


roberth:
My flake-based deployment contains
```
network.storage.legacy = {
  databasefile = "~/.nixops/deployments-nixops-example.nixops";
};
```

which points to a new path. Yet, I see all my usual legacy deployments in `nixops list`.

I can't find `databasefile` anywhere but in the NixOps 1->2 upgrade message.

abbradar:
Together with the new NixOps way of having explicit network directories this creates confusion and possibility of devastating errors. Consider two deployments, `production` and `development`, which use different network directories. On NixOps 1 I would have set different configuration paths and use `nixops -d production ...`. Currently when I'm in `development` network directory I see this:

```
+--------------------------------------+-------------+----------------------------------------------+------------+------+
| UUID                                 | Name        | Description                                  | # Machines | Type |
+--------------------------------------+-------------+----------------------------------------------+------------+------+
| e0a78705-3cff-11ea-85a3-8600003a0125 | development | Development deployment                       |          1 | none |
| 9c212957-69f2-11ea-88d6-8600003a0125 | production  | Development deployment                       |          1 | none |
+--------------------------------------+-------------+----------------------------------------------+------------+------+
```

I could by mistake run `nixops deploy -d production` here and destroy my servers.

PR #1525, opened 2022-07-05
docs: mention `$NIXOPS_STATE`

yajo:
As seen in https://github.com/NixOS/nixops/issues/1472, the current documented way to specify a different storage location does not really work.

However, one can add `$NIXOPS_STATE` and it works. Also this is a very comfortable options for integrating with i.e. devshell.

Thus I'm documenting it here, while the other issue gets fixed.


roberth:
Thanks!

PR #1529, opened 2022-07-08
feat(eval): opt-in support for nix-build

yajo:
Plugins might need to call this function to actually build something in the nix store.

See https://github.com/nix-community/nixops-vbox/pull/28#issuecomment-1177446597 for example.

We need a way to tell eval to do a build. Here it is.

@moduon MT-904


roberth (approved):
I suppose this might have been feasible to implement in two steps calling `nix-store -r` on the result of the eval, but this is more efficient.

yajo:
I didn't know that 😅 

Thanks anyway.

PR #1531, opened 2022-07-11
build: update backend to poetry-core

yajo:
This is [recommended by upstream poetry][1]. Without this new backend, you cannot do an editable install.

[1]: https://python-poetry.org/docs/pyproject/#poetry-and-pep-517

@moduon MT-904


roberth:
Works. Thanks!

PR #1533, opened 2022-07-25
Manage deployment locks in $NIXOPS_STATE dir if provided

kampka:
The locks feel a bit inconsistent atm. as the state is stored in $NIXOPS_STATE while locks are still stored in `$HOME/.nixops`.

### DIFF
--- a/nixops/deployment.py
+++ b/nixops/deployment.py
@@ -338,7 +338,7 @@ class Deployment:
         self,
     ) -> Any:  # FIXME: DeploymentLock is defined inside the function
         if self._lock_file_path is None:
-            lock_dir = os.environ.get("HOME", "") + "/.nixops/locks"
+            lock_dir = os.environ.get("NIXOPS_STATE", os.environ.get("HOME", "") + "/.nixops" ) + "/locks"
             if not os.path.exists(lock_dir):
                 os.makedirs(lock_dir, 0o700)
             self._lock_file_path = lock_dir + "/" + self.uuid


roberth:
I understand the feeling, but I'm not confident that this is the right thing to do in the context of state and locking plugins. How are these features supposed to interact?

roberth:
Perhaps the local deployment lock and the remote lock interface should be merged into one interface. Instead of a noop remote lock by default, we'd have the local deployment lock as the default, and the remote lock would replace the local lock.

Another option would be to move the locks to `/run`, essentially saying "this is non-persistent state that's managed for you; you don't need to care about its placement".

I guess what I'm trying to say is I haven't decided and I don't think I know enough about the design and implementation at this point to say anything definitive.
I might prioritize other PRs first, although I'd be happy to discuss the nixops state design and requirements.

kampka:
> Perhaps the local deployment lock and the remote lock interface should be merged into one interface. Instead of a noop remote lock by default, we'd have the local deployment lock as the default, and the remote lock would replace the local lock.

Honestly, I was not aware that nixops had a concept of remote logs till yesterday.
I agree with your suggestion, I think the local lock offers little benefit in presence of a remote lock, so merging the interfaces is probably a good idea.

> Another option would be to move the locks to `/run`, essentially saying "this is non-persistent state that's managed for you; you don't need to care about its placement".

This would of course solve my personal aesthetic requirements of programs not cluttering my `$HOME`.
I think it also makes sense regarding clarification about what those locks are intended for.

> I guess what I'm trying to say is I haven't decided and I don't think I know enough about the design and implementation at this point to say anything definitive. I might prioritize other PRs first, although I'd be happy to discuss the nixops state design and requirements.

Fair enough. As I said, this issue is not really a priority any more.

PR #1538, opened 2022-09-17
Fix flake outputs

Gabriella439:
Newer versions of Nix do not work with the old `flake.nix` because of the change described in:

https://github.com/NixOS/nix/issues/5532

This change fixes that while still preserving compatibility with older versions of Nix.


roberth (approved):

roberth:
Thanks!

PR #1539, opened 2022-09-27
Use new inotify tools alias

scottbot95:
A recent commit of nixpkgs removed the old alias of `inotifyTools` and causes it to fail to evaluate. This PR simply uses the new alias.


roberth (approved):

roberth:
Thanks!

